PR #1, opened 2025-02-20
Bring in mcp-* crates from goose as a starting point

alexhancock:
Initialize rust SDK with `mcp-*` crates from https://github.com/block/goose/tree/main/crates

@salman1993
@kalvinnchau

wrote much of this code, and @Jeadie also made a contribution before it moved from the goose repo here

### DIFF
--- a/crates/mcp-client/Cargo.toml
+++ b/crates/mcp-client/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
 name = "mcp-client"
-version = "0.1.0"
-edition = "2021"
+version.workspace = true
+edition.workspace = true
 
 [dependencies]
-mcp-core = { path = "../mcp-core" }
+mcp-core = { workspace = true }
 tokio = { version = "1", features = ["full"] }
 reqwest = { version = "0.11", default-features = false, features = ["json", "stream", "rustls-tls"] }
 eventsource-client = "0.12.0"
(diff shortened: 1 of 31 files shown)


crates/mcp-client/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "mcp-client"
+version = "0.1.0"
+edition = "2021"
```
dsp-ant: This should probably

```suggestion
version.workspace = true
edition.workspace = true
```

crates/mcp-client/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dependencies]
+mcp-core = { path = "../mcp-core" }
```
dsp-ant: I personally prefer having internal crates be workspace dependency, so in the workspace Cargo.toml you'd define it as a dependency 
```
[workspace.dependencies]
mcp-core = { path = "./mcp-core" }
```
and then use it here:
```suggestion
mcp-core = { workspace = true }
```

dsp-ant (approved):

alexhancock:
I neglected to include this in the commit message but it should be noted that:

@salman1993
@kalvinnchau

wrote this code, and @Jeadie also made a contribution before it moved from Goose repo here

Issue #6, opened 2025-02-26
Reorganise?


itsaphel:
May I gently suggest moving `crates/mcp-client` and `crates/mcp-server` to the `examples` directory? As they seem to be like sample implementations one could use, but are not actually part of the library as far as I can tell.

jlvihv:
Can someone come out and say a word? I agree with you. And it seems that the Router / Server / ByteTransport / BoundedService in crates/mcp-server and those defined errors should be moved to core to give the user a cleaner and leaner code example, can we do anything for this repository?

AdamSimpson:
Just wanted to gently _third_ that I find this repo quite confusing. As a rust developer I find `offeryn` has a familiar layout, with a top level `example` directory showing how to use the contained library. I've spent a few minutes looking at #8 and am honestly still not quite sure how to use this library efficiently.

baxen:
Yes we're overdue for examples and documentation! The layout across a few crates (similar to `offeryn`) I think is somewhat necessary, but we'll add top level examples to make this clear

itsaphel:
Shouldn't https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/mcp-server/src/main.rs at least be an example, since it's a particular usage of an mcp-server (a counter server), showing how the library should be used?

WSL0809:
> Yes we're overdue for examples and documentation! The layout across a few crates (similar to `offeryn`) I think is somewhat necessary, but we'll add top level examples to make this clear是的，我们确实需要提供示例和文档了！虽然跨多个 crate 的布局（类似于 `offeryn` ）是必要的，但我们会添加顶层示例来让这一点更加清晰

I completely agree with clearly documenting the project's examples and roadmap in the README.

itsaphel:
@kalvinnchau I think this remaining chunk is also part of an example: https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/mcp-server/src/router.rs#L395-L431

baxen:
Pausing this reorganization for a second so we can coordinate several other MCP builders who have reached out

PR #8, opened 2025-02-27
fix: rename core to spec

baxen:
Renaming core to spec to clarify organization (#6) and to avoid naming conflicts on crates.io


kalvinnchau (approved):

stevohuncho:
I run the mcp-core crate and was wondering if we could collaborate and solve this issue so this could be the official mcp-core

RWDai:
> I run the mcp-core crate and was wondering if we could collaborate and solve this issue so this could be the official mcp-core我运行了 mcp-core crate，想知道我们能否合作解决这个问题，这样就可以成为官方的 mcp-core

you delete you mcp-core package？

cbrit:
Somewhat related, I mentioned in #17 using generated bindings from the official spec's JSON schema rather than the current hand-rolled types. I've published a simple version at https://crates.io/crates/mcp-types, if there is interest, happy to remove so the name can be used. mcp-spec is nice too.

stevohuncho:
@baxen I'm the author of that crate and I don't mind changing it as long as all of these crates will be live on crates.io and this project is production ready as I currently run remote SSE mcp servers with my crate.

baxen:
@stevohuncho Hey! Thank you so much for reaching out and sorry for my slow reply, i've been offline last week. We're looking to bring everyone building these together into this space, and would be very enthusiastic for you to come contribute here.

Can you please hop into #29 - i'd like to kick off the discussion of combining what we all have got into this space there. We probably have a little up front work to pick between some interface choices and then let's get this into a stable release

baxen:
@cbrit Same comment! Let's start a discussion in #29 - generating the types seems like a great idea to me but want to connect this to other work

baxen:
given this discussion i'm going to hold off merging this PR

PR #28, opened 2025-03-13
pin transport once, use project in write_message

xuorig:
Bit of a drive-by PR while reading the rust implementation. Nice work!

I noticed the triple `Pin::new` in `write_message`. What do you think of pinning the `transport` once instead and using `project()` in `write_message`?


baxen (approved):
Yes good call!

Issue #31, opened 2025-03-14
Defining tools with macros


itsaphel:
I've been experimenting with what an MCP server looks like when tools are defined with the `#[tool]` macro, rather than implementing a whole router as the current example does. I think having regular functions annotated with decorators is a more ergonomic way to write MCP servers, from the developer experience point of view, and IMO what the experience should look like in the long term. The files can be found in my fork: https://github.com/itsaphel/modelcontextprotocol-rust-sdk/tree/main/examples/server_tool_macros

I've also made changes to allow for the registration of tools, and avoid the need to write routers in usages of the crate.

Creating this issue to document my experience and raise some things for discussion.

There's two server examples:

**1\. A stateless calculator** [[source](https://github.com/itsaphel/modelcontextprotocol-rust-sdk/blob/2ec28a8d9dbc8d9d264af4290ac52d2fad9676f7/examples/server_tool_macros/src/calculator_server.rs)]

This example works well enough. The only icky bit is the `call` function returns a JSON value (`Ok(serde_json::to_value(result).expect("should serialize"))`), which then needs to be deserialised again and converted into a `Content` variant.

**2\. Stateful functions** [[source](https://github.com/itsaphel/modelcontextprotocol-rust-sdk/blob/2ec28a8d9dbc8d9d264af4290ac52d2fad9676f7/examples/server_tool_macros/src/counter_server.rs)]

It's not currently possible to re-create the counter example in this repo using macros AFAICS, since the generated `call` invocation won't pass anything other than the schema as parameters to the function, making it completely pure with no dependencies.

I've played around with a few different ways of dealing with this. The first is allowing some kind of context to get injected in as a parameter. This is similar to what [actix](https://actix.rs/) does. Another might be to change the level of the tool decorator to a struct rather than a function. This would mean people write methods rather than pure functions, allowing them to refer to `&self` for context; [offeryn](https://github.com/avahowell/offeryn) does this.

Separately, I've been playing around with what this looks like without heavy macros (ie: more [axum](https://github.com/tokio-rs/axum) style). It's a matter of taste, but I think overuse of macros leads to frustrating developer experience when debugging errors, and creates a bit too much 'magic' behind the scenes. Offeryn's macros are pretty heavy on the amount of code they generate. So I guess the first question is: do we go with a macro-heavy approach or not?

Perhaps worth some discussion on what this should look like?

baxen:
[...]

PR #39, opened 2025-03-19
Remove Send + Sync bounds

itsaphel:
In the interest of starting PRs rather than more issues, I'd like to use this PR discuss whether we use `Send + Sync` bounds everywhere.

## Motivation and Context
There's a fair bit of writing on how async Rust becomes more elegant without `Send + Sync` everywhere [[1]] [[2]], some discussion [[3]], and for balance some disagreement too [[4]]. Ultimately, these bounds are only required because tokio is a work-stealing multi-OS-threaded executor. Their unfortunate side effect is you need to use `Arc`s and `Mutex`es everywhere.

In my fork (which I'll break down and PR up), I've been experimenting with injecting state structs into tool handlers (discussion [here](https://github.com/modelcontextprotocol/rust-sdk/issues/31#issuecomment-2736285245)). One thing I did to make life easier while POC'ing was remove the `Send + Sync` bounds. Aside from making the internal code nicer, more importantly, it made the application code easier to write.

I wonder whether we should have them at all. I'd expect workloads to be primarily IO bound. In which case, running a non-work-stealing executor (or tokio in local mode with, say, a `LocalSet`) should let us get rid of the `Send`, and if it's single-threaded then the `Sync` can go too. That would mean user code can also avoid `Arc` and `Mutex` if they don't need it. There's also some inspiration from actix-web here, with [their individual workers](https://docs.rs/actix-web/latest/actix_web/rt/index.html) using single-threaded runtimes, and the application being able to opt in to multiple workers / shared state across workers if it wishes (and if so, they'd then need to be `Send + Sync` on the state), see [here](https://actix.rs/docs/application#shared-mutable-state).

The cost of futures would no longer work with the work-stealing functionality of the runtime.

## How Has This Been Tested?
`counter_server` compiles and runs. The axum example does not. I think it can be modified to work though.

## Breaking Changes
Yes, but we're pre-alpha!

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

[1]: https://emschwartz.me/async-rust-can-be-a-pleasure-to-work-with-without-send-sync-static/
[2]: https://maciej.codes/2022-06-09-local-async.html
[3]: https://blog.yoshuawuyts.com/tasks-are-the-wrong-abstraction/
[4]: https://without.boats/blog/thread-per-core/

### DIFF
--- a/crates/mcp-core/src/handler.rs
+++ b/crates/mcp-core/src/handler.rs
@@ -39,7 +39,7 @@ pub enum PromptError {
 
 /// Trait for implementing MCP tools
 #[async_trait]
-pub trait ToolHandler: Send + Sync + 'static {
+pub trait ToolHandler: 'static {
     /// The name of the tool
     fn name(&self) -> &'static str;
 
@@ -55,7 +55,7 @@ pub trait ToolHandler: Send + Sync + 'static {
 
 /// Trait for implementing MCP resources
 #[async_trait]
-pub trait ResourceTemplateHandler: Send + Sync + 'static {
+pub trait ResourceTemplateHandler: 'static {
     /// The URL template for this resource
     fn template() -> &'static str;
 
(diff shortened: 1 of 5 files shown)


4t145:
Hello, I will close this pr since we have a totally different implementation now.

PR #51, opened 2025-03-28
fix(server): add error enum while deal server info

jokemanfire:
1. wrap the error type for more standardized
2. add more information in error for debug trace
3. wrap helper func for more user-friendly code
    
## Motivation and Context
I found the error's message is not great while debugging.

## How Has This Been Tested?
CI 

## Breaking Changes
No

## Types of changes
Enhance the error deal in server

## Checklist
all check

### DIFF
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -1,14 +1,15 @@
 use futures::{SinkExt, StreamExt};
+use thiserror::Error;
 
 use super::*;
 use crate::model::{
-    CancelledNotification, CancelledNotificationParam, ClientInfo, ClientNotification,
-    ClientRequest, ClientResult, CreateMessageRequest, CreateMessageRequestParam,
-    CreateMessageResult, ListRootsRequest, ListRootsResult, LoggingMessageNotification,
-    LoggingMessageNotificationParam, ProgressNotification, ProgressNotificationParam,
-    PromptListChangedNotification, ResourceListChangedNotification, ResourceUpdatedNotification,
-    ResourceUpdatedNotificationParam, ServerInfo, ServerMessage, ServerNotification, ServerRequest,
-    ServerResult, ToolListChangedNotification,
+    CancelledNotification, CancelledNotificationParam, ClientInfo, ClientJsonRpcMessage,
+    ClientMessage, ClientNotification, ClientRequest, ClientResult, CreateMessageRequest,
+    CreateMessageRequestParam, CreateMessageResult, ListRootsRequest, ListRootsResult,
+    LoggingMessageNotification, LoggingMessageNotificationParam, ProgressNotification,
+    ProgressNotificationParam, PromptListChangedNotification, ResourceListChangedNotification,
+    ResourceUpdatedNotification, ResourceUpdatedNotificationParam, ServerInfo, ServerMessage,
+    ServerNotification, ServerRequest, ServerResult, ToolListChangedNotification,
 };
 
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/service/server.rs
```diff
@@ -26,6 +20,24 @@ impl ServiceRole for RoleServer {
...
 
+/// It represents the error that may occur when serving the server.
+///
+/// if you want to handle the error, you can use `serve_server_with_ct` or `serve_server` with `Result<RunningService<RoleServer, S>, ServerError>`
+#[derive(Error, Debug)]
+pub enum ServerError {
+    #[error("expect initialize request, but received: {0:?}")]
+    ExpectedInitRequest(Option<ClientMessage>),
+    
+    #[error("expect initialize notification, but received: {0:?}")]
```
4t145: "expect **initialized** notification" is the more accurate expression.

crates/rmcp/src/service/server.rs
```diff
@@ -1,13 +1,7 @@
...
-    CancelledNotification, CancelledNotificationParam, ClientInfo, ClientNotification,
-    ClientRequest, ClientResult, CreateMessageRequest, CreateMessageRequestParam,
-    CreateMessageResult, ListRootsRequest, ListRootsResult, LoggingMessageNotification,
-    LoggingMessageNotificationParam, ProgressNotification, ProgressNotificationParam,
-    PromptListChangedNotification, ResourceListChangedNotification, ResourceUpdatedNotification,
-    ResourceUpdatedNotificationParam, ServerInfo, ServerMessage, ServerNotification, ServerRequest,
-    ServerResult, ToolListChangedNotification,
+    CancelledNotification, CancelledNotificationParam, ClientInfo, ClientJsonRpcMessage, ClientMessage, ClientNotification, ClientRequest, ClientResult, CreateMessageRequest, CreateMessageRequestParam, CreateMessageResult, ListRootsRequest, ListRootsResult, LoggingMessageNotification, LoggingMessageNotificationParam, ProgressNotification, ProgressNotificationParam, PromptListChangedNotification, ResourceListChangedNotification, ResourceUpdatedNotification, ResourceUpdatedNotificationParam, ServerInfo, ServerMessage, ServerNotification, ServerRequest, ServerResult, ToolListChangedNotification
```
4t145: It seems like we have different fmt configuration and should we add a `rustfmt.toml` to root menu?

crates/rmcp/src/service/server.rs
```diff
@@ -70,54 +122,46 @@ where
...
-        .ok_or(std::io::Error::new(
-            std::io::ErrorKind::InvalidData,
-            "expect initialize request",
-        ))?;
+    // Convert ServerError to std::io::Error, then to E
+    let handle_server_error = |e: ServerError| -> E {
+        match e {
+            ServerError::Io(io_err) => io_err.into(),
+            other => std::io::Error::new(
+                std::io::ErrorKind::Other,
```
4t145: I'm not sure should we add a tracing log here or let users log the error outside?

4t145 (changes requested):
Thanks for you pr, this looks good to me. Should we do the same thing in the client side?

jokemanfire:
> Thanks for you pr, this looks good to me. Should we do the same thing in the client side?感谢您的 PR，看起来不错。我们在客户端是否也应该做同样的事情？

I will do same work in client, and file a another pr.
As we can see now, the CI was not triggered correctly. I think we should wait until the CI is normal before making any code modifications, otherwise we cannot guarantee the standardization of the code.

crates/rmcp/src/service/server.rs
jokemanfire: I think this is necessary.

crates/rmcp/src/service/server.rs
jokemanfire: done

crates/rmcp/src/service/server.rs
jokemanfire: I think at this stage, we should return to the situation where transport uses io, otherwise it will be difficult to locate errors

4t145 (approved):

PR #52, opened 2025-03-28
refactor(sse_server): separate router and server startup

Ronbb:
This PR refactors SSE server initialization by separating router creation from server startup. Added `SseServer::new()` method to create router and server instance independently.

---

## Motivation and Context  
  
**Problem**: The original implementation coupled server startup (`serve_with_config`) with router creation, making it hard to test or reuse the router configuration.  
**Solution**: Introduced a factory method `SseServer::new()` that:  
1. Decouples router/state creation from server execution  
2. Enables reusing the router configuration in tests  
3. Allows pre-building services before binding ports

---

## How Has This Been Tested?  
  

---

## Breaking Changes  
  
**No breaking changes**

---

## Types of Changes  

- [ ] Documentation update

---

## Checklist  
  
---

### DIFF
--- a/crates/rmcp/src/transport/sse_server.rs
+++ b/crates/rmcp/src/transport/sse_server.rs
@@ -219,9 +219,9 @@ impl SseServer {
         .await
     }
     pub async fn serve_with_config(config: SseServerConfig) -> io::Result<Self> {
-        let (sse_server, service) = Self::new(&config);
-        let listener = tokio::net::TcpListener::bind(config.bind).await?;
-        let ct = config.ct.child_token();
+        let (sse_server, service) = Self::new(config);
+        let listener = tokio::net::TcpListener::bind(sse_server.config.bind).await?;
+        let ct = sse_server.config.ct.child_token();
         let server = axum::serve(listener, service).with_graceful_shutdown(async move {
             ct.cancelled().await;
             tracing::info!("sse server cancelled");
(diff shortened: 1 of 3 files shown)


4t145:
The original purpose of the sse-server transport layer is to provide users with a quick start entry just like what they do in python sdk `FastMcp`. https://github.com/4t145/rmcp/issues/27 . Personally I believe people who want to intergrate router to their existed http service should rewrite these logic. They could use axum, actix-web or the other http frameworks and we can't support those all in this crate.

Adding a method allowing user to get the router is not something bad definitely. So it's ok for me to apply these changes. But I think we should docment that we don't encourage people to do so when they are developing a large complex application based on that.

Ronbb:
@4t145 I agree that implementing MCP across all HTTP frameworks here isn't ideal. 

And, nested router behavior might introduce edge cases (needs further testing, `post_path` may be wrong). 

I think MCP SSE, Streamable HTTP, and session management should likely be SDK-owned. However, currently, these are tightly coupled to Axum (via non-public handlers and the `App`), that makes future migrations difficult. 

Would you like to clarify how this will evolve in future crate updates?

4t145:
@Ronbb My plan is to provide a session manager provides serveral request related message streams. And we may still remain a quick acsess by using axum since the users who want a quick start would not mind which web framework they are using.

4t145 (approved):

Issue #49, opened 2025-03-28
Add the verification of the signature in the commit in CI


jokemanfire:
I think it is necessary to add the verification of the signature in the commit in CI, as it has not been added in the current submission.

4t145:
Yes we do need a ci for rustdoc

Z9RTM:
I'm interested in this one.

Besides the points already listed, I think it might also be beneficial to consider adding CI for:
- build-workspace
- cross-compile
- coverage
- bench
- security (e.g., using cargo audit)
- deps-update (dependabot)
- hack (e.g., using cargo hack)

jokemanfire:
> I'm interested in this one.我对这个感兴趣。
> 
> Besides the points already listed, I think it might also be beneficial to consider adding CI for:除了已列出的几点，我认为还应该考虑添加 CI 来：
> 
> * build-workspace
> * cross-compile  跨平台编译
> * coverage  覆盖率
> * bench  基准测试
> * security (e.g., using cargo audit)安全（例如，使用 cargo audit）
> * deps-update (dependabot)
> * hack (e.g., using cargo hack)hack (例如，使用 cargo hack)

Thanks , but there are some CI that should be considered after the first version is released.
1.cross-compile
2.benchmark
3.deps-update
If you could help with this work, thank you very much.

Z9RTM:
Got it. I'll take this on. Please assign it to me.

jokemanfire:
May we need show code coverage rate in readme.

Z9RTM:
Sounds good. I'll take it into consideration.

Z9RTM:
Considering the conversation, a simple approach to dynamically display a coverage badge could be achieved by integrating grcov and Codecov. I was wondering if we could look into using Codecov?

Here's an example: https://github.com/codecov/example-rust/blob/main/.github/workflows/rust.yml

jokemanfire:
Yes, we need to consider using it

Z9RTM:
@dsp-ant 
Hi,
As discussed, we're looking into ways to improve the visibility of the code coverage for the `rust-sdk` project.

A potential solution is to use Codecov for `rust-sdk`. This service allows us to upload coverage reports generated (e.g., using `grcov`) and provides a convenient way to display the coverage as a badge in the `rust-sdk` README. This badge would update automatically with each build, giving a clear and current overview of our test coverage for `rust-sdk`.

The benefits of using Codecov for `rust-sdk` include:

[...]

PR #59, opened 2025-03-29
ci: add documentation generation job

jokemanfire:
1. add doc ci in workflow
2. remove the readme in rmcp crate

## Motivation and Context
Doc's CI is missing

## How Has This Been Tested?
no

## Breaking Changes
no

## Types of changes

- [x] Documentation update

## Checklist


jokemanfire:
#49

crates/rmcp/src/lib.rs
4t145, line 1: I am not sure it's OK when publishing package? I used to do so but it turns out cargo can not find the readme when package rmcp.

jokemanfire: It should be a separate crate file that has been opened in crate. doc, so there is no way to access the readme in the upper level directory.
I would like to add a separate readme to each crate in the future and use the top-level readme as an index to view the documentation for each crate. What do you think of this ？(to be modified before publishing)

4t145: Can we copy the readme in root dir to crate dir when packaging crate? Or can we create a symlink from crate level readme to root dir?

jokemanfire: I think copying a past plan is not feasible. We should try our best to avoid duplicate file copying and usage, The symbolic link of crate should not be feasible. I think providing links to the documentation of each crate through the root readme is the best way, as most projects do this.

jokemanfire:
@4t145  How about we first submit and merge this (now CI needs to include this check)? and I will modify the doc before crate's publishing

4t145 (approved):

PR #67, opened 2025-03-30
Sse client transport trait

EItanya:
Add an `SseClient` trait which allows for users to plug in their own transports for client connection.

## Motivation and Context

I am developing an SSE client for MCP, however I cannot use the `reqwest` library for what I am trying to do, therefore I abstracted the logic of the `reqwest` library into a trait which can be potentially implemented by many client types.

## Types of changes

- [ ] Documentation update

## Checklist


crates/rmcp/src/transport/sse.rs
```diff
@@ -288,21 +316,19 @@ impl Sink<ClientJsonRpcMessage> for SseTransport {
...
-        tokio::spawn(async move {
-            let result = request_builder
-                .send()
-                .await
-                .and_then(|resp| resp.error_for_status())
-                .map_err(SseTransportError::from)
-                .map(drop);
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        let session_id = session_id.clone();
+        futures::executor::block_on(async move {
```
4t145: I think we should use `tokio::spawn` instead of `block_on` of another async runtime?

examples/clients/src/sse.rs
```diff
@@ -16,7 +16,8 @@ async fn main() -> Result<()> {
         )
         .with(tracing_subscriber::fmt::layer())
         .init();
-    let transport = SseTransport::start("http://localhost:8000/sse").await?;
```
4t145: I still want to remain the usage of `SseTransport::start("http://localhost:8000/sse")` as a way to quick start. And we can leave `rmcp::transport::SseTransport::start_with_client(client)` for a general usage. Sounds good to you?

crates/rmcp/src/transport/sse.rs
```diff
@@ -60,67 +62,119 @@ impl Default for SseTransportRetryConfig {
...
+        let last_event_id = last_event_id.clone();
+        let fut = async move {
+            let mut request_builder = client.get(&sse_url).header(ACCEPT, MIME_TYPE);
+            if let Some(last_event_id) = last_event_id {
+                request_builder = request_builder.header(HEADER_LAST_EVENT_ID, last_event_id);
+            }
+            let response = request_builder.send().await?;
+            let response = response.error_for_status()?;
+            match response.headers().get(reqwest::header::CONTENT_TYPE) {
+                Some(ct) => {
```
4t145, line 146: Oh I forgot there's a bug here, we should use `start_with` instead of `eq`, could you fix this together in this pr?

crates/rmcp/src/transport/sse.rs
```diff
@@ -60,67 +62,119 @@ impl Default for SseTransportRetryConfig {
...
+                    return Err(SseTransportError::UnexpectedContentType(None));
+                }
+            }
+            let event_stream = SseStream::from_byte_stream(response.bytes_stream()).boxed();
+            Ok(event_stream)
+        };
+        fut.boxed()
+    }
+
+    fn post(&self, session_id: &str, message: ClientJsonRpcMessage) -> BoxFuture<'static, Result<(), SseTransportError>> {
```
4t145: Why we use ident `session_id` here? I think `endpoint` or `post_url` is better since it could be a path and the session id is optional. (It depends on server implementation.)

4t145 (changes requested):

crates/rmcp/src/transport/sse.rs
EItanya: So I switched this because I couldn't get 'tokio::spawn' to work. The client lock isn't movable because the compiler can't figure out what the lifetime of the thread is. I'm relatively new to rust so I wasn't sure if there's a typical way to deal with that scenario.

crates/rmcp/src/transport/sse.rs
EItanya: Makes sense, I'll switch to endpoint

crates/rmcp/src/transport/sse.rs
EItanya: Never mind, figured it out

crates/rmcp/src/transport/sse.rs
EItanya: Which `eq` are you talking about? I'm not 100% sure what you mean

jokemanfire:
Can you rebase to one or two commits? I think this pr, no need some many commits. And  rebase to latest main pls.

jokemanfire:
I think too many synchronization operations here will affect the scheduling threads of tokio, and I think the implementation of this area should be optimized.

EItanya:
@jokemanfire I completed the rebase. 

> I think too many synchronization operations here will affect the scheduling threads of tokio, and I think the implementation of this area should be optimized.

I agree with you, but I think this PR is big enough already, can we do that as a follow-up?

crates/rmcp/src/transport/sse.rs
4t145: use `!ct.start_with(MIME_TYPE)` instead of `ct != MIME_TYPE`

4t145:
> I think too many synchronization operations here will affect the scheduling threads of tokio, and I think the implementation of this area should be optimized.

We can squash and merge, so maybe it's OK since those commits will become one when they are merged into main branch.

EItanya:
Hello @4t145, I have fixed all of the review comments, lmk if anything else is needed!

jokemanfire (approved):

jokemanfire:
pls fix it . CI not pass.

EItanya:
The tests are failing even though I changed nothing related, it says it's a tracing library issue. Can you take a look? I will fix the linting issues.

4t145 (dismissed):

4t145:
@EItanya  Tests should pass after #68 and #70 merged. You can merge the main branch and test again.

EItanya:
Sorry for the commit mangling, but everything should be properly formatted now

4t145:
I know that is a little bit annoying but could you run `cargo +nightly fmt --all` again. As for the commit message, I think we can fix it when squash and merge.

EItanya:
Ok, I think I finally got it 😆

4t145 (approved):

4t145:
Thank you for PR!

PR #84, opened 2025-04-02
Revision 2025-03-26 (without streamable http)

4t145:
1. Suppot revision 2025-03-26 data types
2. Support meta, progress tokne, and extensions in request/notification,
3. Remove `Message`, use `JsonRpcMessage` directly

## Motivation and Context

Support revision 2025-03-26

## Breaking Changes

1. Add new types and fields to satisfy the revision 2025-03-26
2. Now `PaginationParam` is the previous `PaginationParamInner`

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -2,18 +2,24 @@ use std::{borrow::Cow, sync::Arc};
 mod annotated;
 mod capabilities;
 mod content;
+mod extension;
+mod meta;
 mod prompt;
 mod resource;
+mod serde_impl;
 mod tool;
-
 pub use annotated::*;
 pub use capabilities::*;
 pub use content::*;
+pub use extension::*;
+pub use meta::*;
 pub use prompt::*;
 pub use resource::*;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 pub use tool::*;
+
+/// You can use [`crate::object!`] or [`crate::model::object`] to create a json object quickly.
 pub type JsonObject<F = Value> = serde_json::Map<String, F>;
 
 /// unwrap the JsonObject under [`serde_json::Value`]
@@ -28,6 +34,7 @@ pub fn object(value: serde_json::Value) -> JsonObject {
     }
 }
 
+/// Use this macro just like [`serde_json::json!`]
 #[cfg(feature = "macros")]
 #[macro_export]
 macro_rules! object {
@@ -85,7 +92,7 @@ macro_rules! const_string {
 
 const_string!(JsonRpcVersion2_0 = "2.0");
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, PartialOrd)]
 pub struct ProtocolVersion(Cow<'static, str>);
 
 impl Default for ProtocolVersion {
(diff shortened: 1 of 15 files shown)


crates/rmcp/src/model.rs
```diff
@@ -116,17 +141,28 @@ impl<'de> Deserialize<'de> for ProtocolVersion {
         #[allow(clippy::single_match)]
         match s.as_str() {
             "2024-11-05" => return Ok(ProtocolVersion::V_2024_11_05),
+            "2024-03-26" => return Ok(ProtocolVersion::V_2025_03_26),
```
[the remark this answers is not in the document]
4t145: fixed

4t145:
I would like to fix the ci lint message when merge.

crates/rmcp/src/model/extension.rs
```diff
@@ -0,0 +1,335 @@
+//! Copied from http
```
jokemanfire: Is there any problem with the license? We may need to confirm here

jokemanfire (commented):
The amount of modification is quite large. Please give me more time to take a look

crates/rmcp/src/model.rs
```diff
@@ -94,8 +101,26 @@ impl Default for ProtocolVersion {
...
+            let s_day = rest;
+            let year = s_year.parse::<u16>().ok()?;
+            let month = s_month.parse::<u16>().ok()?;
+            let day = s_day.parse::<u16>().ok()?;
+            Some((year, month, day))
+        }
+        let self_date = parse(self.0.as_ref())?;
+        let other_date = parse(other.0.as_ref())?;
+        Some(self_date.cmp(&other_date))
+    }
```
kalvinnchau: do we need to implement `PartialOrd` for this? Since versions are ISO8601 date strings they are naturally lexicographically ordered already

```rust
use std::borrow::Cow;

#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
pub struct ProtocolVersion(Cow<'static, str>);

impl ProtocolVersion {
    pub const LATEST: Self = Self::V_2025_03_26;
}
```

Should still pass the `test_protocol_version_order` test.

crates/rmcp/src/model/tool.rs
```diff
@@ -18,6 +18,92 @@ pub struct Tool {
...
+///
+/// NOTE: all properties in ToolAnnotations are **hints**.
+/// They are not guaranteed to provide a faithful description of
+/// tool behavior (including descriptive properties like `title`).
+///
+/// Clients should never make tool use decisions based on ToolAnnotations
+/// received from untrusted servers.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct ToolAnnotations {
```
kalvinnchau, line 35: a little unclear on the spec here:
all these fields are optional per https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts#L730, but the booleans have specific default values


do we need accessors like
```rust
    pub fn is_destructive(&self) -> bool {
        self.destructive_hint.unwrap_or(true)
    }

    pub fn is_idempotent(&self) -> bool {
        self.idempotent_hint.unwrap_or(false)
    }
```

to match the spec?

crates/rmcp/src/model.rs
4t145: Indeed, we don't really need to impl PartialOrd

crates/rmcp/src/model/tool.rs
4t145: Looks good!

crates/rmcp/src/model/extension.rs
4t145: `http` has a MIT license, I will add more information including a link to the license and how I modified this file.

crates/rmcp/src/transport/sse_server.rs
4t145, line 1: That is for debug and I will revert these changes.

4t145:
In China we are going to take a vacation for Qingming Festival these days, so the reply may not be so timely.

jokemanfire (approved):

EItanya (approved):

PR #93, opened 2025-04-04
Upgrade Rust 1.86

blue-orange-yellow:
# Motivation and Context

Upgraded Rust version to 1.86.
Fixed code according to new Clippy lints introduced in this version.
Also removed warnings shown during cargo check.

# How Has This Been Tested?

Tested locally using Rust 1.86.
Confirmed all tests pass with cargo test and no warnings from cargo check.

# Breaking Changes

No breaking changes in functionality.
However, developers will need to use Rust 1.86 or later.

# Checklist

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -54,7 +54,7 @@ rand = { version = "0.9", optional = true }
 tokio-stream = { version = "0.1", optional = true }
 
 # macro
-rmcp-macros = { version = "0.1", workspace = true, optional = true }
+rmcp-macros = { workspace = true, optional = true }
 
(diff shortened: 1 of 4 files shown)


blue-orange-yellow:
I'd appreciate it if someone could review this PR.

4t145:
I guess we are not supposed to upgrade msrv when we haven't use the newest features?

blue-orange-yellow:
You're right — we're not using any 1.86-specific features at the moment.
I upgraded the version mainly to align with our current toolchain and CI environment, and to keep up with the latest stable release.
But I'm happy to revert the change if we want to stick to the current MSRV policy.

PR #112, opened 2025-04-09
fix: fix 2 cargo warnings

rikhuijzer:
This fixes three warnings that `cargo test` produced.

## Motivation and Context

Locally and in CI the following warnings were visible:

```
$ cargo test --all-features
warning: /home/runner/work/rust-sdk/rust-sdk/examples/rig-integration/Cargo.toml: unused manifest key: dependencies.rmcp.no-default-features
warning: /home/runner/work/rust-sdk/rust-sdk/crates/rmcp/Cargo.toml: unused manifest key: dependencies.rmcp-macros.version
warning: /home/runner/work/rust-sdk/rust-sdk/examples/wasi/Cargo.toml: version requirement `0.14.2+wasi-0.2.4` for dependency `wasi` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
```

## How Has This Been Tested?

I re-ran all tests and also manually "tested" (built) `rig-integration` via

```sh
$ cd examples/righ-integration

$ cargo build
```

For `rig-integration` I verified that setting `default-features = false` will not be able to build the project, so that's why I removed it. Note that `default-features = false` is interpreted by Cargo while the setting `no-default-features = true` was ignored.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -51,7 +51,7 @@ rand = { version = "0.9", optional = true }
 tokio-stream = { version = "0.1", optional = true }
 
 # macro
-rmcp-macros = { version = "0.1", workspace = true, optional = true }
+rmcp-macros = { workspace = true, optional = true }
 
 [features]
 default = ["base64", "macros", "server"]
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -51,7 +51,7 @@ rand = { version = "0.9", optional = true }
 tokio-stream = { version = "0.1", optional = true }
 
 # macro
-rmcp-macros = { workspace = true, optional = true }
+rmcp-macros = { version = "0.1", workspace = true, optional = true }
 
 [features]
 default = ["base64", "macros", "server"]
--- a/examples/rig-integration/Cargo.toml
+++ b/examples/rig-integration/Cargo.toml
@@ -18,7 +18,7 @@ rmcp = { path = "../../crates/rmcp", features = [
     "client",
     "transport-child-process",
     "transport-sse",
-], no-default-features = true }
+] }
 anyhow = "1.0"
 serde_json = "1"
 serde = { version = "1", features = ["derive"] }
--- a/examples/wasi/Cargo.toml
+++ b/examples/wasi/Cargo.toml
@@ -17,7 +17,7 @@ readme = { workspace = true }
 crate-type = ["cdylib"]
 
 [dependencies]
-wasi = { version = "0.14.2+wasi-0.2.4"}
+wasi = { version = "0.14.2"}
 tokio = { version = "1", features = ["rt", "io-util", "sync", "macros", "time"] }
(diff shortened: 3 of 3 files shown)


crates/rmcp/Cargo.toml
```diff
@@ -51,7 +51,7 @@ rand = { version = "0.9", optional = true }
 tokio-stream = { version = "0.1", optional = true }
 
 # macro
-rmcp-macros = { version = "0.1", workspace = true, optional = true }
+rmcp-macros = { workspace = true, optional = true }
```
rikhuijzer: This version was ignored according to Cargo. As far as I understand, `workspace` already states that the crate inside the workspace should be used so it makes sense that the version is ignored. To double check I've also looked at a sub-crate in wasmtime and they also do not specify a version (for example, https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-nn/Cargo.toml).

examples/rig-integration/Cargo.toml
```diff
@@ -18,7 +18,7 @@ rmcp = { path = "../../crates/rmcp", features = [
     "client",
     "transport-child-process",
     "transport-sse",
-], no-default-features = true }
```
rikhuijzer, line 21: This was ignored by cargo because it should be `default-features = false`.

examples/wasi/Cargo.toml
```diff
@@ -17,7 +17,7 @@ readme = { workspace = true }
 crate-type = ["cdylib"]
 
 [dependencies]
-wasi = { version = "0.14.2+wasi-0.2.4"}
```
rikhuijzer, line 20: This was also ignored by Cargo according to 
```
 warning: /home/runner/work/rust-sdk/rust-sdk/examples/wasi/Cargo.toml: version requirement `0.14.2+wasi-0.2.4` for dependency `wasi` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
```

crates/rmcp/Cargo.toml
4t145: This needs a further research on how they publish crate. As far as I know, omit the dependency version can be a hard error when publish crate to crates.io.

rikhuijzer: Thanks for the review @4t145. I reverted this change in https://github.com/modelcontextprotocol/rust-sdk/pull/112/commits/e735fa0687e19c055fa4a3cf5f6dfed6db92af55.

rikhuijzer: I'll think about the publishing situation but I personally have no experience with that so that's why I reverted the change.

4t145: I just have a look at [tokio](https://github.com/tokio-rs/tokio/blob/817fa605ee6a2549fe8e6057ec23a8309d42d2e9/tokio/Cargo.toml#L89C1-L89C81), maybe we should adopt this approach?

4t145 (approved):
LGTM, we can fix the rest one warning latter. Thank you for PR.

PR #116, opened 2025-04-09
docs: add an overview to `rmcp/src/lib.rs`

rikhuijzer:
This PR adds a global overview of the crate to the docs homepage. Also it include some example code to get started building a server and some example code to build a client.

The text and code is based on reading through <https://modelcontextprotocol.io>, the examples in this repository, and a bit of reading in the Python SDK.

## Motivation and Context

I always like `docs.rs` documentation a lot because the docs can show whether there are features that I should enable and often docs can easily be jumped around in. I always am a bit sad if I go to the docs and find nothing. This PR suggests to change that by having at least a bit of a global outline. A sort of starting point so to say. 

## How Has This Been Tested?

These changes are tested as part of `cargo test`. I ran this myself multiple times during writing. The first example (server) actually runs during the test. The second example (client) is only tested for being able to compile. I tried to run it too as part of the tests, but then the tests hang. Probably some server that isn't cleaned up properly, so just to be safe lets not start it in the first place.

## Breaking Changes

No.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/lib.rs
+++ b/crates/rmcp/src/lib.rs
@@ -1,3 +1,92 @@
+//! The official Rust SDK for the Model Context Protocol (MCP).
+//!
+//! The MCP is a protocol that allows AI assistants to communicate with other
+//! services. `rmcp` is the official Rust implementation of this protocol.
+//!
+//! There are two ways in which the library can be used, namely to build a
+//! server or to build a client.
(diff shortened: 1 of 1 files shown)


jokemanfire (approved):
thanks

PR #122, opened 2025-04-10
docs: simplify counter example

rikhuijzer:
I thought `_create_resource_text` was part of the public API, but noticed that it's only used inside this file. To make that more clear, I moved the function definition closer to the callsite.

## Motivation and Context

It makes the example a bit clearer.

## How Has This Been Tested?

I manually ran the examples.

## Breaking Changes

No

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/examples/simple-chat-client/Cargo.toml
+++ b/examples/simple-chat-client/Cargo.toml
@@ -17,4 +17,4 @@ rmcp = { git = "https://github.com/modelcontextprotocol/rust-sdk", features = [
     "client",
     "transport-child-process",
     "transport-sse",
-], no-default-features = true }
+] }
--- a/examples/servers/src/common/counter.rs
+++ b/examples/servers/src/common/counter.rs
@@ -1,8 +1,7 @@
 use std::sync::Arc;
 
 use rmcp::{
-    Error as McpError, RoleServer, ServerHandler, const_string, model::*, schemars,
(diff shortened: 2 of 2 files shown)


examples/simple-chat-client/Cargo.toml
```diff
@@ -17,4 +17,4 @@ rmcp = { git = "https://github.com/modelcontextprotocol/rust-sdk", features = [
     "client",
     "transport-child-process",
     "transport-sse",
-], no-default-features = true }
```
rikhuijzer, line 20: This had no effect like in https://github.com/modelcontextprotocol/rust-sdk/pull/112. Setting `default-features = false` does not compile so I just removed it.

jokemanfire: Sorry , I didn't meet this problem, could you pls provide more detail?

rikhuijzer: Uhm yes I'll try. The setting `no-default-features` does not exists, so `no-default-features = true` is ignored by Cargo. The setting should be `default-features = true`, but if you set that the code does not compile. So that's why I removed it.

rikhuijzer: The confusion might be because of `cargo build --no-default-features`. That is a valid option.

PR #127, opened 2025-04-11
feat: client-extensions

EItanya:
Add extensions to the client macros

## Motivation and Context

I am developing a client and I need to be able to pass extensions through my requests.

## Breaking Changes

This breaks the current macro code. I tried thinking of a way to do it without breaking, but I couldn't come up with any good ideas. Maybe a request builder where we can call `with_extensions`. Open to ideas here.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -193,12 +193,16 @@ macro_rules! method {
         }
     };
     (peer_req $method:ident $Req:ident($Param: ident) => $Resp: ident ) => {
-        pub async fn $method(&self, params: $Param) -> Result<$Resp, ServiceError> {
+        pub async fn $method(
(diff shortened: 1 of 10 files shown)


salman1993:
i couldn't find any reference to extensions in [MCP spec](https://modelcontextprotocol.io/specification/2025-03-26). don't think it should be in the sdk

4t145:
Sorry for late response. My opinion is to keep the basic api simple. If `extension` is a parameter here, why not other options like timeout or other things else. Maybe we could add extensions by calling `send_request`, or we could add a helper method to help to specific response type. Maybe we can have a trait to convert different request into json rpc resquest and give it response type.

EItanya:
Sounds good, I will close for now

PR #128, opened 2025-04-12
feat: add constructors to `RequestContext`

rikhuijzer:
Add some default constructors to `RequestContext` which make it easier to work with the object.

## Motivation and Context

It makes the tests in `test_message_protocol` easier to read and understand and also allows other programmers to more easily work with the objects. The constructors are essentially a kind of `Default` constructor.

## How Has This Been Tested?

It's part of the tests.

## Breaking Changes

No.

## Types of changes

- [ ] Documentation update

## Checklist


jokemanfire (approved):

jokemanfire (changes requested):
How about just change to ..default::Default() in struct ,  that we needn't add two function

rikhuijzer:
> How about just change to ..default::Default() in struct , that we needn't add two function

I'm don't think that's possible because `Peer` does not have a `Default` implementation.

crates/rmcp/tests/test_message_protocol.rs
```diff
@@ -66,13 +65,7 @@ async fn test_context_inclusion_integration() -> anyhow::Result<()> {
...
         .handle_request(
             request.clone(),
-            RequestContext {
-                peer: client.peer().clone(),
-                ct: CancellationToken::new(),
-                id: NumberOrString::Number(1),
-                meta: Default::default(),
-                extensions: Default::default(),
-            },
+            RequestContext::from_peer(client.peer().clone()),
```
jokemanfire, line 68: I mean like this
RequestContext {
    peer: client.peer().clone(),
    id: NumberOrString::Number(1),
    ..Default::default()
}

crates/rmcp/src/service.rs
```diff
@@ -476,6 +476,23 @@ pub struct RequestContext<R: ServiceRole> {
     pub peer: Peer<R>,
 }
 
+impl<R: ServiceRole> RequestContext<R> {
```
jokemanfire, line 479: add these two func , i think usfuless

crates/rmcp/tests/test_message_protocol.rs
rikhuijzer: Yes I understood what you meant. And as I said, that is not possible because `Peer` does not have a `Default` implementation. This is what happens if you try to add `Default`:

image

and for `RequestId` it's simple to add a `Default` implementation but not for `Peer` since `Peer` needs to know a `request_id_provider` and `peer_info`.

jokemanfire: It needn't use default derive for ReuestContext , just do like this. Simplify the construction.
image

rikhuijzer: Does that compile for you? 

image

jokemanfire: Sorry for that , I've add default trait in my dev branch

jokemanfire: + impl<R: ServiceRole> Default for RequestContext<R> {
    fn default() -> Self {
        Self {
            ct: CancellationToken::new(),
            id: NumberOrString::Number(0),
            meta: Meta::default(),
            extensions: Extensions::default(),
            peer: Peer::new(Arc::new(AtomicU32RequestIdProvider::default()), R::PeerInfo::default()).0,
        }
    }
}

+  type PeerInfo: TransferObject + Default;

jokemanfire: may be impl default will be greater

rikhuijzer: > `Peer::new(Arc::new(AtomicU32RequestIdProvider::default()), R::PeerInfo::default()).0`

Yes exactly and this I think is not a good idea. This is a sort of “useless” default because it cannot be used. There is nobody at the other end of the connection. I think it would be better then to make `peer` an `Option`

rikhuijzer: On second thought, I have no time for these kinds of discussions. I'll close the PR

Issue #55, opened 2025-03-28
Support Protocol Revision 2025-03-26


4t145:
Here's a todo list:

- [ ] New [schema](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts) and corresponded service logic.
- [ ] [Authorazation](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/authorization/) support.
- [ ] [Streamable HTTP transport](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#streamable-http) client and server support.

4t145:
Firstly, I want to remove the `Message` structure in model to avoid unnecessary memory allocation in `into_message`, and also simplify the whole process. This will not be a breaking change because user don't really use this structure.

dbettin:
@4t145 I would love to help out. How can I assist in supporting the latest revision?

4t145:
@dbettin I plan to merge #84 firstly and then implement streamable http transport and authorization. Works in #84 are almost done. If you have a good implementation for streamable http transport, you can post it so we can have a discussion on it.

jokemanfire:
I will add the oauth support.

tuananh:
is there anyone working on streamable http transport?

jokemanfire:
> is there anyone working on streamable http transport?有人在做可流式传输的 HTTP 传输吗？
No one should be doing it now. If you can take on this job, thank you very much

4t145:
@tuananh I am working on streamable http transport here: https://github.com/4t145/rust-sdk/tree/feature-streamable-http

jokemanfire:
I think the work after your httpstream ,and improve the document . If we can publish the first release ?@4t145

4t145:
@jokemanfire sounds good!

4t145:
I am not going to support batch jsonrpc in streamable http client since they plan to delete this part from latest protocol specification.
(https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416)

fbjork:
Any plans on supporting OAuth for servers?

4t145:
@fbjork @jokemanfire done some work in [#130](https://github.com/modelcontextprotocol/rust-sdk/pull/130)

fbjork:
[...]

PR #130, opened 2025-04-12
chore: try to add oauth2 support

jokemanfire:
add oauth2 client support

## Motivation and Context
oauth2 is in mcp's sepc file

## How Has This Been Tested?
it

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1,0 +1,428 @@
+use std::sync::Arc;
+use std::time::Duration;
+use futures::future::BoxFuture;
+use oauth2::basic::BasicTokenType;
+use oauth2::{
+    AuthorizationCode, ClientId, ClientSecret, CsrfToken, PkceCodeChallenge, PkceCodeVerifier,
+    RedirectUrl, TokenResponse, Scope, AuthUrl, TokenUrl, RefreshToken, 
+    StandardTokenResponse, TokenType, AccessToken, EmptyExtraTokenFields,
+    basic::BasicClient, reqwest::http_client, RefreshTokenRequest, AuthorizationRequest
+};
+use reqwest::{Client as HttpClient, header::AUTHORIZATION, StatusCode, Url, IntoUrl};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use tokio::sync::{Mutex, RwLock};
+use tokio::time::{self, Instant};
+
+/// 错误定义
+#[derive(Debug, Error)]
+pub enum AuthError {
+    #[error("OAuth authorization required")]
+    AuthorizationRequired,
+    
+    #[error("OAuth authorization failed: {0}")]
+    AuthorizationFailed(String),
+    
+    #[error("OAuth token exchange failed: {0}")]
+    TokenExchangeFailed(String),
+    
+    #[error("OAuth token refresh failed: {0}")]
+    TokenRefreshFailed(String),
+    
+    #[error("HTTP error: {0}")]
+    HttpError(#[from] reqwest::Error),
+    
+    #[error("OAuth error: {0}")]
+    OAuthError(String),
+    
+    #[error("Metadata error: {0}")]
+    MetadataError(String),
+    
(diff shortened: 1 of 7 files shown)


rikhuijzer:
Would it maybe be possible to add a test?

jokemanfire:
> Would it maybe be possible to add a test?可能可以添加一个测试吗？

I will add that

jokemanfire:
I think that I need to add a third oauth server to verify this process.

jokemanfire:
Basic functions can be reviewed now. :)

crates/rmcp/src/transport/auth.rs
```diff
@@ -0,0 +1,558 @@
...
+    pub client_secret: String,
+    pub client_name: String,
+    pub redirect_uris: Vec<String>,
+}
+
+impl AuthorizationManager {
+    /// create new auth manager
+    pub async fn new<U: IntoUrl>(base_url: U) -> Result<Self, AuthError> {
+        let base_url = base_url.into_url()?;
+        let http_client = HttpClient::builder()
```
4t145, line 136: Is it possible to allow passing a custom http client here?

crates/rmcp/src/transport/auth.rs
```diff
@@ -0,0 +1,558 @@
...
+/// oauth2 auth manager
+pub struct AuthorizationManager {
+    http_client: HttpClient,
+    metadata: Option<AuthorizationMetadata>,
+    oauth_client: Option<OAuthClient>,
+    credentials: RwLock<Option<StandardTokenResponse<EmptyExtraTokenFields, BasicTokenType>>>,
+    pkce_verifier: RwLock<Option<PkceCodeVerifier>>,
+    base_url: Url,
+}
+
```
4t145, line 114: We have rw lock here, but using it with a mutex wrapper, so I guess one of them is unnecessary?

crates/rmcp/src/transport/auth.rs
```diff
@@ -0,0 +1,558 @@
...
+        Ok(token_result)
+    }
+
+    /// get access token, if expired, refresh it automatically
+    pub async fn get_access_token(&self) -> Result<String, AuthError> {
+        let credentials = self.credentials.read().await;
+
+        if let Some(creds) = credentials.as_ref() {
+            // check if the token is expired
+            if let Some(expires_in) = creds.expires_in() {
```
4t145: I am confusing that, since the `expire_in` seems wouldn't change automatically since manager received response, are these codes really can be triggered? Should we store the expire time in anthor field, maybe something like `expire_at`?

examples/clients/src/oauth_client.rs
```diff
@@ -0,0 +1,272 @@
...
+) -> Html<String> {
+    tracing::info!("Received callback with code: {}", params.code);
+
+    // Send the code to the main thread
+    if let Some(sender) = state.code_receiver.lock().await.take() {
+        let _ = sender.send(params.code);
+    }
+
+    // Return success page
+    Html(
```
4t145: We can wrtie html in a seperated file, and use include_str! macro to embedding it into code. As well as other html response.

crates/rmcp/src/transport/auth.rs
jokemanfire: This is because at that time, it was considered that the authmanager would be acquired by the session and had some operations that could be used externally. Indeed, this is not very good. I am planning to encapsulate a state machine, which I believe would be better.

crates/rmcp/src/transport/auth.rs
jokemanfire: Sure

crates/rmcp/src/transport/auth.rs
jokemanfire: You're right ,I was ignoring this place

jokemanfire:
@4t145  I 've changed

4t145 (approved):
Nice work

jokemanfire:
#55

PR #136, opened 2025-04-16
fix: fix resource leak when `RunningService` is dropped.

Paul-E:
## Motivation and Context

We found in our service that when we were creating clients, they were not cleaning up their resources when they were dropped. Clients keeping their connections open when dropped does not follow the principle of least astonishment and ended up leaking resource. Where possible users can call `cancel` on their `RunningService`, but it is possible for an early exit ( eg by returning an error with `?`) to lead to resource leaks.

These changes add a `drop` implementation to `RunningService` so that when a service is dropped it cleans up its resources.

## Breaking Changes

This is a breaking change, as the change returns the error returned by `waiting` and `cancel` to `ServiceError`

## Types of changes

- [ ] Documentation update

## Checklist

- [] New and existing tests pass locally

(python tests are not passing, probably due to local dev env)

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -21,7 +21,7 @@ mod server;
 pub use server::*;
 #[cfg(feature = "tower")]
 mod tower;
-use tokio_util::sync::CancellationToken;
+use tokio_util::sync::{CancellationToken, DropGuard};
 #[cfg(feature = "tower")]
 pub use tower::*;
 use tracing::instrument;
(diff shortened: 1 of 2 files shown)


jokemanfire:
thanks  but fix this CI pls , and CC @4t145

jokemanfire (dismissed):
LGTM , and take a look @4t145 ,when you free.

4t145:
Sorry for late response.

I preffer to use [`drop_guard`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.drop_guard). How do you think about it?

jokemanfire:
I haven't used this interface before. Looks like using drop_guard can prevent API changes? Do you have time to take a look ？ @Paul-E

Paul-E:
> I haven't used this interface before. Looks like using drop_guard can prevent API changes? Do you have time to take a look ？ @Paul-E

Ah yes good catch, that looks much simpler.

Paul-E:
I've updated the PR. I think this still represents a breaking API change, mostly impacting servers. Some folks might be surprised that their server shuts down after their setup code exits its function.

4t145:
It looks good now

4t145 (approved):

PR #141, opened 2025-04-17
feat(rmcp-macro): generate description from docs

ssddOnTop:
Enhance the tool macro to automatically extract descriptions from documentation comments when no explicit description is provided, reducing redundancy in code.

## Motivation and Context
Currently, developers need to duplicate information by writing documentation comments and then providing the same content in the description attribute. This change allows the macro to reuse the documentation comments as the description, making the code more maintainable and DRY.

## How Has This Been Tested?
I've added unit tests and tested it in a dummy project.

## Breaking Changes
It has full backward compatiblity. It's not a breaking change, but it's an additional feature.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
It matches the implementation of Python SDK.

### DIFF
--- a/crates/rmcp-macros/src/tool.rs
+++ b/crates/rmcp-macros/src/tool.rs
@@ -405,10 +405,36 @@ pub(crate) fn tool_fn_item(attr: TokenStream, mut input_fn: ItemFn) -> syn::Resu
     // generate get tool attr function
     let tool_attr_fn = {
         let description = if let Some(expr) = tool_macro_attrs.fn_item.description {
+            // Use explicitly provided description if available
             expr
         } else {
-            parse_quote! {
-                ""
+            // Try to extract documentation comments
+            let mut doc_content = String::new();
+
+            for attr in &input_fn.attrs {
+                if attr.path().is_ident("doc") {
+                    if let Ok(lit) = attr.parse_args::<syn::LitStr>() {
+                        let doc_line = lit.value();
+                        if !doc_content.is_empty() {
+                            doc_content.push_str("\n");
+                        }
+                        doc_content.push_str(doc_line.trim());
+                    }
+                }
+            }
+
+            if !doc_content.is_empty() {
+                // Use documentation comments if available
+                let doc_str = doc_content.trim().to_string();
+                // Convert the string to a string literal expression
+                parse_quote! {
+                    #doc_str
+                }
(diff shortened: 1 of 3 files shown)


crates/rmcp-macros/src/tool.rs
```diff
@@ -405,10 +405,30 @@ pub(crate) fn tool_fn_item(attr: TokenStream, mut input_fn: ItemFn) -> syn::Resu
     // generate get tool attr function
     let tool_attr_fn = {
         let description = if let Some(expr) = tool_macro_attrs.fn_item.description {
+            // Use explicitly provided description if available
             expr
         } else {
+            // Try to extract documentation comments
+            let mut doc_content = String::new();
+
+            for attr in &input_fn.attrs {
```
jokemanfire: the code if let may be much.
how about like this:
```
fn extract_doc_line(attr: &Attribute) -> Option<String> {
    let syn::Meta::NameValue(name_value) = &attr.meta else { return None };
    if !name_value.path().is_ident("doc") {
        return None;
    }
    
    let syn::Expr::Lit(expr_lit) = &name_value.value else { return None };
    let syn::Lit::Str(lit_str) = &expr_lit.lit else { return None };
    
    Some(lit_str.value())
}
```  
and call it  like
```
for attr in &input_fn.attrs {
    if let Some(doc_line) = extract_doc_line(attr) {
        if !doc_content.is_empty() {
            doc_content.push('\n');
        }
        doc_content.push_str(doc_line.trim());
    }
}
```

ssddOnTop: done

jokemanfire (approved):
LGTM

crates/rmcp-macros/src/tool.rs
```diff
@@ -296,6 +296,23 @@ pub(crate) fn tool_impl_item(attr: TokenStream, mut input: ItemImpl) -> syn::Res
     })
 }
 
+fn extract_doc_line(attr: &syn::Attribute) -> Option<String> {
```
jokemanfire, line 299: I don't think multi-layer nesting is very good. Can you modify it here?

ssddOnTop: fixed

jokemanfire (approved):

PR #148, opened 2025-04-28
Improve error logging in SSE server message handling

erenatas:
Passes the actual error to error log

## Motivation and Context
Add detailed error information to the tracing::error! call when message sending fails, making it easier to diagnose communication issues between clients and the server.

## How Has This Been Tested?
Tested in a local project, for some reason I can reproduce this error each time I change resources with `npx @modelcontextprotocol/inspector`

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/sse_server.rs
+++ b/crates/rmcp/src/transport/sse_server.rs
@@ -79,6 +79,7 @@ async fn post_event_handler(
             .ok_or(StatusCode::NOT_FOUND)?
             .clone()
     };
+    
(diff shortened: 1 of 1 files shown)


4t145:
The content of the error here is just the unsent message so I don't think we should log it out.

erenatas:
> The content of the error here is just the unsent message so I don't think we should log it out.

Reasonable, will close it. Thanks for quick look!

PR #152, opened 2025-05-02
feat(transport): support streamable http server

4t145:
Support streamable http server:

There are two parts
## 1. Session
The reusable streamable http session management codes.

## 2. Axum server
An implementation in axum.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

I fixed the log level issue in #83

### DIFF
--- a/crates/rmcp/src/transport/sse_server.rs
+++ b/crates/rmcp/src/transport/sse_server.rs
@@ -57,8 +57,7 @@ impl App {
 }
 
 fn session_id() -> SessionId {
-    let id = format!("{:016x}", rand::random::<u128>());
-    Arc::from(id)
+    uuid::Uuid::new_v4().to_string().into()
 }
 
 #[derive(Debug, serde::Deserialize)]
--- a/crates/rmcp/src/transport/sse_server.rs
+++ b/crates/rmcp/src/transport/sse_server.rs
@@ -19,13 +19,12 @@ use crate::{
     RoleServer, Service,
     model::ClientJsonRpcMessage,
     service::{RxJsonRpcMessage, TxJsonRpcMessage},
+    transport::common::axum::{DEFAULT_AUTO_PING_INTERVAL, SessionId, session_id},
 };
-type SessionId = Arc<str>;
+
 type TxStore =
     Arc<tokio::sync::RwLock<HashMap<SessionId, tokio::sync::mpsc::Sender<ClientJsonRpcMessage>>>>;
 pub type TransportReceiver = ReceiverStream<RxJsonRpcMessage<RoleServer>>;
-
-const DEFAULT_AUTO_PING_INTERVAL: Duration = Duration::from_secs(15);
 
 #[derive(Clone)]
 struct App {
@@ -54,10 +53,6 @@ impl App {
             transport_rx,
         )
     }
-}
-
-fn session_id() -> SessionId {
-    uuid::Uuid::new_v4().to_string().into()
 }
 
 #[derive(Debug, serde::Deserialize)]
--- a/crates/rmcp/src/transport/streamable_http_server/axum.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/axum.rs
@@ -242,7 +242,7 @@ async fn delete_handler(
 }
 
 #[derive(Debug, Clone)]
-pub struct SseServerConfig {
+pub struct StreamableHttpServerConfig {
     pub bind: SocketAddr,
     pub path: String,
     pub ct: CancellationToken,
@@ -252,12 +252,12 @@ pub struct SseServerConfig {
 #[derive(Debug)]
 pub struct StreamableHttpServer {
     transport_rx: tokio::sync::mpsc::UnboundedReceiver<SessionTransport>,
-    pub config: SseServerConfig,
+    pub config: StreamableHttpServerConfig,
 }
 
 impl StreamableHttpServer {
     pub async fn serve(bind: SocketAddr) -> io::Result<Self> {
-        Self::serve_with_config(SseServerConfig {
+        Self::serve_with_config(StreamableHttpServerConfig {
             bind,
             path: "/".to_string(),
             ct: CancellationToken::new(),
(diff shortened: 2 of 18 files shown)


jokemanfire:
Sorry, I'm currently on a trip and may need to be late to see it.

crates/rmcp/src/transport/sse_server.rs
```diff
@@ -57,8 +57,7 @@ impl App {
 }
 
 fn session_id() -> SessionId {
```
jokemanfire: Seesion_id looks like a frequently reused code.How about take out code to reduce reuse.

crates/rmcp/src/transport/streamable_http_server/axum.rs
```diff
@@ -0,0 +1,331 @@
...
+    EventId, HEADER_LAST_EVENT_ID, Session, SessionId, SessionTransport,
+    StreamableHttpMessageReceiver,
+};
+use crate::{
+    RoleServer, Service, model::ClientJsonRpcMessage,
+    transport::streamable_http_server::session::HEADER_SESSION_ID,
+};
+type SessionManager = Arc<tokio::sync::RwLock<HashMap<SessionId, Session>>>;
+
+const DEFAULT_AUTO_PING_INTERVAL: Duration = Duration::from_secs(15);
```
jokemanfire: I remember this constant seems to be reused in some places

crates/rmcp/src/transport/streamable_http_server/session.rs
```diff
@@ -0,0 +1,759 @@
...
+        let index = self.cache.back().map_or(0, |m| m.event_id.index + 1);
+        let event_id = EventId {
+            http_request_id: self.http_request_id,
+            index,
+        };
+        let message = ServerSessionMessage {
+            event_id: event_id.clone(),
+            message: Arc::new(message),
+        };
+        if self.cache.len() >= self.capacity {
```
jokemanfire, line 117: how about use BoundedDeque trait , I'd prefer this:
```
trait BoundedDeque<T> {
    fn push_bounded(&mut self, item: T, max_len: usize);
}

impl<T> BoundedDeque<T> for VecDeque<T> {
    fn push_bounded(&mut self, item: T, max_len: usize) {
        if self.len() >= max_len {
            self.pop_front();
        }
        self.push_back(item);
    }
}
```

crates/rmcp/src/transport/streamable_http_server/session.rs
```diff
@@ -0,0 +1,759 @@
...
+            self.cache.pop_front();
+            self.cache.push_back(message.clone());
+        } else {
+            self.cache.push_back(message.clone());
+        }
+        let _ = self.tx.send(message).await;
```
jokemanfire: If channel close , can we print a warn level log here?

crates/rmcp/src/transport/streamable_http_server/session.rs
```diff
@@ -0,0 +1,759 @@
...
+    #[instrument(name = "streamable_http_session", skip_all, fields(id = self.id.as_ref()))]
+    pub async fn run(mut self, ct: CancellationToken) -> SessionQuitReason {
+        // waiting for initialize request
+        let Some(evt) = self.event_rx.recv().await else {
+            return SessionQuitReason::ServiceTerminated;
+        };
+        let SessionEvent::InitializeRequest { request, responder } = evt else {
+            return SessionQuitReason::ExpectInitializeRequest;
+        };
+        let send_result = self.send_to_service(request).await;
```
jokemanfire, line 467: map_err?

crates/rmcp/src/transport/sse_server.rs
4t145: Yeah, I'd like to extract the common part of axum related code later.

crates/rmcp/src/transport/streamable_http_server/session.rs
4t145: Will we support different cache in the future? If so, I think we should expose the interface, other wise I don't think we need another trait.

crates/rmcp/src/transport/streamable_http_server/session.rs
4t145: the return type here is not a `Result` so I cannot throw error through `map_err` + `?`

crates/rmcp/src/transport/streamable_http_server/session.rs
4t145: Sure, I'll add more log.

4t145:
@jokemanfire I think It's ready for rewiew now.

jokemanfire:
I'll take a look later. Work has been a bit heavy lately

crates/rmcp/src/transport/streamable_http_server/session.rs
```diff
@@ -0,0 +1,767 @@
...
+    ) {
+        if let ClientNotification::CancelledNotification(n) = &notification.notification {
+            let request_id = n.params.request_id.clone();
+            let resource = ResourceKey::McpRequestId(request_id);
+            self.unregister_resource(&resource);
+        }
+    }
+    fn next_http_request_id(&mut self) -> HttpRequestId {
+        let id = self.next_http_request_id;
+        self.next_http_request_id = self.next_http_request_id.saturating_add_signed(1);
```
jokemanfire: all is u64 ,use saturating_add ? and if the id is max,it will always return max , if there's need a resycle, (Although the possibility of reaching the maximum U64 is relatively low, there are rarely services that run for such a long time)

crates/rmcp/src/transport/streamable_http_server/axum.rs
```diff
@@ -0,0 +1,328 @@
...
+) -> Result<StatusCode, Response> {
+    if let Some(session_id) = header_map.get(HEADER_SESSION_ID) {
+        let session_id = session_id
+            .to_str()
+            .map_err(|e| (StatusCode::BAD_REQUEST, e.to_string()).into_response())?;
+        let mut sm = app.session_manager.write().await;
+        let session = sm
+            .remove(session_id)
+            .ok_or((StatusCode::NOT_FOUND, "session not found").into_response())?;
+        session.handle();
```
jokemanfire: it should be cancel?

crates/rmcp/src/transport/streamable_http_server/session.rs
```diff
@@ -0,0 +1,767 @@
...
+                } else {
+                    return Err(SessionError::ChannelClosed(Some(id)));
+                }
+            }
+            OutboundChannel::Common => self.common.send(message).await,
+        }
+        Ok(())
+    }
+    async fn resume(
+        &mut self,
```
jokemanfire, line 389: If there is data in the previous channel, will it be discarded？

jokemanfire (commented):
Only these questions

crates/rmcp/src/transport/streamable_http_server/session.rs
4t145: Yeah I intend to use wrapping add. It's a mistake.

crates/rmcp/src/transport/streamable_http_server/session.rs
4t145: Data in cache will be resend when calling `sync`, and data in mpsc channel should be received since we have awaited every `send()` call.

crates/rmcp/src/transport/streamable_http_server/axum.rs
4t145: Yeah I intended to call cancel here.

jokemanfire (approved):
LGTM

4t145:
@jokemanfire Wait a minute, just fixed some naming issue. I forgot to rename sse to streamable http when migrate codes.

jokemanfire (approved):

PR #156, opened 2025-05-05
Cleanup zombie processes for child process client

humbertoyusta:
Ensure child processes kill all subprocesses on drop (for transport-child-process)

## Motivation and Context

Starting a tokio child processes, in several mcp servers, like using `uv some-mcp-server` in MacOS, will create subprocesses that will not be cleaned when the first process is killed with the `.kill_on_drop(true)`

Using a JobObject in Windows and a Process Group in Unix, killing the child process will also clean all subprocesses, avoiding resource leaking

## How Has This Been Tested?
We (refact.ai) use this library for integrating mcp servers with our client, we recently switched from another library, so it's not in production yet.

Scenarios about starting servers like with `uv` were tested to avoid resource leaking, also fakeish scenarios like `sh -c "npx some-mcp-server"` were tested, causing subprocess leaking before, not now

## Breaking Changes
TokioChildProcess::new(command) takes now tokio::process::Command entirely, instead of &mut tokio::process::Command
* The reasion for this, is that the wrapping library used for process group and job object, requires taking the whole command instead of &mut ref
* Example tests have been updated accordingly

## Types of changes

- [ ] Documentation update

## Checklist

- [?] My code follows the repository's style guidelines
- [x] I have added appropriate error handling (not needed)


4t145:
Thank you for PR, I will have a look on this!

4t145:
Looks good!

Is there a way to provide a chained call api to create a `cmd`? Just want to give out a more easy to use interface.

And it will be great if you can update the document like `README`.

humbertoyusta:
Thanks for the review,

I updated `README` and the other example that was missing.

I couldn't find a way to provide an inline chained way, due to the fact that most std/tokio::process::Command methods like args(), envs() take and return &mut Command instead of Command, so if you chain it in one line, there's no way to get the Command again.

I checked what [process-wrap](https://crates.io/crates/process-wrap) does for this, but they just use command in non-inline way, or create their own structs directly.

TokioChildProcess::new({
  let mut cmd = tokio::process::Command::new("npx");
  cmd.arg("mcp-server-git")
  cmd
})

will workfor inline way, but this is just the same as doing it before, I coulnd't find a clean way to chain it inline

jokemanfire:
How about add a event listener , ant use wait_pid to recycle the children process ?

humbertoyusta:
One option for inline could be to modify examples to do like:

```rust
TokioChildProcess::new(tokio::process::Command::new("node").map(
    |cmd| { cmd.arg("tests/test_with_js/client.js"); cmd }
))
```

Another could be for people who use tap crate, using tap_mut, but it's more or less the same, we could provide that trait to command, but it would need to be imported and it's not much less verbose

```rust
use rmcp::transport::child_process::CommandTapMut;

TokioChildProcess::new(tokio::process::Command::new("node").tap_mut(
    |cmd| { cmd.arg("tests/test_with_js/client.js"); }
))
```

Or we could do a wrapper to command like proposed before, but there will be several Command structs then, not sure which is the best approach.

4t145:
@humbertoyusta The first approach looks great. I am okay with it.

humbertoyusta:
Actually `.map()` didn't work, I thought Command::new() was giving result but it's not, so I needed to add a trait for `tokio::process::Command`.

I still think it's better than a full Command wrapper, but if you think otherwise let me know.

4t145:
@humbertoyusta I am ok with the ext trait. 

Looks there's confict and ci didn't pass. So could you merge it with the main branch, for I just made a lot of changes, and fix the formatting and test ci? Thanks for your patience!

4t145:
It's okey with the ci message, I can rewrite it when merge.

4t145 (approved):

4t145:
And thanks for your work!

PR #157, opened 2025-05-06
feat: several oauth improvements: support reusing credentials. capture additional fields instead of failing

the-rooster:
## Motivation and Context

I am writing an application (oneserver.ai) that uses this sdk for connecting to remote servers, and encountered a few issues with the new OAuth work:
- Currently I cannot use reuse OAuth credentials conveniently (must do the whole flow each time)
- I encounter decoding errors on dynamic client registration responses + OAuth metadata discovery due to fields that are not included in the serde struct definitions

This PR fixes both of these:
- OAuthState now has two new methods:
get_credentials: returns client_id and OAuthTokenResponse
set_credentials: sets the client_id and OAuthTokenResponse, and moves the state machine into the Authorized state for use in the transport
- Added an `additional_fields` field to both the AuthorizationMetadata struct and the ClientRegistrationResponse struct which will contain all additional fields available in the response. This lets us safely ignore fields that we don't need for our flow.
- Additionally, client_secret is now an optionally returned value in ClientRegistrationResponse (some implementations of Oauth2 dynamic client registration do not return this value it seems)

## How Has This Been Tested?

I tested this from within my application oneserver.ai against our hosted hello world remote MCP server at https://mcp.oneserver.ai and it works perfectly :)

You will notice that the current version fails when run against this server, and the new version succeeds.

## Breaking Changes

Will break some implementations that use the ClientRegistrationResponse and AuthorizationMetadata structs, but considering this oauth work is very new, I think this is okay.

## Types of changes

- [ ] Documentation update

## Checklist

- [] I have added or updated documentation as needed

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     sync::Arc,
     time::{Duration, Instant},
 };
@@ -70,6 +71,9 @@ pub struct AuthorizationMetadata {
     pub issuer: Option<String>,
     pub jwks_uri: Option<String>,
     pub scopes_supported: Option<Vec<String>>,
+    // allow additional fields
+    #[serde(flatten)]
+    pub additional_fields: HashMap<String, serde_json::Value>,
 }
 
 /// oauth2 client config
@@ -100,6 +104,7 @@ type OAuthClient = oauth2::Client<
     oauth2::EndpointNotSet,
     oauth2::EndpointSet,
 >;
+type Credentials = (String, Option<OAuthTokenResponse>);
 
 /// oauth2 auth manager
 pub struct AuthorizationManager {
@@ -124,9 +129,12 @@ pub struct ClientRegistrationRequest {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ClientRegistrationResponse {
     pub client_id: String,
-    pub client_secret: String,
+    pub client_secret: Option<String>,
     pub client_name: String,
     pub redirect_uris: Vec<String>,
+    // allow additional fields
+    #[serde(flatten)]
+    pub additional_fields: HashMap<String, serde_json::Value>,
 }
 
 impl AuthorizationManager {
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -590,9 +632,52 @@ impl OAuthState {
         if let Some(client) = client {
             manager.with_client(client)?;
         }
+
         Ok(OAuthState::Unauthorized(manager))
     }
 
+    /// Get client_id and OAuth credentials
+    pub async fn get_credentials(&self) -> Result<(String, Option<OAuthTokenResponse>), AuthError> {
```
jokemanfire: match self {
        OAuthState::Unauthorized(manager)
        | OAuthState::Session(manager)
        | OAuthState::Authorized(manager) => manager.get_credentials().await,
        OAuthState::AuthorizedHttpClient(client) => 
            client.auth_manager.get_credentials().await,
    }

crates/rmcp/src/transport/auth.rs
```diff
@@ -513,6 +549,12 @@ impl AuthorizationSession {
         })
     }
 
+
+    /// get client_id and credentials
+    pub async fn get_credentials(&self) -> Result<(String, Option<OAuthTokenResponse>), AuthError> {
```
jokemanfire: use alias
`type Credentials = (String, Option<OAuthTokenResponse>);`

crates/rmcp/src/transport/auth.rs
the-rooster: Session uses AuthorizationSession, but I can combine the two that use manager!

the-rooster: done

jokemanfire:
test_with_js_client test has some problems , it should be fixed firstly @4t145

jokemanfire:
LGTM , but just wait for the CI fixed.

jokemanfire (approved):

4t145:
I guess you may merge the lasted code in main branch so we can test it in a test fixed version?

the-rooster:
merged

PR #165, opened 2025-05-08
Fix SSE Transport Memory Leak

lnbc1QWFyb24:
When running the SSE transport in production, we noticed a massive memory leak. We have a Kubernetes health check process that was hitting the SSE endpoint periodically and then just disconnecting which was causing the memory leak. Digging in to the current implementation, there is no disconnection handling in the case of an abrupt disconnection, so this PR adds some cleanup when the connection drops.

## Motivation and Context
Removes a memory leak in the case that a client disconnects abruptly.

## How Has This Been Tested?
You can replicate the memory leak by simply making a GET request to the SSE endpoint and then disconnecting many times to see memory grow quite rapidly. I applied this fix to our production environment and the instance memory usage has been steady since.

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
There were two errors that were failing on main already, so these changes have not added any new errors.

### DIFF
--- a/crates/rmcp/src/transport/sse_server.rs
+++ b/crates/rmcp/src/transport/sse_server.rs
@@ -89,6 +89,8 @@ async fn sse_handler(
     use tokio_util::sync::PollSender;
     let (from_client_tx, from_client_rx) = tokio::sync::mpsc::channel(64);
     let (to_client_tx, to_client_rx) = tokio::sync::mpsc::channel(64);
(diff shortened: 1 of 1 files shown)


4t145:
Good point, would it better to release the resource when `SseServerTransport` drop?
```rust
impl Drop for SseServerTransport {
    // release the resource...
}

PR #167, opened 2025-05-09
refactor: Transport trait and worker transport, and streamable http client with those new features.

4t145:
related issue: #170, #55

# 1. Refactor transport trait
Now we have a real transport trait:
```rust
pub trait Transport<R>: Send
where
    R: ServiceRole,
{
    type Error;
    fn send(
        &mut self,
        item: TxJsonRpcMessage<R>,
    ) -> impl Future<Output = Result<(), Self::Error>> + Send + 'static;
    fn receive(&mut self) -> impl Future<Output = Option<RxJsonRpcMessage<R>>> + Send + '_;
    fn close(&mut self) -> impl Future<Output = Result<(), Self::Error>> + Send;
}
```

and this new transport trait is compatible with old Sink + Stream constrainment, with a thread shared sink.

```rust
pub struct SinkStreamTransport<Si, St> {
    stream: St,
    sink: Arc<Mutex<Si>>,
}

impl<Si, St> SinkStreamTransport<Si, St> {
    pub fn new(sink: Si, stream: St) -> Self {
        Self {
            stream,
            sink: Arc::new(Mutex::new(sink)),
        }
    }
}

impl<Role: ServiceRole, Si, St> Transport<Role> for SinkStreamTransport<Si, St>
where
    St: Send + Stream<Item = RxJsonRpcMessage<Role>> + Unpin,
    Si: Send + Sink<TxJsonRpcMessage<Role>> + Unpin + 'static,
{
    type Error = Si::Error;

    fn send(
        &mut self,
        item: TxJsonRpcMessage<Role>,
    ) -> impl Future<Output = Result<(), Self::Error>> + Send + 'static {
        use futures::SinkExt;
        let lock = self.sink.clone();
        async move {
            let mut write = lock.lock().await;
            write.send(item).await
        }
    }

    fn receive(&mut self) -> impl Future<Output = Option<RxJsonRpcMessage<Role>>> + '_ {
        use futures::StreamExt;
        self.stream.next()
    }

    async fn close(&mut self) -> Result<(), Self::Error> {
        Ok(())
    }
}
```

Based on those changes, we can make serve_inner loop not blocked by sending request.

2. New worker transport type:
For the case those we need to create a new worker task to run the transport, there's a new trait `Worker`, you can implement `Worker` and hand it over to `WorkerTransport`

```rust
pub trait Worker: Sized + Send + 'static {
    type Error: std::error::Error + Send + Sync + 'static;
    type Role: ServiceRole;
    fn err_closed() -> Self::Error;
    fn err_join(e: tokio::task::JoinError) -> Self::Error;
    fn run(
        self,
        context: WorkerContext<Self>,
    ) -> impl Future<Output = Result<(), WorkerQuitReason>> + Send;
    fn config(&self) -> WorkerConfig {
        WorkerConfig::default()
    }
}
```

And with this new trait, I refactor many existed implementation based on it.

3. Streamable http client
Implemented streamable http client transport based on new worker.

4. Auth client
Trying to make auth client reusable between different http client.

5. Add cfg-features for document
## Motivation and Context

#170

## Breaking Changes

1. Rename some feature:
sse -> sse_client (it's more concret)

2. Refactor the sse client transport, so do the construct methods.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/tests/test_with_js.rs
+++ b/crates/rmcp/tests/test_with_js.rs
@@ -128,7 +128,7 @@ async fn test_with_js_streamable_http_server() -> anyhow::Result<()> {
         .spawn()?;
 
     // waiting for server up
-    tokio::time::sleep(tokio::time::Duration::from_millis(200)).await;
+    tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
 
     let client = ().serve(transport).await?;
     let resources = client.list_all_resources().await?;


4t145:
@jokemanfire I'am planing to extract the common part of `reqwest ` client, so there may have the necessary to rewrite some part of current sse client, and then, both of streamable client and sse client can share the logic of Oauth2.

jokemanfire:
Thanks, That's well.

4t145:
I am going to refactor sse client in those aspect:
# Cleint trait:
1. move error from generic parameter to associated type (usually the error type should be a associated type)
2. use `impl Future` as return type
```rust
pub trait NeoSseClient: Clone + Send + Sync + 'static  {
    type Error: std::error::Error + Send + Sync + 'static;
    fn post_message(
        &self,
        uri: Arc<str>,
        message: ClientJsonRpcMessage,
    ) -> impl Future<Output = Result<(), SseTransportError<Self::Error>>>
    + Send
    + '_;
    fn get_stream(
        &self,
        uri: Arc<str>,
        last_event_id: Option<String>,
    ) -> impl Future<
        Output = Result<
            BoxedSseResponse,
            SseTransportError<Self::Error>,
        >,
    > + Send
    + '_;
}
```

# Move the connection logic to a worker
This can make code more maintainable and easier to understand.

Just like what I've done with this streamable http client in this PR.

# Implement client trait directly for raw reqwest client
1. The reason we have a wrapper at the first is we don't have a trait. And now we have a trait so we can remove the wrapper. (I guess) 
2. User can share the same configured client between diffent http based client transport types.

# Make OAuth2 a common wrapper for different client.

jokemanfire:
No problem.

4t145:
As messioned in #170, There are a lot of refactor. But the changes in examples is minimal. So it should not be impact to users.

4t145:
It'a big refactor, please check it again @jokemanfire.

jokemanfire:
I will check ,after this PR merger, I plan to include the example in the integration testing and further improve the documentation. I think we can release the first release version, but we need to agree that minor modifications should not include interface destructive changes, and only consider introducing it in the larger version. We also need to consider forward compatibility issues. What do you think？

4t145:
@jokemanfire 
I guess there could be possible breaking changes in transport part like web server support. And we still have a broken macro support, which could also be unstable for a while.

jokemanfire:
I will watch it after work tomorrow.

4t145:
I think it's ready for review now. But still need to add an example to show how to use streamable http client.

crates/rmcp/src/service.rs
```diff
@@ -546,29 +553,44 @@ where
     let serve_loop_ct = ct.child_token();
     let peer_return: Peer<R> = peer.clone();
     let handle = tokio::spawn(async move {
-        let (mut sink, mut stream) = transport.into_transport();
-        let mut sink = std::pin::pin!(sink);
-        let mut stream = std::pin::pin!(stream);
+        let mut transport = transport.into_transport();
+        // let mut sink = std::pin::pin!(sink);
```
jokemanfire: It could be removed?

crates/rmcp/src/transport/sse_client.rs
```diff
@@ -0,0 +1,234 @@
...
+                        Some(message) => Event::FromHandler(message),
+                        None => break WorkerQuitReason::HandlerTerminated,
+                    }
+                }
+
+            };
+            let next_sse = match event {
+                Event::Sse(Some(Ok(next_sse))) => next_sse,
+                Event::Sse(Some(Err(e))) => {
+                    tracing::warn!("sse stream error: {e}");
```
jokemanfire: I think it deserve a error level log.

crates/rmcp/src/transport/sse_client.rs
```diff
@@ -0,0 +1,234 @@
...
+                    let post_result = self
+                        .client
+                        .post_message(post_uri.clone(), send_request.message, None)
+                        .await;
+                    send_request.responder.send(post_result).ok();
+                    continue;
+                }
+            };
+            // set the retry interval
+            if let Some(server_retry_interval) = next_sse.retry {
```
jokemanfire: It is recommended to use an exponential backoff algorithm， like this
```
if let Some(server_retry_interval) = next_sse.retry {
    let server_duration = Duration::from_millis(server_retry_interval);
    if server_duration > retry_interval {
        retry_interval = server_duration;
    } else {
        retry_interval = retry_interval.saturating_mul(2).max(server_duration);
    }
    retry_interval = retry_interval.min(Duration::from_secs(30));
}
```

crates/rmcp/src/transport/worker.rs
```diff
@@ -0,0 +1,213 @@
...
+pub struct WorkerConfig {
+    pub name: Option<String>,
+    pub channel_buffer_capacity: usize,
+}
+
+impl Default for WorkerConfig {
+    fn default() -> Self {
+        Self {
+            name: None,
+            channel_buffer_capacity: 16,
```
jokemanfire, line 79: Is it so small?

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -0,0 +1,513 @@
...
+                }
+                _ = ct.cancelled() => {
+                    tracing::debug!("cancelled");
+                    break;
+                }
+            };
+            let next_sse = match event {
+                Some(Ok(next_sse)) => next_sse,
+                Some(Err(e)) => {
+                    tracing::warn!("sse stream error: {e}");
```
jokemanfire: error level log

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -0,0 +1,513 @@
...
+                            .client
+                            .get_stream(
+                                self.config.uri.clone(),
+                                session_id.clone(),
+                                last_event_id.clone(),
+                                None,
+                            )
+                            .await;
+                        retry_times += 1;
+                        match retry_result {
```
jokemanfire: it looks similar with sse-client, this code can be merge?

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -0,0 +1,513 @@
...
+        }
+    }
+}
+
+impl Default for StreamableHttpClientTransportConfig {
+    fn default() -> Self {
+        Self {
+            uri: "localhost".into(),
+            retry_config: SseRetryConfig::default(),
+            channel_buffer_capacity: 16,
```
jokemanfire, line 489: If it too small?

crates/rmcp/src/transport/streamable_http_server/axum.rs
```diff
@@ -223,18 +221,19 @@ async fn delete_handler(
...
+        };
+        session.close().await.map_err(|e| {
             (
                 StatusCode::INTERNAL_SERVER_ERROR,
                 format!("fail to cancel session {session_id}: tokio join error: {e}"),
             )
                 .into_response()
         })?;
-        tracing::info!(session_id, ?cancel_result, "session deleted");
+        tracing::info!(session_id, "session deleted");
```
jokemanfire: debug level may be better?

crates/rmcp/src/transport/worker.rs
```diff
@@ -0,0 +1,213 @@
...
+        context: impl Into<Cow<'static, str>>,
+    ) -> impl FnOnce(E) -> Self {
+        |e| Self::Fatal {
+            error: Cow::Owned(format!("{e}")),
+            context: context.into(),
+        }
+    }
+}
+// #[derive(Debug, Clone)]
+// pub enum WorkerQuitReason {
```
jokemanfire: remove?

jokemanfire (commented):
I'll have to keep watching tomorrow.

crates/rmcp/src/transport/sse_client.rs
4t145: As long as it's still trying to reconnect, it haven't reached an unrecoverable error, that's what I think.

crates/rmcp/src/transport/sse_client.rs
4t145: We can add a retry policy trait to support different policy.

crates/rmcp/src/transport/streamable_http_client.rs
4t145: I will try to extract a method out to serve sse-stream.

crates/rmcp/src/transport/streamable_http_server/axum.rs
4t145: Debug is better.

crates/rmcp/src/transport/streamable_http_client.rs
4t145: There are big different between those two loop, I decide to seperate an auto reconnectable sse stream as an async stream.

jokemanfire:
Actually, for me, things like sse_client, sse_client streamable_tttp_client streamable_tttp_derver require a higher level of work async_rw, sink_stream, and so on. As they are called, I tend to separate these two parts in the code organization of transport for a fresher code structure.  I think there shouldn't be any other issues.

4t145:
@jokemanfire Now it's ready for review again
1. I move the sse login to `common::client_side_sse`
2. Still remain the older interface for sse client, instead of use `WorkerTransport`
3. For worker channel buffer size, I think 16 is enough for most case. Sending to transport and handling by handler both won't block the service loop both now, so there are little chance to really pending on the await point.

jokemanfire (approved):

PR #184, opened 2025-05-17
Adding new Tool Call "Annotations" to tool macro

abhay-agarwal:
I added the new tool [annotation](https://modelcontextprotocol.io/docs/concepts/tools#tool-definition-structure) feature to the tool call macro. Currently it just parses the annotation field as json, so that one might potentially be able to simply copy and paste the field from elsewhere into the macro invocation. This might not be optimal for several reasons, one being that the signature must be written in camel case, with no indication to the user that a non-camel-cased value is incorrect, even if they looked at the expanded macro. A solution to this would be to create a wrapper over ToolAnnotations (or annotate ToolAnnotations itself) with `deny_unknown_fields`. However, this would also prevent the annotation macro from being future proof (currently if the annotation were to be deprecated, then we could silently ignore it).

Here's an example usage taken from the test I added

```rs
        #[tool(
            name = "direct-annotated-tool",
            annotations = {
                title: "Annotated Tool", 
                readOnlyHint: true
            }
        )]
        pub async fn direct_annotated_tool(&self, #[tool(param)] input: String) -> String {
            format!("Direct: {}", input)
        }
```

### DIFF
--- a/crates/rmcp-macros/Cargo.toml
+++ b/crates/rmcp-macros/Cargo.toml
@@ -18,6 +18,7 @@ proc-macro = true
 syn = {version = "2", features = ["full"]}
 quote = "1"
 proc-macro2 = "1"
+serde_json = "1.0"
 
 
 [features]
--- a/crates/rmcp-macros/src/tool.rs
+++ b/crates/rmcp-macros/src/tool.rs
@@ -2,10 +2,44 @@ use std::collections::HashSet;
 
 use proc_macro2::TokenStream;
 use quote::{ToTokens, quote};
+use serde_json::json;
 use syn::{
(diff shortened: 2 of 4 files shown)


4t145 (approved):
LGTM

I need to fix ci error introduced by lasted merge before merge this.

4t145:
Could you please merge the main branch and then we can run ci again? Also fix the commit message ci by the way.

abhay-agarwal:
> Could you please merge the main branch and then we can run ci again? Also fix the commit message ci by the way.

Squashed and reworded.

4t145:
Ci failed again but never mind, I will make it lowercase when merge.

PR #218, opened 2025-05-25
feat: add cancellation_token method to RunningService

lucasfernog-crabnebula:
## Motivation and Context

currently both RunningService::waiting and RunningService::cancel needs ownership of the RunningService instance
this makes it impossible to poll waiting() while other async task can cancel it

A cancellation_token() function is solves this, cloning the tokio_util::sync::CancellationToken
so we can dispatch a cancel from a separate task

## How Has This Been Tested?

- run a service
- call cancellation_token() to grab its cancellation token
- spawn a separate task to run RunningService::waiting() while other task calls cancel()

## Breaking Changes
No breaking change

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -421,9 +421,7 @@ impl<R: ServiceRole> Peer<R> {
 
 #[derive(Debug)]
 pub struct RunningService<R: ServiceRole, S: Service<R>> {
-    service: Arc<S>,
-    peer: Peer<R>,
-    handle: tokio::task::JoinHandle<QuitReason>,
+    handle: RunningServiceHandle<R, S>,
     /// cancellation token with drop guard
     dg: DropGuard,
 }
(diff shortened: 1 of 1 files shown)


4t145:
Perhaps we can simply add a `cancellation_token` field on it, while remain the grop guard? So user can clone the cancellation token.

lucasfernog-crabnebula:
but DropGuard can't be cloned, unless you want to change how cancellation is handled (if you clone a cancellation token and keep a copy in the RunningService the drop will never be triggered - and currently it also needs an owned value)

4t145:
I believe it still can be triggered, you can see the source code. https://docs.rs/tokio-util/0.7.15/src/tokio_util/sync/cancellation_token/guard.rs.html#22-26

lucasfernog-crabnebula:
@4t145 pushed

lucasfernog-crabnebula:
I explicitly added the cancellation token as part of the service struct because i don't want to call .disarm() on the drop guard - i want to preserve the "cancel on drop" functionality

4t145:
> I explicitly added the cancellation token as part of the service struct because i don't want to call .disarm() on the drop guard - i want to preserve the "cancel on drop" functionality
> 
> alternatively we could also implement drop on our own

That's exactly what I meant

4t145 (approved):
LGTM

PR #245, opened 2025-06-04
allow using a TokioCommandWrap for TokioChildProcess::new closes #243

### LINKED ISSUE #243: Stdio doesn't seem to respect cmd flags anymore
**Describe the bug**
The command line configuration is not respected by the TokioChildProcess.

I think this bug is introduced by https://github.com/modelcontextprotocol/rust-sdk/pull/156, which replace the tokio::process::Child with use process_wrap::tokio::{TokioChildWrapper, TokioCommandWrap};

**To Reproduce**

```
        let transport = TokioChildProcess::new(Command::new(&command).configure(|cmd| {
            cmd.args(&args);
            
            #[cfg(windows)]
            cmd.creation_flags(windows::Win32::System::Threading::CREATE_NO_WINDOW.0);

            // we want to ensure the MCP service is killed when our process exits
            cmd.kill_on_drop(true);

            if let Some(env) = &env {
                for (key, value) in env {
                    cmd.env(key, value);
                }
            }
        }))?;
        let client = ().serve(transport).await?;
```

The `creation_flags` is not respected, and on Windows, the mcp creates new window regardless.

The `process-wrap` seems to have a different API for  creation_flags, but users are not able to pass this flag into the TokioChildProcess::new: https://crates.io/crates/process-wrap/6.0.0 document reads:

Image

or maybe I just don't know the correct way to do it? @humbertoyusta maybe you can shed some lights on is this expected?

**Expected behavior**
tokio::process api should be respected. 

**Why do I need this**
we are creating a desktop app in both windows and macos, the desktop is end-user facing, and pop up the terminal windows is quite scary. that's why we choose to hide it. but we found we cannot suppress the terminal window anymore even if we explicitly set the flag. 

**Logs**

**Additional context**

lucasfernog-crabnebula:
## Motivation and Context

we need to be able to set KillOnDrop and most importantly CreationFlags for the process spawned - we're now using process-wrap so we need to expose that interface

see https://docs.rs/process-wrap/latest/process_wrap/#killondrop-and-creationflags

## How Has This Been Tested?
I'm working on an app that relies on CreationFlags - so i just tested it on Windows :)

## Breaking Changes
TokioCommandWrap implements From<tokio::process::Command> so this is not a breaking change, but also downstream users have more control over flags such as KillOnDrop and CreationFlags

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/child_process.rs
+++ b/crates/rmcp/src/transport/child_process.rs
@@ -63,9 +63,7 @@ impl TokioChildProcess {
         let mut command_wrap = command.into();
         command_wrap
             .command_mut()
-            .stdin(std::process::Stdio::piped());
-        command_wrap
-            .command_mut()
+            .stdin(std::process::Stdio::piped())
             .stdout(std::process::Stdio::piped());
         #[cfg(unix)]
         command_wrap.wrap(process_wrap::tokio::ProcessGroup::leader());
--- a/crates/rmcp/src/transport/child_process.rs
+++ b/crates/rmcp/src/transport/child_process.rs
@@ -59,11 +59,14 @@ impl AsyncRead for TokioChildProcessOut {
 }
 
 impl TokioChildProcess {
-    pub fn new(mut command: tokio::process::Command) -> std::io::Result<Self> {
-        command
-            .stdin(std::process::Stdio::piped())
+    pub fn new(command: impl Into<TokioCommandWrap>) -> std::io::Result<Self> {
+        let mut command_wrap = command.into();
+        command_wrap
+            .command_mut()
+            .stdin(std::process::Stdio::piped());
+        command_wrap
+            .command_mut()
             .stdout(std::process::Stdio::piped());
-        let mut command_wrap = TokioCommandWrap::from(command);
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/child_process.rs
```diff
@@ -59,11 +59,14 @@ impl AsyncRead for TokioChildProcessOut {
 }
 
 impl TokioChildProcess {
-    pub fn new(mut command: tokio::process::Command) -> std::io::Result<Self> {
-        command
-            .stdin(std::process::Stdio::piped())
+    pub fn new(command: impl Into<TokioCommandWrap>) -> std::io::Result<Self> {
+        let mut command_wrap = command.into();
+        command_wrap
```
jokemanfire, line 64:
```
        let mut command_wrap = command.into();
        command_wrap
            .command_mut()
            .stdin(std::process::Stdio::piped())
            .stdout(std::process::Stdio::piped());
```

lucasfernog-crabnebula: pushed!

Issue #220, opened 2025-05-25
tokio task not closed at shutdown


3DRaven:
**Describe the bug**
When performing a graceful shutdown with a Tokio cancellation token, the task that reads from stdin does not terminate properly, causing the process to remain running.

**To Reproduce**
1. Attached full project example with the problem. [mcp-problem-example.zip](https://github.com/user-attachments/files/20431140/mcp-problem-example.zip)
2. Add integration with this MCP server to MCP inspector or cursor IDE with stdio transport.
As example integration with cursor IDE
```json
{
  "mcpServers": {
    "mcp-problem-example": {
      "command": "sh",
      "args": [
        "path/to/mcp-problem-example/start.sh"
      ],
      "env": []
    },
}
```
4. Run in terminal tokio-console https://crates.io/crates/tokio-console
5. Disconnect/Turn off from the MCP server in client (IDE or inspector)

**Expected behavior**
After stopping the MCP server, the process should be terminated.

**Logs**
All code has finished executing, but the task remains unclosed.

```log
2025-05-25T13:24:17.338Z  INFO mcp_problem_example: Tracing initialized successfully
2025-05-25T13:24:17.338Z  INFO mcp_problem_example: Starting server in work directory: example/mcp-problem-example    
2025-05-25T13:24:17.340Z  INFO runtime.spawn{kind=task task.name= task.id=24 size.bytes=1336 loc.file="src/main.rs" loc.line=17 loc.col=11}: mcp_problem_example: Starting MCP service
2025-05-25T13:24:17.341Z  INFO runtime.spawn{kind=task task.name= task.id=24 size.bytes=1336 loc.file="src/main.rs" loc.line=17 loc.col=11}:serve_inner: rmcp::service: Service initialized as server peer_info=InitializeRequestParam { protocol_version: ProtocolVersion("2024-11-05"), capabilities: ClientCapabilities { experimental: None, roots: Some(RootsCapabilities { list_changed: Some(false) }), sampling: None }, client_info: Implementation { name: "cursor-vscode", version: "1.0.0" } }
2025-05-25T13:24:17.342Z  INFO runtime.spawn{kind=task task.name= task.id=24 size.bytes=1336 loc.file="src/main.rs" loc.line=17 loc.col=11}: mcp_problem_example: MCP server initialized
[...]

PR #272, opened 2025-06-19
fix: handle SIGTERM

mxfactorial:
closes https://github.com/modelcontextprotocol/rust-sdk/issues/220 with https://github.com/tokio-rs/axum/blob/6d78e843982ca299b5d15b7b9095c249e6cdf978/examples/graceful-shutdown/src/main.rs#L56C1-L76C6

### DIFF
--- a/crates/rmcp/src/transport/sse_server.rs
+++ b/crates/rmcp/src/transport/sse_server.rs
@@ -11,6 +11,7 @@ use axum::{
     routing::{get, post},
 };
 use futures::{Sink, SinkExt, Stream};
+use tokio::signal;
 use tokio_stream::wrappers::ReceiverStream;
 use tokio_util::sync::{CancellationToken, PollSender};
 use tracing::Instrument;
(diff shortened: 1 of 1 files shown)


4t145:
We shouldn't handle specific signal here, we should wait for those signal outside and cancel the cancellation token.

mxfactorial:
here vs "outside" is another issue

this pr lets the process owner shut down their axum server

4t145:
let me explain this, if you want to use graceful shutdown, you should await the shutdown signal and then trigger the cancellation token.
```rust
{
    let ct = CancellationToken::new();
    let config = new_config_with_ct(ct.clone());
    tokio::spawn(async move {
            graceful_shutdown().await;
            ct.cancel();
    });
    // create server
}
```

In your code, you firstly await cancellation token, and then waiting for shutdown signal, but `ct` has already been in responsibility of shutdown signal.

mxfactorial:
merge the code so process owners can shut down this axum server

"but they need to use CancellationToken"

no they don't

4t145:
Can you provide some concrete example of why you need to put this code here, personally I don't see any necessity.

The disign here is user use cancellation token to shutdown the server. You cannot forcing people wait ctrl_c or SIGTERM, it is not always the case.

mxfactorial:
SIGTERM and cntrl c are concrete enough examples

and the code gives owners simple control over their processes

they dont need disign

thats your last response

4t145:
Sorry maybe I'm not smart enough to understand you. Since you said that's your last response, I will close this pr.

PR #331, opened 2025-07-23
fix: return a 405 for GET and DELETE if non stateful streamable http

nated0g:
When a streamable http server is not in stateful mode, calls from an client to `GET /mcp` and `DELETE /mcp` will fail because there is no session id in the call.  The server should respond in these cases with a `405` status code.

## Motivation and Context
This is needed to support a fully stateless streamable http server (think a remote MCP server hosted on a serverless platform)

From the spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server

> The server MUST either return Content-Type: text/event-stream in response to this HTTP GET, or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE stream at this endpoint.

> Clients that no longer need a particular session (e.g., because the user is leaving the client application) SHOULD send an HTTP DELETE to the MCP endpoint with the Mcp-Session-Id header, to explicitly terminate the session.
>
> The server MAY respond to this request with HTTP 405 Method Not Allowed, indicating that the server does not allow clients to terminate sessions.

Right now this requires users of this crate to manually handle `GET` and `DELETE` requests before they hit the StreamableHttpService.

## How Has This Been Tested?

Change hasn't been tested directly using this code path, but we have the same logic in our service (returning a 405 `METHOD_NOT_ALLOWED` in an outer layer that wraps the service so these requests never make it through), and that is allowing a fully stateless interaction to work.

I can attempt to test this some time in the coming days, but putting this PR up for now to get input.

## Breaking Changes

No, this is not a breaking change.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have added or updated documentation as needed (I don't think this requires a docs update)

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -111,15 +111,20 @@ where
         B::Error: Display,
     {
         let method = request.method().clone();
-        let result = match method {
-            Method::GET => self.handle_get(request).await,
-            Method::POST => self.handle_post(request).await,
-            Method::DELETE => self.handle_delete(request).await,
+        let allowed_methods = match self.config.stateful_mode {
+            true => "GET, POST, DELETE",
+            false => "POST",
+        };
+        let result = match (method, self.config.stateful_mode) {
+            (Method::POST, _) => self.handle_post(request).await,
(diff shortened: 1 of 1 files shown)


4t145:
LGTM, thanks

PR #332, opened 2025-07-23
feat: Add MCP Elicitation support for interactive user input

bug-ops:
This PR implements comprehensive MCP Elicitation support according to the MCP 2025-06-18 specification, enabling servers to request interactive user input during tool execution with full JSON Schema validation and type-safe convenience
   methods.

  ## Motivation and Context
  

  **Problem**: The current rust-sdk implementation lacks support for MCP Elicitation functionality introduced in the MCP 2025-06-18 specification. This prevents servers from creating interactive workflows that require user input during
  tool execution.

  **Solution**: This change adds complete elicitation support, enabling:
  - **Interactive confirmations** for destructive operations (e.g., "Delete this file?")
  - **Dynamic parameter collection** when initial context is insufficient
  - **Structured data input** with JSON Schema validation
  - **User-guided workflows** for complex multi-step processes

  **Context**: Based on [MCP PR #382](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382) which introduced elicitation to the MCP specification. This implementation ensures rust-sdk maintains full feature parity with
  the latest MCP standards.

  ## How Has This Been Tested?
  

  **Comprehensive Test Suite** (11 test cases):
  - [x] **JSON serialization/deserialization** - All data structures correctly serialize to/from JSON
  - [x] **JSON-RPC protocol compliance** - Full request/response cycle validation
  - [x] **MCP 2025-06-18 specification compliance** - Method names, protocol version, enum values
  - [x] **Error handling and edge cases** - Invalid schemas, empty messages, malformed requests
  - [x] **Performance benchmarks** - 1000 serialization cycles under 1000ms
  - [x] **Capabilities integration** - Builder pattern and schema validation flags
  - [x] **Convenience methods validation** - All four helper methods produce correct schemas
  - [x] **Complex JSON Schema support** - Nested objects, arrays, validation rules

  **Quality Assurance**:
  - [x] `cargo check` - No compilation errors
  - [x] `cargo clippy` - No linting warnings
  - [x] `cargo test` - All 11 elicitation tests pass
  - [x] `cargo fmt` - Code properly formatted

  **Scenarios Tested**:
  - Simple yes/no confirmations
  - Text input with/without validation requirements
  - Multiple choice selections
  - Complex structured data collection with nested JSON schemas
  - Error conditions and user cancellation flows

  ## Breaking Changes
  

  **No breaking changes** - This is a purely additive implementation:
  - [x] All existing APIs remain unchanged
  - [x] Backward compatibility maintained for all protocol versions
  - [x] Elicitation is opt-in via capabilities negotiation
  - [x] Servers without elicitation support continue working normally
  - [x] New protocol version V_2025_06_18 added alongside existing versions

  ## Types of changes
  
  - [ ] Documentation update

  ## Checklist
  
  ## Additional context
  

  ### Implementation Highlights:

  **Core Architecture**:
  - **ElicitationAction enum** - `Accept`, `Decline`, `Cancel` actions
  - **Type-safe structures** - `CreateElicitationRequestParam`, `CreateElicitationResult`
  - **Protocol integration** - New V_2025_06_18 version with method constants
  - **JSON-RPC compliance** - Full request/response cycle support

  **Developer Experience**:
  - **4 convenience methods** for common patterns:
    - `elicit_confirmation()` - Simple boolean confirmations
    - `elicit_text_input()` - String input with optional validation
    - `elicit_choice()` - Multiple choice selection
    - `elicit_structured_input()` - Complex JSON Schema validation

  **Capabilities System**:
  - **ElicitationCapability** with optional schema validation
  - **Builder pattern integration** - `enable_elicitation()`, `enable_elicitation_schema_validation()`
  - **Seamless integration** with existing capabilities architecture

  This implementation makes rust-sdk compatible with MCP 2025-06-18 and enables interactive workflows while maintaining full backward compatibility.

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -143,9 +143,10 @@ impl std::fmt::Display for ProtocolVersion {
 }
 
 impl ProtocolVersion {
+    pub const V_2025_06_18: Self = Self(Cow::Borrowed("2025-06-18"));
     pub const V_2025_03_26: Self = Self(Cow::Borrowed("2025-03-26"));
     pub const V_2024_11_05: Self = Self(Cow::Borrowed("2024-11-05"));
-    pub const LATEST: Self = Self::V_2025_03_26;
+    pub const LATEST: Self = Self::V_2025_06_18;
 }
 
 impl Serialize for ProtocolVersion {
@@ -167,6 +168,7 @@ impl<'de> Deserialize<'de> for ProtocolVersion {
         match s.as_str() {
             "2024-11-05" => return Ok(ProtocolVersion::V_2024_11_05),
             "2025-03-26" => return Ok(ProtocolVersion::V_2025_03_26),
+            "2025-06-18" => return Ok(ProtocolVersion::V_2025_06_18),
             _ => {}
         }
         Ok(ProtocolVersion(Cow::Owned(s)))
@@ -1174,6 +1176,75 @@ const_string!(RootsListChangedNotificationMethod = "notifications/roots/list_cha
 pub type RootsListChangedNotification = NotificationNoParam<RootsListChangedNotificationMethod>;
 
 // =============================================================================
+// ELICITATION (INTERACTIVE USER INPUT)
+// =============================================================================
+
+// Method constants for elicitation operations.
+// Elicitation allows servers to request interactive input from users during tool execution.
+const_string!(ElicitationCreateRequestMethod = "elicitation/create");
+const_string!(ElicitationResponseNotificationMethod = "notifications/elicitation/response");
+
+/// Represents the possible actions a user can take in response to an elicitation request.
+///
+/// When a server requests user input through elicitation, the user can:
+/// - Accept: Provide the requested information and continue
+/// - Decline: Refuse to provide the information but continue the operation
+/// - Cancel: Stop the entire operation
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub enum ElicitationAction {
+    /// User accepts the request and provides the requested information
+    Accept,
+    /// User declines to provide the information but allows the operation to continue
+    Decline,
+    /// User cancels the entire operation
+    Cancel,
+}
+
+/// Parameters for creating an elicitation request to gather user input.
+///
+/// This structure contains everything needed to request interactive input from a user:
+/// - A human-readable message explaining what information is needed
+/// - A JSON schema defining the expected structure of the response
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct CreateElicitationRequestParam {
+    /// Human-readable message explaining what input is needed from the user.
+    /// This should be clear and provide sufficient context for the user to understand
+    /// what information they need to provide.
+    pub message: String,
+
+    /// JSON Schema defining the expected structure and validation rules for the user's response.
+    /// This allows clients to validate input and provide appropriate UI controls.
+    /// Must be a valid JSON Schema Draft 2020-12 object.
+    pub requested_schema: JsonObject,
+}
+
+/// The result returned by a client in response to an elicitation request.
+///
+/// Contains the user's decision (accept/decline/cancel) and optionally their input data
+/// if they chose to accept the request.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct CreateElicitationResult {
+    /// The user's decision on how to handle the elicitation request
+    pub action: ElicitationAction,
+
+    /// The actual data provided by the user, if they accepted the request.
+    /// Must conform to the JSON schema specified in the original request.
+    /// Only present when action is Accept.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub content: Option<Value>,
+}
+
+/// Request type for creating an elicitation to gather user input
+pub type CreateElicitationRequest =
+    Request<ElicitationCreateRequestMethod, CreateElicitationRequestParam>;
+
+// =============================================================================
 // TOOL EXECUTION RESULTS
 // =============================================================================
 
@@ -1303,7 +1374,8 @@ ts_union!(
     | SubscribeRequest
     | UnsubscribeRequest
     | CallToolRequest
-    | ListToolsRequest;
+    | ListToolsRequest
+    | CreateElicitationRequest;
 );
 
 ts_union!(
@@ -1315,7 +1387,7 @@ ts_union!(
 );
 
 ts_union!(
-    export type ClientResult = CreateMessageResult | ListRootsResult | EmptyResult;
+    export type ClientResult = CreateMessageResult | ListRootsResult | CreateElicitationResult | EmptyResult;
 );
 
 impl ClientResult {
@@ -1330,7 +1402,8 @@ ts_union!(
     export type ServerRequest =
     | PingRequest
     | CreateMessageRequest
-    | ListRootsRequest;
+    | ListRootsRequest
+    | CreateElicitationRequest;
 );
 
 ts_union!(
@@ -1355,6 +1428,7 @@ ts_union!(
     | ReadResourceResult
     | CallToolResult
     | ListToolsResult
+    | CreateElicitationResult
     | EmptyResult
     ;
 );
--- a/crates/rmcp/src/model/capabilities.rs
+++ b/crates/rmcp/src/model/capabilities.rs
@@ -39,6 +39,22 @@ pub struct RootsCapabilities {
     pub list_changed: Option<bool>,
 }
 
+/// Capability for handling elicitation requests from servers.
+///
+/// Elicitation allows servers to request interactive input from users during tool execution.
+/// This capability indicates that a client can handle elicitation requests and present
+/// appropriate UI to users for collecting the requested information.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct ElicitationCapability {
+    /// Whether the client supports JSON Schema validation for elicitation responses.
+    /// When true, the client will validate user input against the requested_schema
+    /// before sending the response back to the server.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub schema_validation: Option<bool>,
+}
+
 ///
 /// # Builder
 /// ```rust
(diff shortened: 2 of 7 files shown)


bug-ops:
I think we might want to consider implementing different request options (including timeouts) depending on the request type. For elicitation calls, they could potentially require longer timeouts and would benefit from being configurable.

crates/rmcp/src/model.rs
```diff
@@ -143,9 +143,10 @@ impl std::fmt::Display for ProtocolVersion {
 }
 
 impl ProtocolVersion {
+    pub const V_2025_06_18: Self = Self(Cow::Borrowed("2025-06-18"));
     pub const V_2025_03_26: Self = Self(Cow::Borrowed("2025-03-26"));
     pub const V_2024_11_05: Self = Self(Cow::Borrowed("2024-11-05"));
-    pub const LATEST: Self = Self::V_2025_03_26;
+    pub const LATEST: Self = Self::V_2025_06_18;
```
4t145: Please don't change it to 20250618 now, since we haven't implment all the new features

4t145 (changes requested):

4t145:
Sorry for late response, it looks good to me.

crates/rmcp/src/model.rs
bug-ops: Done! Reverted LATEST back to V_2025_03_26 in commit 923d9aa.

jokemanfire:
CI fail ， check pls

bug-ops:
Fixed commit messages and tests. Ready for review

4t145 (approved):

4t145:
Looks good, one more question here, should we make `elicitation` an optional feature? Or it's temporary approach before we release `V_2025_06_18` ?

bug-ops:
@4t145 After implementing and testing the feature flag approach, I think that making elicitation optional is actually quite straightforward with several clear benefits.
Data model: types must be always available - ensures full protocol compatibility
Working functionality (methods): Feature-gated - allows selective inclusion

Benefits of this approach:
- Protocol compatibility - Applications can always deserialize elicitation messages
- Gradual adoption - Teams can opt-in to server-side functionality when ready
- Reduced dependencies - Avoids pulling in schemars for users who only need protocol types
- Clean separation - Clear distinction between "understanding the protocol" vs "implementing the feature"

This gives us the best of both - full protocol compatibility with selective functionality. I think we must keep as optional feature even post-v2025-06-18.

Next step: I'm ready to expand the elicitation feature flag to cover all server-side elicitation logic if that would be helpful - just let me know the scope you'd prefer!

bug-ops:
I wanted to bring up a potential consideration around type safety for the `elicit<T>()` methods. Currently, these methods accept any deserializable type, which could lead to some edge cases:

```rust
// These compile but may not align with MCP client expectations
let name: String = server.elicit().await?;        // JSON: "value" (not an object)
let items: Vec<i32> = server.elicit().await?;     // JSON: [1,2,3] (array, not object)
```

Since MCP clients typically expect **JSON objects** for elicitation schemas, I'm wondering if we should consider adding some compile-time constraints.
One possible approach could be a sealed trait pattern:

```rust
#[derive(Serialize, Deserialize, JsonSchema)]
struct UserInput { name: String, email: String }
impl_elicitation_safe!(UserInput);  // Macro for opt-in safety

let input: UserInput = server.elicit().await?;  // ✅ Guaranteed object
```

Alternative approaches:

- Keep current flexibility and document the object requirement
- Add runtime validation instead of compile-time constraints
- Provide both safe and unsafe variants

What do you think? Is this level of type safety worth the additional complexity, or should we prioritize flexibility here? I'd love to hear the  thoughts on this trade-off. Happy to implement whatever direction feels right for the project!

alexhancock:
@bug-ops Good callout. I think compile-time constraints are a good bet and what you have looks useful to me.

@jamadeo Do you have thoughts?

crates/rmcp/src/service/server.rs
```diff
@@ -338,3 +399,212 @@ impl Peer<RoleServer> {
...
+#[cfg(feature = "elicitation")]
+#[derive(Error, Debug)]
+pub enum ElicitationError {
+    /// The elicitation request failed at the service level
+    #[error("Service error: {0}")]
+    Service(#[from] ServiceError),
+
+    /// User declined to provide input or cancelled the request  
+    #[error("User declined or cancelled the request")]
+    UserDeclined,
```
jamadeo, line 420: I think declined and canceled are materially, if subtly, different. The spec says:

> The three response actions are:
> 
> 1. **Accept** (`action: "accept"`): User explicitly approved and submitted with data
> 
>    - The `content` field contains the submitted data matching the requested schema
>    - Example: User clicked "Submit", "OK", "Confirm", etc.
> 
> 2. **Decline** (`action: "decline"`): User explicitly declined the request
> 
>    - The `content` field is typically omitted
>    - Example: User clicked "Reject", "Decline", "No", etc.
> 
> 3. **Cancel** (`action: "cancel"`): User dismissed without making an explicit choice
>    - The `content` field is typically omitted
>    - Example: User closed the dialog, clicked outside, pressed Escape, etc.
> 
> Servers should handle each state appropriately:
> 
> - **Accept**: Process the submitted data
> - **Decline**: Handle explicit decline (e.g., offer alternatives)
> - **Cancel**: Handle dismissal (e.g., prompt again later)
> 

so probably they should be separated here

crates/rmcp/src/model/capabilities.rs
```diff
@@ -39,6 +39,22 @@ pub struct RootsCapabilities {
...
+/// appropriate UI to users for collecting the requested information.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct ElicitationCapability {
+    /// Whether the client supports JSON Schema validation for elicitation responses.
+    /// When true, the client will validate user input against the requested_schema
+    /// before sending the response back to the server.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub schema_validation: Option<bool>,
```
jamadeo, line 56: didn't see this option in the spec -- what's the motivation for adding it? 

only related bit I found in the spec was
> Both parties SHOULD validate elicitation content against the provided schema

crates/rmcp/src/service/server.rs
4t145: Sounds reasonable

crates/rmcp/src/model/capabilities.rs
bug-ops: Hi @jamadeo, the optional nature of `schema_validation` is actually aligned with how the MCP specification uses the `SHOULD` keyword. According to [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119), `SHOULD` means:

> "there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course."

The [MCP 2025-06-18 specification](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation) states:

> "Both parties SHOULD validate elicitation content against the provided schema"

This deliberate use of `SHOULD` (rather than `MUST`) suggests the protocol designers anticipated scenarios where validation might be bypassed for valid reasons.
Valid use cases for disabling schema validation:

- Performance-critical environments
- High-throughput scenarios
- Legacy integrations where validation happens elsewhere in the stack
- Trusted internal networks with simplified security models
- Progressive enhancement for clients with limited JSON Schema support

However, the implementation should:

- Default to true (following the `SHOULD` recommendation)
- Clearly document the security and reliability implications
- Ensure servers can handle potentially invalid data gracefully
- Provide explicit opt-out rather than accidental disabling

This approach balances protocol compliance with real-world flexibility, which seems to align with MCP's pragmatic design philosophy.
What do you think? Does this rationale make sense for the optional capability?

jamadeo: Yeah I think that's the right interpretation of should vs must. Mostly I just meant is it valid to put it in the capabilities object itself. (I'm not familiar enough with what's permitted there.)

crates/rmcp/src/service/server.rs
bug-ops: Done

crates/rmcp/src/service/server.rs
```diff
@@ -3,6 +3,34 @@ use std::borrow::Cow;
 use thiserror::Error;
 
 use super::*;
+
+/// Validates timeout values to prevent DoS attacks and ensure reasonable limits
+fn validate_timeout(timeout: Option<std::time::Duration>) -> Result<(), ServiceError> {
```
jamadeo: Hmm is this necessary? Why does it need to be limited to 1ms-5min? Values outside of that might be very uncommon but I'd think that should be left to the server developer.

jamadeo (approved):
Looks reasonable to me, but I question the addition of the `validate_timeout` function.

crates/rmcp/src/service/server.rs
bug-ops: @jamadeo You're absolutely right. Hard-coding these limits was an oversight - server developers should have control over timeout policies for their specific use cases.

I'll remove the validation and let developers implement their own timeout constraints as needed. The library shouldn't enforce arbitrary business logic decisions.

Thanks for catching this!

4t145 (approved):
Looks good to me

crates/rmcp/tests/test_elicitation.rs
```diff
@@ -0,0 +1,1507 @@
...
+        "SubscribeRequest",
+        "UnsubscribeRequest",
+        "CallToolRequest",
+        "ListToolsRequest",
+        // CreateElicitationRequest should NOT be here
+    ];
+
+    // Verify the list doesn't contain elicitation
+    assert!(!valid_client_requests.contains(&"CreateElicitationRequest"));
+    assert_eq!(valid_client_requests.len(), 13); // Should be 13, not 14
```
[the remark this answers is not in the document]
4t145: I don't get the point of this testcase, it looks like some LLM generated stuff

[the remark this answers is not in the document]
bug-ops: You're absolutely right - I missed cleaning up this redundant code during refactoring. The presence of `CreateElicitationRequest` in `ServerRequest` is already verified by the `test_elicitation_direction_server_to_client` test, which
  comprehensively checks the protocol direction constraints.

  This match block can be safely removed since it's purely duplicative testing. I'll clean this up.

4t145 (dismissed):

4t145:
@jamadeo Could you review again?

jamadeo (approved):

jamadeo:
I tested this out from a client perspective by adding a trivial handler in Goose + the "everything" test server, and looked good!

crates/rmcp/src/service/server.rs
```diff
@@ -324,11 +328,68 @@ macro_rules! method {
             Ok(())
         }
     };
+
+    // Timeout-only variants (base method should be created separately with peer_req)
```
jokemanfire, line 332: I think these two macros have same public code ,we can extract it.

jokemanfire:
Thanks , LGTM  , If you can add an example about this feature, that will be greater.

bug-ops:
@jokemanfire Added a simple elicitation example demonstrating the feature in action

4t145:
Fix ci please!

bug-ops:
@4t145 ci fixed

4t145 (approved):

4t145:
thanks!

PR #334, opened 2025-07-25
fix: propagate tracing spans when spawning new tokio tasks

nated0g:
This change instruments futures passed to `tokio::spawn` so that tracing spans are propagated through and associated with a common root span.

## Motivation and Context
Currently, when emitting tracing events or spans in tool calls, the spans aren't attached to a parent.  This means that if running something like an axum service, spans for http requests processed by axum don't include spans emitted for tool calls.

## How Has This Been Tested?
Yes, tested this in an existing MCP server, running on axum.  We have a tracing subscriber that rolls all spans and events in connected to a common root span into a single log line.  Without this change, the tool call's spans got emitted in separate log line.  With this change, all spans emitted have a common root span, and we get a single log line including all associated spans and events.

## Breaking Changes
I don't think this is breaking.. may change the format of people's logs depending how they have their subscriber configured.

## Types of changes

- [ ] Documentation update

## Checklist

- [n/a] I have added appropriate error handling
- [n/a] I have added or updated documentation as needed

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -30,7 +30,7 @@ use tokio_util::sync::{CancellationToken, DropGuard};
 #[cfg(feature = "tower")]
 #[cfg_attr(docsrs, doc(cfg(feature = "tower")))]
 pub use tower::*;
-use tracing::instrument;
+use tracing::{Instrument as _, instrument};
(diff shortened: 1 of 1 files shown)


jokemanfire (approved):

4t145:
thanks!

PR #350, opened 2025-08-04
feat: Adjust log levels for trace logging

kyle-leonhard:
Reduce log levels on per-connection/request/message events to `trace` from `info`.

I considered using `DEBUG` which is the default level used by Axum for request logging. WDYT?

## Motivation and Context
Emitting per-connection and message logs at INFO level is unexpected, in my experience. Using a finer level allows filtering to be effective without requiring users to set module level configuration.

## How Has This Been Tested?
Locally with my MCP server

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
None

### DIFF
--- a/crates/rmcp/src/handler/server.rs
+++ b/crates/rmcp/src/handler/server.rs
@@ -213,7 +213,7 @@ pub trait ServerHandler: Sized + Send + Sync + 'static {
         &self,
         context: NotificationContext<RoleServer>,
     ) -> impl Future<Output = ()> + Send + '_ {
-        tracing::info!("client initialized");
+        tracing::debug!("client initialized");
         std::future::ready(())
     }
     fn on_roots_list_changed(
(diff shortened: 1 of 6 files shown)


jokemanfire:
I'm not sure if the level of these logs needs to be lowered. What's your opinion? @4t145

4t145:
Sorry for late response
>  Using a finer level allows filtering to be effective without requiring users to set module level configuration.

Current log level are base on the following standards
- session/connection level: info
- reqeust/message level: debug
- more verbose log: trace

I don't think set all level to trace could be "finer", there are more logs during the connect/request, if they are all in trace level, it would be inconvenient to distinguish them. I think the debug level could be OK. But after all, set module level configuration won't cost you a lot of time. So I am not going to close this pr and maybe more people can give me some suggestion.

Issue #355, opened 2025-08-06
structured_content validate: Error: Unexpected response type


cougarwww:
https://github.com/modelcontextprotocol/rust-sdk/blob/984cd87ed85aeba2bee7ada77b9152ecca837fcc/crates/rmcp/src/model.rs#L1265-L1272

use python/fastmcp will return both content and structured_content
```
Result: 
 - CallToolResult {
    content: Some(
        [
            Annotated {
                raw: Text(
                    RawTextContent {
                        text: "2025-08-06 14:24:47",
                    },
                ),
                annotations: None,
            },
        ],
    ),
    structured_content: Some(
        Object {
            "result": String("2025-08-06 14:24:47"),
        },
    ),
    is_error: Some(
        false,
    ),
}
```

crepererum:
I think the assumption in `CallToolResult` is wrong:

https://github.com/modelcontextprotocol/rust-sdk/blob/984cd87ed85aeba2bee7ada77b9152ecca837fcc/crates/rmcp/src/model.rs#L1185

The [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content) states:

> For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

This issue affects the client -- where the validation fails as https://github.com/modelcontextprotocol/rust-sdk/issues/355#issue-3295351138 illustrates -- and the server (at least the example that should probably follow the recommendation and return both).

michaelneale:
we hit this also in goose - with some python MCP servers. If we patch the server to only return serialized json it works, but I expect will cause others to fail which can't be patched as quickly.

michaelneale:
patch for this: https://github.com/modelcontextprotocol/rust-sdk/pull/357 thanks @crepererum @cougarwww that suggestion worked in my case (when I patched it).

michaelneale:
this is fixed now - can close it I think @cougarwww, nice work!

cougarwww:
Thinks all for the nice work!

PR #357, opened 2025-08-07
fix(model): allow backwards compat as per MCP spec

michaelneale:
This is a very small fix for issue: https://github.com/modelcontextprotocol/rust-sdk/issues/355 

MCP servers (especially in python) will likely fail with 0.4.0 with an error like "Error: Unexpected response type"
this is due to the validation  `content and structured_content are mutually exclusive`, but this is a bit strict. 

## How Has This Been Tested?

Tested with goose and a python based MCP server - it worked with < 0.4.0 and failed with 0.4.0. 
I then tested this patch with the goose agent and exactly the same server it works. 

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/tests/test_structured_output.rs
+++ b/crates/rmcp/tests/test_structured_output.rs
@@ -2,7 +2,7 @@
 use rmcp::{
     Json, ServerHandler,
     handler::server::{router::tool::ToolRouter, tool::Parameters},
-    model::{CallToolResult, Content, Tool},
+    model::{CallToolResult, Tool},
     tool, tool_handler, tool_router,
 };
 use schemars::JsonSchema;
--- a/crates/rmcp/tests/test_structured_output.rs
+++ b/crates/rmcp/tests/test_structured_output.rs
@@ -145,27 +145,6 @@ async fn test_structured_error_in_call_result() {
 }
 
 #[tokio::test]
-async fn test_mutual_exclusivity_validation() {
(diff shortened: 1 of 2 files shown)


alexhancock (approved):

michaelneale:
@alexhancock PTAL again - not sure about the comment lint check though

crates/rmcp/tests/test_structured_output.rs
```diff
@@ -144,27 +144,6 @@ async fn test_structured_error_in_call_result() {
     assert_eq!(result.is_error, Some(true));
 }
 
-#[tokio::test]
```
crepererum, line 147: maybe keep the test and check that it it doesn't error at the end?

crepererum (commented):
GitHub doesn't let me comment outside the diff, but I think the `CallToolResult` struct docstring needs adjustment. The last line needs to be removed:

```rust
///
/// Note: `content` and `structured_content` are mutually exclusive - exactly one must be provided.
```

alexhancock:
@crepererum I addressed the feedback in https://github.com/modelcontextprotocol/rust-sdk/pull/359 - not pushing to this one since it's @michaelneale's fork and I won't have access

alexhancock:
Merged https://github.com/modelcontextprotocol/rust-sdk/pull/359 which has these changes!

PR #364, opened 2025-08-11
fix: make stdio shutdown more graceful

jokemanfire:
According to the protocol specifications

## Motivation and Context
Make shutdown more graceful.

## How Has This Been Tested?
Ut

## Breaking Changes
No breaking change

## Types of changes

- [ ] Documentation update

## Checklist


crates/rmcp/src/transport/child_process.rs
```diff
@@ -41,13 +42,35 @@ pub struct TokioChildProcess {
...
+    inner: Option<Box<dyn TokioChildWrapper>>,
 }
 
 impl Drop for ChildWithCleanup {
     fn drop(&mut self) {
-        if let Err(e) = self.inner.start_kill() {
-            tracing::warn!("Failed to kill child process: {e}");
+        // Close child more graceful
+        if let Some(mut child) = self.inner.take() {
+            // Spawn a background task to clean up
```
crepererum: So the child will eventually be cleaned up. For certain applications however it would be nice if the caller / API user would have a way to have a more controlled way to clean up / shutdown, something like `child.shutdown().await` that also returns the error instead of just logging it.

jokemanfire:
I'm not sure if adding a shutdown interface is necessary, although it may increase some flexibility for the caller, it may be redundant for the SDK. For me, the previous direct killing behavior was not very elegant, and I would like to hear more opinions.How about you ?
@4t145  @alexhancock

4t145:
I prefer to have a easier approch to cleanup in drop, and give user more flexibility in another explicit method. Or maybe we can just provide an into_inner method so the user can do anything they like with inner process. We can also wrap this wait-and-kill approch into a method, just as @crepererum metioned.

jokemanfire:
Approve , I will change it.

jokemanfire:
How about expose the transport  in running service? @4t145

4t145:
@jokemanfire Just have a look on the original issue
1. We can return back the original transport with quit reason. It's not good to put the transport into the `RunningService` structure, since we already hand over the control of transport into the service worker task, and we can only take it back after task over.
2. We don't have to close the transport in `drop`, it's a historical issue. The `ChildProcess` get the transport abilities from `AsyncRWTransport`, and this transport just return a `Ok(())` in `close` method, where we should implment the closing logic of `ChildProcess` in. So I think what we need to do is implement `Transport` trait for it instead of inherit from `AsyncRWTransport`.
3. By default, we shall follow the recommand approch of official document. But we can still remain some flexibility to user. I think the proper way for user to costomize the closing logic is to create a new type based on current `ChildProcess` transport and shadow the original closing logic, in other word, inherit and overload.

jokemanfire:
@4t145  @crepererum  Have a look , and pls check.

crates/rmcp/src/transport/child_process.rs
```diff
@@ -157,19 +194,61 @@ impl TokioChildProcessBuilder {
...
+    fn send(
+        &mut self,
+        item: TxJsonRpcMessage<R>,
+    ) -> impl Future<Output = Result<(), Self::Error>> + Send + 'static {
+        let json = serde_json::to_string(&item).unwrap();
+        let child_stdin = Arc::clone(&self.child_stdin);
+
+        async move {
+            let mut child_stdin = child_stdin.lock().await;
+            let serialized = format!("{}\n", json);
```
4t145: this may allocate memory twice for the same content

1. you can write json firstly and then write a `\n` later.
2. we aleady have codec implementation in `async_rw`, maybe you can reuse it?

```rust
pub struct AsyncRwTransport<Role: ServiceRole, R: AsyncRead, W: AsyncWrite> {
    read: FramedRead<R, JsonRpcMessageCodec<RxJsonRpcMessage<Role>>>,
    write: Arc<Mutex<FramedWrite<W, JsonRpcMessageCodec<TxJsonRpcMessage<Role>>>>>,
}
```

crates/rmcp/src/transport/child_process.rs
```diff
@@ -157,19 +194,61 @@ impl TokioChildProcessBuilder {
...
         Ok((proc, stderr_opt))
     }
 }
 
-impl<R: ServiceRole> IntoTransport<R, std::io::Error, ()> for TokioChildProcess {
-    fn into_transport(self) -> impl Transport<R, Error = std::io::Error> + 'static {
-        IntoTransport::<R, std::io::Error, super::async_rw::TransportAdapterAsyncRW>::into_transport(
-            self.split(),
-        )
+impl<R: ServiceRole> Transport<R> for TokioChildProcess {
```
4t145: Should we specify the role as `RoleClient`?

4t145:
And what's your opinion about give back the ownership of transport after shutdown? Maybe we don't implement this feature in this pr, but it still a point to be discussed.

crates/rmcp/src/transport/child_process.rs
jokemanfire: If use the async_rw trait , It's better to limit to RoleClient.

crates/rmcp/src/transport/child_process.rs
jokemanfire: It is also possible to use inheritance instead of overloading

jokemanfire:
> And what's your opinion about give back the ownership of transport after shutdown? Maybe we don't implement this feature in this pr, but it still a point to be discussed.

If we want to return ownership of the transport, I don't think it should be in the close interface. This can lead to misunderstandings in our understanding of the close interface. I suggest that if necessary, we can set up new interface extensions, and not every transport needs to implement ownership. I have added the into_inner interface  for future extensions.

crates/rmcp/src/transport/child_process.rs
jokemanfire: I've wrapper the AsyncRwTransport for reused.

4t145:
> This can lead to misunderstandings in our understanding of the close interface

I mean, return it as serve task's result 

maybe change this
```rust
pub async fn waiting(self) -> Result<QuitReason, tokio::task::JoinError>
```

to

```rust
pub async fn waiting(self) -> Result<(QuitReason, T), tokio::task::JoinError>
```

But it will make more difficult to make `RunningService` a unified type, makes it harder for server management. Just a idea in case of anybody want transport ownership back after service shutdown, in the future.

crepererum:
Thank you :slightly_smiling_face:

Issue #374, opened 2025-08-14
Wrong implementation on json schema validation


4t145:
**Describe the bug**

current implementation
```rust
/// Validate that a JSON value conforms to basic type constraints from a schema.
///
/// Note: This is a basic validation that only checks type compatibility.
/// For full JSON Schema validation, a dedicated validation library would be needed.
pub fn validate_against_schema(
    value: &serde_json::Value,
    schema: &JsonObject,
) -> Result<(), crate::ErrorData> {
    // Basic type validation
    if let Some(schema_type) = schema.get("type").and_then(|t| t.as_str()) {
        let value_type = get_json_value_type(value);

        if schema_type != value_type {
            return Err(crate::ErrorData::invalid_params(
                format!(
                    "Value type does not match schema. Expected '{}', got '{}'",
                    schema_type, value_type
                ),
                None,
            ));
        }
    }

    Ok(())
}
```

Apparently this is too simple to do it, it only cover the basic types.

Type aslo could be other non-basic type like `integer` or multiple type like `["string", "number"]`. So we need a correct implementation.

**To Reproduce**
1. 

**Expected behavior**

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

4t145:
update: Just found it is client should do the validation: https://modelcontextprotocol.io/specification/2025-06-18/server/tools

PR #375, opened 2025-08-14
fix(tool): remove unnecessary schema validation

4t145:
1. remove unnecessary schema validation
2. add a method directly convert the structured output into typed data.
## Motivation and Context

We don't really need to do the validation in rust, because deserilization already is a validation. And if user want to do the validation, they can use the [jsonschame](https://crates.io/crates/jsonschema) crate.

And also, the original validation method is a wrong implementation, as I mentioned in #374

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/handler/server/router/tool.rs
+++ b/crates/rmcp/src/handler/server/router/tool.rs
@@ -6,7 +6,6 @@ use schemars::JsonSchema;
 use crate::{
     handler::server::tool::{
         CallToolHandler, DynCallToolHandler, ToolCallContext, schema_for_type,
-        validate_against_schema,
     },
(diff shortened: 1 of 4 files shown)


4t145:
@ccamel55 please review this. I didn't notice the problem when I merge the pr.

ccamel55 (approved):
LGTM but wouldn't this technically be a breaking change since it removes public facing API? I'll leave it to the maintainers to to decide though 😄

4t145:
@ccamel55 It's ok to introduce breaking changes, we haven't release the deleted api.

jokemanfire (approved):
It can  be Deserialized , so it is structed json data, If I understand it ?

PR #351, opened 2025-08-04
feat: Add prompt support

snowmead:
## Add Prompt Support

This PR introduces a complete prompt system for the MCP Rust SDK, enabling servers to provide reusable prompt templates that LLMs can discover and use. The implementation follows the MCP specification for prompts and provides a developer-friendly API through procedural macros.

#### 🎯 Core Functionality
- **Prompt Macros**: `#[prompt]`, `#[prompt_router]`, and `#[prompt_handler]` for easy prompt definition
- **Type-Safe Parameters**: Automatic JSON schema generation from Rust types with compile-time validation
- **Unified Architecture**: Shared parameter extraction infrastructure between tools and prompts
- **Flexible Handlers**: Support for async/sync methods, standalone functions, and closures

#### 🔧 Technical Improvements
- **Extracted `Parameters<T>` wrapper** to dedicated module for better organization
- **Unified trait system** using `FromContextPart` for both tools and prompts (removed duplicate traits)
- **Enhanced `Json<T>` wrapper** with structured output support via `IntoCallToolResult`
- **Improved error messages** with descriptive panics for schema serialization failures
- **Clean architecture** with shared extractors reducing code duplication by ~300 lines

#### Macro Support
```rust
#[prompt(description = "Generate code based on requirements")]
async fn code_generator(params: Parameters<CodeRequest>) -> Vec<PromptMessage> {
    // Type-safe parameter handling with automatic validation
}
```

#### Router Integration
```rust
#[prompt_router]
impl MyServer {
    #[prompt]
    async fn greeting_prompt(&self, params: Parameters<Args>) -> Vec<PromptMessage> {
        // Automatically registered in router
    }
}
```

### File Structure
- `crates/rmcp-macros/src/prompt*.rs` - Macro implementations
- `crates/rmcp/src/handler/server/prompt.rs` - Core prompt handling logic
- `crates/rmcp/src/handler/server/wrapper/parameters.rs` - Unified parameter extraction
- `examples/servers/src/prompt_stdio.rs` - Complete working example

### Testing
- **Comprehensive test coverage** with 5 new test files
- **Integration tests** validating prompt discovery, invocation, and parameter handling
- **Example server** demonstrating real-world usage patterns

### Breaking Changes
None - The refactoring maintains backward compatibility while unifying the internal architecture.

### Migration Guide
For users implementing custom extractors:
- Replace `FromToolCallContextPart` with `FromContextPart<ToolCallContext>`
[...]

PR #385, opened 2025-08-20
fix(macros): Allow `#[tool]` to work even if Future is not in scope

swlynch99:
This changes the expansion of `#[tool]` to emit `std::future::Future` instead of `Future`.

## Motivation and Context

This is a pretty small papercut, but if you use `#[tool]` on an async function then it requires that `Future` be imported.

## How Has This Been Tested?

`cargo check --all-targets` passes

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp-macros/src/prompt.rs
+++ b/crates/rmcp-macros/src/prompt.rs
@@ -108,10 +108,10 @@ pub fn prompt(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream>
             }
             match &fn_item.sig.output {
                 syn::ReturnType::Default => {
-                    quote! { -> futures::future::BoxFuture<#lt, ()> }
+                    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = ()> + Send + #lt>> }
                 }
                 syn::ReturnType::Type(_, ty) => {
-                    quote! { -> futures::future::BoxFuture<#lt, #ty> }
+                    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = #ty> + Send + #lt>> }
                 }
             }
         })?;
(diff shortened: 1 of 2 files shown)


4t145:
The current MSRV is 1.85.1, so I think by default `Future` is supposed to be in the scope?

swlynch99:
As far I can tell that only applies to the 2024 edition. Previous editions do not have `Future` in the prelude.

Anyway it seems like #351 changed this to use `futures::BoxFuture`. Unfortunately that is still wrong. Now you're requiring that downstream crates have `futures` available locally.

4t145:
> As far I can tell that only applies to the 2024 edition. Previous editions do not have Future in the prelude.

I am not saying to reject to support 2021 edition, just wondering in what case you have to use 2021 edtion instead 2024 edition. There was a developer also requested the compatibility for the 2021 edition, it's because he is still using windows 7.

> Anyway it seems like https://github.com/modelcontextprotocol/rust-sdk/pull/351 changed this to use futures::BoxFuture. Unfortunately that is still wrong.

You are right, I missed this when review code.

4t145 (approved):

4t145:
It should have the same problem in the newly merged prompt macro's code, could you fix it together in this PR?

swlynch99:
> I am not saying to reject to support 2021 edition, just wondering in what case you have to use 2021 edtion instead 2024 edition. There was a developer also requested the compatibility for the 2021 edition, it's because he is still using windows 7.

The features added in the 2024 edition are small enough that actually doing the upgrade is just low on my priority list. It's not an MSRV constraint, I am still using the latest stable rustc with edition 2021.

> It should have the same problem in the newly merged prompt macro's code, could you fix it together in this PR?

I have updated the `prompt` macro impl to match. Everything should be good to go now.

4t145 (approved):

4t145:
thanks

PR #409, opened 2025-09-01
David/ignore nonparseable stdout

vorporeal:
Ignore messages that don't parse as a JSONRPC object.

This will fix issues where we fail to start up spec non-compliant servers that write out status logs to stdout instead of stderr.


vorporeal:
oops, this was supposed to be merged into our fork, not upstream, sorry.

PR #416, opened 2025-09-03
Fix JSON-RPC request ID type to match specification

dsp-ant:
Fixes the issue that the Json RPC request id is u32. However the JSON RPC spec mentions that this is supposed to be a number, meaning it can be negative and is typically 64bit. So we need to use an i64 here.


4t145:
It looks like you are working on an outdated version, which hasn't been updated since half years ago?

dsp-ant:
Whoops. Somehow completely botched that. I'll update the PR. The problem is still valid that, since `NumberOrString` takes a u32 but likely needs to accept an i64

alexhancock (approved):

PR #421, opened 2025-09-05
Skip notification in initialization handshake

### LINKED ISSUE #412: Client implementation is not robust to ill-conforming servers during initialization handshake
**Describe the bug**

I'm a developer/maintainer of an application that's utilizing this crate for MCP client functionality.  We've gotten a couple reports from our users that their stdio-based MCP servers fail to start correctly.

After debugging, I've discovered two initialization failure modes (so far) that are fatal:
1. If the server prints out log lines to stdout instead of stderr, the handshake fails.
    * `@circleci/mcp-server-circleci` is an example of this - it prints out a message as soon as it starts, which `rmcp` tries to parse as an `InitializeResult`.
1. If the server sends notifications before the initialization handshake completes.
    * `effect-mcp` fails in this way - it emits `resources/list_changed` and `tools/list_changed` notifications as soon as it starts up.

**To Reproduce**

1. Try using this crate's stdio transport client support to connect to either of the above servers (`npx -y @circleci/mcp-server-circleci` or `npx -y effect-mcp`).
2. Notice that both of them will fail during the initialization handshake.

**Expected behavior**

In an ideal world, this crate would handle such errors more gracefully.

In our fork, I've made two changes to be more robust to these sorts of non-conforming servers:
1. In the `decode()` implementation for `JsonRpcMessageCodec`, we read a new message if `try_parse_with_compatibility()` returns an error, instead of propagating that error to the caller.
    * This fixes the first case (server logging to stdout instead of stderr).
    * The spec is very clear that this should not be necessary.  In the stdio transport [spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#stdio), it states:
      > The server **MUST NOT** write anything to its `stdout` that is not a valid MCP message.

      Despite this, it may be worth being resilient to servers that do not comply here.
1. In `serve_client_with_ct_inner`, I added a `loop` around the call to `expect_response` and the associated response checking logic, only breaking out of the loop once we receive an `InitializeResult`.
    * This fixes the second case (notifications emitted before the handshake completes).
    * The spec is less restrictive about server behavior here.  It states:
      > The server **SHOULD NOT** send requests other than pings and logging before receiving the `initialized` notification.

      Notably, it says "SHOULD NOT" and not "MUST NOT" - given a server is technically permitted to emit other valid JSONRPC messages before completing the initialization handshake, this library should probably be resilient to them.

I'm happy to open PRs for the two changes I've made in our fork, if that would be a helpful starting point for discussing how best to address these issues.

quangIO:
This allows supporting nonconforming MCP servers by skipping notification in initialization handshake.

Fix #412

## How Has This Been Tested?
Tested with DesktopCommanderMCP. All standard MCP servers continue to work normally.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -85,11 +85,6 @@ where
     T: Transport<RoleClient>,
     S: Service<RoleClient>,
 {
-    let notification_context = NotificationContext {
-        meta: Meta::new(),
-        extensions: Default::default(),
-        peer,
-    };
     loop {
         let message = expect_next_message(transport, context).await?;
         match message {
@@ -98,14 +93,27 @@ where
                 break Ok((result, id));
             }
             // Server could send logging messages before handshake
-            ServerJsonRpcMessage::Notification(notification)
-                if matches!(
-                    notification.notification,
-                    ServerNotification::LoggingMessageNotification(_)
-                ) =>
-            {
+            ServerJsonRpcMessage::Notification(mut notification) => {
+                let ServerNotification::LoggingMessageNotification(logging) =
+                    &mut notification.notification
+                else {
+                    tracing::warn!(?notification, "Received unexpected message");
+                    continue;
+                };
+
+                let mut context = NotificationContext {
+                    peer: peer.clone(),
+                    meta: Meta::default(),
+                    extensions: Extensions::default(),
+                };
+
+                if let Some(meta) = logging.extensions.get_mut::<Meta>() {
+                    std::mem::swap(&mut context.meta, meta);
+                }
+                std::mem::swap(&mut context.extensions, &mut logging.extensions);
+
                 if let Err(error) = service
-                    .handle_notification(notification.notification, notification_context.clone())
+                    .handle_notification(notification.notification, context)
                     .await
                 {
                     tracing::warn!(?error, "Handle logging before handshake failed.");
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -75,18 +75,51 @@ where
 }
 
 /// Helper function to expect a response from the stream
-async fn expect_response<T>(
+async fn expect_response<T, S>(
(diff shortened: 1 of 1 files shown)


4t145:
@vorporeal Could you help review this pr?

crates/rmcp/src/service/client.rs
```diff
@@ -82,11 +82,21 @@ async fn expect_response<T>(
...
-        _ => Err(ClientInitializeError::ExpectedInitResponse(Some(msg))),
+    loop {
+        let msg = expect_next_message(transport, context).await?;
+        match msg {
+            ServerJsonRpcMessage::Response(JsonRpcResponse { id, result, .. }) => {
+                break Ok((result, id));
+            }
+            ServerJsonRpcMessage::Notification(notification) => {
+                let notification = serde_json::to_string(&notification).unwrap_or_default();
+                tracing::debug!(
```
4t145: should we use handler's implementation to handle notifications?

quangIO: Could you point out how should we do that? 

Sorry I am not familiar with the code base enough so I try to keep it simple in this PR. Moreover, the type of MCP doing this doesn't follow the specs (EDIT: specs' recommendations). The connection is not consider established so we only debug log for courtesy. The handler assumes the connection is established right?

4t145: > The connection is not consider established so we only debug log for courtesy.

The spec allow both server and client send log and ping notification before initialization. But I am not sure about that should the peer side handle those notifications. I assume that some one may use a customized logger in handler's `on_log_notification` method, and that person would miss the log before initialization.

> Could you point out how should we do that?

We can pass the handler into this function and let the handler handle ping and log notifications. (If all of us think it's reasonable).

quangIO: Done! Please review them again.

4t145 (approved):

PR #435, opened 2025-09-11
Require CSRF token as part of the OAuth authorization flow.

vorporeal:
## Motivation and Context

The current logic discards the CSRF token that is generated as part of the OAuth authorization flow, instead of retaining it and verifying that the same token was provided back to the client by the authorization server.

This is an important step in OAuth to avoid CSRF attacks, and it is to the benefit of users of this library for it to require users to provide the token (ensuring that it gets validated).

## How Has This Been Tested?

I discovered this issue while integrating OAuth support, and decided it made more sense to make a change on the `rmcp` side instead of adding logic in our application to extract the `state` param from the authorization URL, and doing the equality check ourselves.

## Breaking Changes

Yes, users who support OAuth will need to modify their code to pass through the CSRF token when completing the authorization flow.

I assert this is a _good_ thing, as it improves their security posture.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/OAUTH_SUPPORT.md
+++ b/docs/OAUTH_SUPPORT.md
@@ -44,8 +44,9 @@ rmcp = { version = "0.1", features = ["auth", "transport-sse-client"] }
     println!("Please open the following URL in your browser for authorization:\n{}", auth_url);
     
     // Handle callback - In real applications, this is typically done in a callback server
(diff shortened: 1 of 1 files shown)


jokemanfire:
Thanks!

Issue #445, opened 2025-09-19
List tools fails when there are tools without parameters


Paul-E:
**Describe the bug**
The list tools MCP json RPC fails, returning errors like this

```json
  {
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      0,
      "inputSchema",
      "type"
    ],
    "message": "Invalid literal value, expected \"object\""
  }
```

**To Reproduce**

Start the example counter streamable HTTP server by navigating into the `examples/servers` directory and run `cargo run --example servers_counter_streamhttp`

Start up the MCP inspector with `npx @modelcontextprotocol/inspector`.

It will open your browser, or navigate in your browser to the URL given on the command line. Look for the text `MCP Inspector is up and running at:  http://localhost:6274/?MCP_PROXY_AUTH_TOKEN=`...

Enter `http://localhost:8000/mcp` into the "URL" on the inspector webpage. Navigage to "tools" and click "list tools". In the history, you will see an error response.

**Expected behavior**

I expect to see a list of valid tools. In this case it should be the tools in the `counter` example: `increment`, `decrement`, `get_value`, `say_hello`, `echo`, and `sum`. 

**Additional Note**

The tests against the examples do not compile when run with `cargo test --examples` at the top level.

glcrazier:
I believe that was because the inputSchema is an empty object which is validated as failure. The code should be modified not to output inputSchema when there is no inputSchema.

alexjpayne:
I'm observing this too. I followed @Paul-E 's steps to reproduce using @vincent0426 's forked version but unfortunately this did not resolve the problem.

Full output:

```json

  {
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      1,
      "inputSchema",
      "type"
    ],
    "message": "Invalid literal value, expected \"object\""
  },
  {
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      2,
      "inputSchema",
      "type"
    ],
    "message": "Invalid literal value, expected \"object\""
  },
  {
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      3,
      "inputSchema",
      "type"
    ],
    "message": "Invalid literal value, expected \"object\""
  },
  {
    "code": "invalid_literal",
    "expected": "object",
    "path": [
      "tools",
      4,
      "inputSchema",
      "type"
    ],
[...]

PR #446, opened 2025-09-21
fix: generate default schema for tools with no params

vincent0426:
## Motivation and Context

closes https://github.com/modelcontextprotocol/rust-sdk/issues/445

provide empty object json schema when there is no params for a tool, following https://github.com/modelcontextprotocol/typescript-sdk/blob/c94ba4b43cd305e39d88985c73d6b9bc1153da84/src/server/mcp.ts#L1202-L1205

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp-macros/src/tool.rs
+++ b/crates/rmcp-macros/src/tool.rs
@@ -199,9 +199,12 @@ pub fn tool(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream> {
                 rmcp::handler::server::common::cached_schema_for_type::<#params_ty>()
             })?
         } else {
-            // if not found, use a simple empty JSON object
+            // if not found, use a default empty JSON schema object
             syn::parse2::<Expr>(quote! {
-                std::sync::Arc::new(serde_json::Map::new())
+                std::sync::Arc::new(serde_json::json!({
+                    "type": "object",
+                    "properties": {}
+                }).as_object().unwrap().clone())
             })?
         }
     };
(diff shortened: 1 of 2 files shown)


jokemanfire:
If this pr has been merged , the empty struct will be updated , how about wait for the newest schema description release?
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/881/files#diff-b562e63d9fb1f6bbb6ddf2807a7f4e9558662e454b33076540a053ae55e46fc0

asm89:
@jokemanfire Ideally this fix is merged, because it seems like clients in the wild do expect the object type to be present. Servers using rmcp 0.7.0 are not compatible with many clients due to the `{}` being used to signal empty parametrs.

jokemanfire:
Then please add a todo comment. This should be updated according to the new specifications soon.

crates/rmcp-macros/src/tool.rs
```diff
@@ -199,9 +199,12 @@ pub fn tool(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream> {
                 rmcp::handler::server::common::cached_schema_for_type::<#params_ty>()
             })?
         } else {
-            // if not found, use a simple empty JSON object
+            // if not found, use a default empty JSON schema object
```
jokemanfire, line 202: just add a todo comment here.

alexhancock:
It LGTM now if also OK with @jokemanfire

The commit message `input: fix: generate default schema for tools with no params` could work

jokemanfire (approved):

vincent0426:
@jokemanfire any reason that CI doesn't pass? thanks

jokemanfire:
> @jokemanfire any reason that CI doesn't pass? thanks

It should be CI's problem ,I will check.

jokemanfire:
#458 fix rustdoc build by scoping

PR #451, opened 2025-09-23
Return auth errors

jamadeo:
When streamable HTTP servers return 401s, surface them to the client.

## Motivation and Context
This is necessary for clients to determine whether oauth is needed.

## How Has This Been Tested?
Tested with https://github.com/block/goose

## Breaking Changes
It adds a new error variant to StreamableHttpError

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -18,6 +18,11 @@ use crate::{
 
 type BoxedSseStream = BoxStream<'static, Result<Sse, SseError>>;
 
+#[derive(Debug)]
+pub struct AuthRequiredError {
+    pub www_authenticate_header: String,
+}
+
 #[derive(Error, Debug)]
 pub enum StreamableHttpError<E: std::error::Error + Send + Sync + 'static> {
     #[error("SSE error: {0}")]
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -284,12 +290,21 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
...
+            Ok(res) => {
+                let _ = responder.send(Ok(()));
+                res.expect_initialized::<C::Error>().await.map_err(
+                    WorkerQuitReason::fatal_context("process initialize response"),
+                )?
+            }
+            Err(err) => {
+                let _ = responder.send(Err(err));
+                return Err(WorkerQuitReason::fatal(
+                    StreamableHttpError::TransportChannelClosed,
```
alexhancock, line 304: this is the only bit I don't understand. why do we only return this here? are the other variants of `StreamableHttpError` not relevant?

jamadeo: This is an unfortunate result of the fact that you can't generally clone an Err. But it doesn't matter too much here: the returned error here is logged/thrown away when .close is called on the transport here: https://github.com/modelcontextprotocol/rust-sdk/blob/3310718b9c922a618007649eeaf9bc3d9a53fe80/crates/rmcp/src/transport/worker.rs#L208. The important part is the line before, where we .send() it to the caller.

That said I could put an actual message here, which would improve it

alexhancock: All good. Makes sense!

Issue #461, opened 2025-09-29
OAuth: discovery is mandatory


Pvlerick:
**Describe the bug**
There is not way to use OAuth without going through the discovery phase; however the discovery phase might crash as the ```AuthorizationMetadata``` struct does not follow the RFC 8414: registration_endpoint is optionan, its absence meaning that there is no automatic registration available.

It could be possible to go down and use ```AuthorizationManager``` manually, but non of the fields on that struct are pub, so it's in effect impossible to use any of the mechanics unless resorting to some kind of mocking of the http client, which isn't a good idea.

**To Reproduce**
1. provide a .well-known/oauth-authorization-server endpoint that does not contain the registration_endpoint  element, it will then fail here: https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/rmcp/src/transport/auth.rs#L227

**Expected behavior**
There should be a way to provide OAuth configuration data without using the discovery.

**Logs**
```
Metadata error: Failed to parse metadata: error decoding response body
```

**Additional context**
Ideally, it should be possible to have the OAuth module discover the different endpoints at the well known location, but provide the registration via configuration, if the resource does not allow automatic registration.

jokemanfire:
registration_endpoint should be optional , that's a mistake, I will change it first.

DaleSeo:
This was addressed in #463. Closing as resolved.

PR #463, opened 2025-09-30
fix(oauth): The dynamic client registration should be optional

jokemanfire:
The dynamic registration can be skipped, and it should be optional.
## Motivation and Context
Accoding to rfc8414.

## How Has This Been Tested?
No test

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -14,7 +14,7 @@ use reqwest::{Client as HttpClient, IntoUrl, StatusCode, Url, header::AUTHORIZAT
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::{Mutex, RwLock};
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 const DEFAULT_EXCHANGE_URL: &str = "http://localhost";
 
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -290,7 +290,7 @@ impl AuthorizationManager {
         Ok(AuthorizationMetadata {
             authorization_endpoint: create_endpoint("authorize"),
             token_endpoint: create_endpoint("token"),
-            registration_endpoint: create_endpoint("register"),
+            registration_endpoint: None,
```
alexhancock, line 293: Given that supporting DCR is a SHOULD in the MCP spec, If we cannot get metadata at all and we construct these defaults, should we continue to populate a default uri here?

jokemanfire: It aimed to achieve better compatibility with the RFC 8414 protocol, according to the rfc 8414 , it should be optional 
image
.
And in the spec file ,the registrat should be alt in the picture, and  user can setting the  client ID by himself which we provide this interface ,  
image

vorporeal: GitHub, for example, does not provide a value for `registration_endpoint` in their response, as they do not support DCR and require pre-registration of clients.  (I had to make this same change in a fork, to support GitHub properly, and had not yet had an opportunity to open a PR here.)

jokemanfire:
#461

jokemanfire:
@4t145 Did you forget submit your review? :）

4t145:
approved

crates/rmcp/src/transport/auth.rs
alexhancock: I see

alexhancock:
LGTM as well after adjusting commit message to satisfy the linter

jokemanfire:
> LGTM as well after adjusting commit message to satisfy the linter

Change it while merge .

PR #486, opened 2025-10-15
Don't treat non-success HTTP codes as transport errors

SteffenDE:
The spec states that:

> If the server cannot accept the input, it MUST return an HTTP error
> status code (e.g., 400 Bad Request).
> The HTTP response body MAY comprise a JSON-RPC error response that has no id.

But the rust-sdk treated any POST responses with non successful status as an error.

Before this patch, trying to call list_resources on a server that does return HTTP 400 with JSON-RPC error:

```
TransportSend(
    DynamicTransportError {
        transport_name: "rmcp::transport::worker::WorkerTransport<rmcp::transport::streamable_http_client::StreamableHttpClientWorker<reqwest::async_impl::client::Client>>",
        transport_type_id: TypeId(0xf7bacf3cf3889d471ead32d7a3cc8155),
        error: Client(
            reqwest::Error {
                kind: Status(
                    400,
                    None,
                ),
                url: "http://localhost:4000/mcp",
            },
        ),
    },
)
```

After this patch:

```
McpError(
    ErrorData {
        code: ErrorCode(
            -32601,
        ),
        message: "Method not found",
        data: Some(
            Object {
                "name": String("resources/list"),
            },
        ),
    },
)
```

## Motivation and Context

Clients should be able to see the error the server sends. The code already checks for JSON content type, I assume the error_for_status for included by accident.

## How Has This Been Tested?

Tests still seem to pass, but I did not add a new one. For testing, I created a demo client, but someone with more knowledge of the code should add a proper test case.

```rust
use anyhow::Result;
use rmcp::{
    ServiceExt,
    model::{ClientCapabilities, ClientInfo, Implementation},
    transport::StreamableHttpClientTransport,
};
use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};

#[tokio::main]
async fn main() -> Result<()> {
    tracing_subscriber::registry()
        .with(
            tracing_subscriber::EnvFilter::try_from_default_env()
                .unwrap_or_else(|_| format!("debug,{}=debug", env!("CARGO_CRATE_NAME")).into()),
        )
        .with(tracing_subscriber::fmt::layer())
        .init();

    let transport = StreamableHttpClientTransport::from_uri("http://localhost:4000/mcp");
    let client_info = ClientInfo {
        protocol_version: Default::default(),
        capabilities: ClientCapabilities::default(),
        client_info: Implementation {
            name: "test unknown method client".to_string(),
            title: None,
            version: "0.0.1".to_string(),
            website_url: None,
            icons: None,
        },
    };

    let client = client_info.serve(transport).await.inspect_err(|e| {
        tracing::error!("client error during connection: {:?}", e);
    })?;

    let server_info = client.peer_info();
    tracing::info!("Connected to server: {server_info:#?}");

    tracing::info!("Calling list_resources (which the server doesn't support)...");
    match client.list_resources(Default::default()).await {
        Ok(result) => {
            tracing::info!("Success: {result:#?}");
        }
        Err(e) => {
            tracing::error!("Error calling list_resources: {e:#?}");
        }
    }

    client.cancel().await?;
    Ok(())
}
```

## Breaking Changes

Could be seen as a breaking change if people relied on getting the transport error.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/common/reqwest/sse_client.rs
+++ b/crates/rmcp/src/transport/common/reqwest/sse_client.rs
@@ -33,7 +33,6 @@ impl SseClient for reqwest::Client {
         request_builder
             .send()
             .await
-            .and_then(|resp| resp.error_for_status())
             .map_err(SseTransportError::from)
             .map(drop)
     }
--- a/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
@@ -118,7 +118,6 @@ impl StreamableHttpClient for reqwest::Client {
                 }));
             }
         }
-        let response = response.error_for_status()?;
         if response.status() == reqwest::StatusCode::ACCEPTED {
             return Ok(StreamableHttpPostResponse::Accepted);
         }


4t145:
Good catch, but should we have some log when response status is not success?

SteffenDE:
I don’t think there’s a need to log, since it’s part of the spec to have non-200 codes and a client will still receive this as an error, so users can decide to log if they want.

SteffenDE:
@alexhancock No worries! We're still using the SSE code, so I don't want to update that branch for now. As it's only one line to remove

https://github.com/modelcontextprotocol/rust-sdk/blob/95d3e3f9403ea853f18ef692143cecf8bc156b20/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs#L124

I think it's fine if I just close the PR and you do that change directly :)

PR #494, opened 2025-10-21
fix(streamable-http): gracefully shutdown while client connected

### LINKED ISSUE #493: Gracefully shutdown is hang while a SSE connection is established
**Describe the bug**
While using `streamable-http` transport. The `SseBody` donnot handle the signal of gracefully shutdown, thus, blocking the shutdown process until client disconnect.

**To Reproduce**
1. start the server
```bash
RUST_LOG=trace cargo run --example servers_counter_streamhtt
```

2. create a client connection to server via `@modelcontextprotocol/inspector`

3. press `CTRL` + `C` to trigger server exits

Image

> I noted the `sse` transport has similar issue, but I don't dig it deep.

**Expected behavior**
While trigger `cancelled`, the server will exits immediately.

**Logs**

**Additional context**

greenhat616:
Close #493 

I’m adding a cancellation token to Streamable HTTP to support graceful shutdowns. Because I’m unclear on the MCP protocol’s semantics for initiating termination, I’m currently just cutting off the body stream.

## Motivation and Context

In desktop apps, we need a mechanism for the process to proactively terminate (e.g., for updates or user-initiated quits). Today, the MCP Client holds the connection open, which effectively prevents the app from shutting down; most users aren’t aware this is intentional client behavior.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/examples/transport/src/named-pipe.rs
+++ b/examples/transport/src/named-pipe.rs
@@ -12,11 +12,11 @@ async fn main() -> anyhow::Result<()> {
         let mut server = ServerOptions::new()
             .first_pipe_instance(true)
             .create(name)?;
-        while let Ok(_) = server.connect().await {
+        while server.connect().await.is_ok() {
             let stream = server;
             server = ServerOptions::new().create(name)?;
             tokio::spawn(async move {
-                match serve_server(Calculator, stream).await {
+                match serve_server(Calculator::new(), stream).await {
                     Ok(server) => {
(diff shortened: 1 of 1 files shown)


greenhat616:
I checked the doc: <https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#shutdown>

It seems that trigger a connection close is enough

PR #509, opened 2025-11-04
fix(oauth): fix oauth credential refresh when the original token exchange was done in a different process

gpeal:
## Motivation and Context
Codex has gotten internal and external reports of MCP servers frequently logging out (https://github.com/openai/codex/issues/6164).

I tracked it down to the fact that auth.rs was using an in-memory expires-at which is only set on initial token exchange. Instead, this PR switches it to use the expires-at set in the credentials that are passed in.

## How Has This Been Tested?
Validated with the notion mcp that it had an expired token, did _not_ refresh the token before and does now.

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1,8 +1,4 @@
-use std::{
-    collections::HashMap,
-    sync::Arc,
-    time::{Duration, Instant},
-};
+use std::{collections::HashMap, sync::Arc, time::Duration};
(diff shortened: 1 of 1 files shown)


enriquemorenotent:
The problem persists in 0.54

alexhancock:
@enriquemorenotent Do you mean 0.8.4 and above?

cc @gpeal

PR #538, opened 2025-11-12
fix: add token update callback

alexhancock:
Code originally from https://github.com/ianschenck

@ianschenk The original patch you had linked didn't apply, but I made an updated version. I cannot seem to add you as a reviewer, but take a look and let me know if it's faithful.

Original bug report https://github.com/block/goose/issues/5259

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -146,6 +146,33 @@ type OAuthClient = oauth2::Client<
 >;
 type Credentials = (String, Option<OAuthTokenResponse>);
 
+/// Callback trait for OAuth token updates
+///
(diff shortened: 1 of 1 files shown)


jokemanfire:
I feel confuse ,the interface refresh_token ,store the new token ,what's the problem?

```
   /// refresh access token
    pub async fn refresh_token(
        &self,
    ) -> Result<StandardTokenResponse<EmptyExtraTokenFields, BasicTokenType>, AuthError> {
....
        // store new credentials
        *self.credentials.write().await = Some(token_result.clone());
...
```

alexhancock:
Ah yes, I see that now. Will continue to debug https://github.com/block/goose/issues/5259. It must be some issue with how we're using the refresh_token method.

alexhancock:
I have a line on a better scoped solution and will reopen something new

PR #541, opened 2025-11-14
fix: exclude WASI example from workspace to avoid linker errors 

tanish111:
## Motivation and Context
Cargo uses the macOS host linker for workspace builds, but that linker cannot 
resolve WASI WebAssembly imports (such as _wasi:cli/run@0.2.x). As a result, 
including a WASI crate in the workspace causes builds to fail on macOS systems 
using arm64.

Solution: Mark the WASI example as non-default by excluding it from the workspace 
members. This makes the WASI package optional, so `cargo build` won't touch it,
and it can be built explicitly when needed with:
  `cargo build -p wasi-mcp-example --target wasm32-wasip2`

## How Has This Been Tested?
Now cargo build works fine on Mac/arm64

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
CI will continue to build the WASI example explicitly during release workflows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 members = ["crates/rmcp", "crates/rmcp-macros", "examples/*"]
-exclude = ["examples/wasi"]
+default-members = ["crates/rmcp", "crates/rmcp-macros"]
 resolver = "2"
 
 [workspace.dependencies]
--- a/examples/wasi/Cargo.toml
+++ b/examples/wasi/Cargo.toml
@@ -1,18 +1,15 @@
 [package]
 name = "wasi-mcp-example"
-edition = "2024"
-version = "0.8.5"
-authors = ["4t145 <u4t145@163.com>"]
-license = "MIT"
-repository = "https://github.com/modelcontextprotocol/rust-sdk/"
-description = "Rust SDK for Model Context Protocol"
-keywords = ["mcp", "sdk", "tokio", "modelcontextprotocol"]
-homepage = "https://github.com/modelcontextprotocol/rust-sdk"
-categories = [
-    "network-programming",
-    "asynchronous",
-]
-readme = "README.md"
+edition = { workspace = true }
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+description = { workspace = true }
+keywords = { workspace = true }
+homepage = { workspace = true }
+categories = { workspace = true }
+readme = { workspace = true }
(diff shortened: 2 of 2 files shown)


alexhancock (dismissed):

alexhancock:
@tanish111 Need to look into:

```
Building examples/wasi/
error: failed to parse manifest at `/home/runner/work/rust-sdk/rust-sdk/examples/wasi/Cargo.toml`

Caused by:
  error inheriting `edition` from workspace root manifest's `workspace.package.edition`

Caused by:
  failed to find a workspace root
```

tanish111:
@alexhancock The problem was that examples/wasi/Cargo.toml used { workspace = true },
which fails when the package is excluded from the workspace. Replaced all
workspace inheritance with explicit values. This should fix the build issue.

alexhancock:
Can we look into not excluding the wasi example from the workspace, but instead using `members` and `default-members`?

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

If we add everything to `members`  but exclude the wasi example from `default-members` I think we might get what we want in terms of clean local builds without having to remove it from the workspace entirely like this?

tanish111:
@alexhancock we can do that but in that case we would need to maintain a long list of `default-members` like this
```
default-members = [
    "crates/rmcp",
    "crates/rmcp-macros",
    "examples/clients",
    "examples/rig-integration",
    "examples/servers",
    "examples/simple-chat-client",
    "examples/transport",
]
```
Moreover whenever someone adds new example he would also need to make changes to Cargo.toml. Is that what we want ?

alexhancock:
I think there are other problems with doing it this way, like all the info is now duplicated into the wasi example's `Cargo.toml` and needs to be kept up to date. For example, it wouldn't automatically rely on new versions of rmcp and we'd need to manually version bump it with every change to rmcp to make sure it still works.

I think the best solution is probably:

```
members = ["crates/rmcp", "crates/rmcp-macros", "examples/*"]
default-members = ["crates/rmcp", "crates/rmcp-macros"]
```

this way, only "crates/rmcp", "crates/rmcp-macros"` will be operated on by default with local commands, then in CI we can make sure we run tests/etc against everything. Then the wasi example can be part of the workspace without causing any issue.

tanish111:
@alexhancock got it
Avoiding all examples from default workspace build makes sense

tanish111:
@alexhancock I have made the changes

tanish111:
@alexhancock @jokemanfire Thanks

PR #542, opened 2025-11-14
feat(auth): CredentialStore trait

alexhancock:
**Description**

Adds the concept of a configurable `CredentialStore` for `AuthorizationManager` to use

The current implementation writes to `credentials` in memory. This change gives the flexibility for users of `AuthorizationManager` to provide a custom implementation that persists credentials (to keychain, disk, etc).

There is no change to default behavior, as a default implementation is provided which does the same as now (`InMemoryCredentialStore`)

**Motivation**

The motivation came from this bug report to goose that new refresh tokens sent from the server during token exchange were not being saved/used https://github.com/block/goose/issues/5259

We store access tokens and refresh tokens via `keyring`, but we only did it on the initial auth and previously had no way to be informed of new refresh tokens received when `rmcp` exchanges tokens.

This allows for a provided `CredentialStore` to always load/save/clear token information from the storage mechanism of choice, and will fix https://github.com/block/goose/issues/5259

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -14,6 +14,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
+async-trait = "0.1.89"
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 thiserror = "2"
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
+use async_trait::async_trait;
 use oauth2::{
     AuthUrl, AuthorizationCode, ClientId, ClientSecret, CsrfToken, EmptyExtraTokenFields,
     PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, RefreshToken, RequestTokenError, Scope,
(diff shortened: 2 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -152,6 +208,7 @@ pub struct AuthorizationManager {
     metadata: Option<AuthorizationMetadata>,
     oauth_client: Option<OAuthClient>,
     credentials: RwLock<Option<OAuthTokenResponse>>,
+    credential_store: Arc<dyn CredentialStore>,
```
jamadeo, line 211: I think I would expect the credential_store to replace the `credentials: RwLock<Option<OAuthTokenResponse>>`, and for reads/writes to self.credentials to be replaced by the same to self.credential_store

Then the application doesn't call load/save itself, it just provides the implementation

crates/rmcp/src/transport/auth.rs
```diff
@@ -223,13 +281,56 @@ impl AuthorizationManager {
...
+            client_id,
+            token_response,
+        };
+
+        self.credential_store.save(stored).await
+    }
+
+    pub async fn clear_credentials(&self) -> Result<(), AuthError> {
+        self.credential_store.clear().await
+    }
```
jamadeo, line 316: I think if you use the credential store as a replacement for the current `self.credentials`, you don't need these methods, because the caller would only decide how, not when, to load/save

crates/rmcp/Cargo.toml
```diff
@@ -14,6 +14,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
+async-trait = "0.1.89"
```
jamadeo, line 17: not sure how big a concern it is to have this, but if you want to avoid adding this dependency you can have the trait return Futures

crates/rmcp/src/transport/auth.rs
alexhancock: Ah not removing `credentials` was an oversight. I will update.

I agree conceptually about 

> Then the application doesn't call load/save itself, it just provides the implementation

Will look at an update

alexhancock: Done

crates/rmcp/src/transport/auth.rs
alexhancock: Done

jamadeo (dismissed):

crates/rmcp/Cargo.toml
jokemanfire: I often use this dependency and I am planning to introduce it in the future to make the code more concise. Do you want to change other implementations to this dependency implementation as well？

jokemanfire (dismissed):
LGTM

crates/rmcp/Cargo.toml
alexhancock: Yes, we can use it more - it seems nice

Issue #518, opened 2025-11-07
Implement SEP-986: Tool Name Guidance


felixweinberger:
This is a tracking issue for implementation of [SEP-986](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986).

### Summary

This SEP establishes clear guidelines and format specifications for tool naming conventions within the MCP protocol. The Rust SDK needs to implement validation for tool name formats, provide clear error messages when tool names don't follow the specified patterns, and ensure consistency with the standardized naming conventions to improve cross-implementation compatibility and prevent naming conflicts.

### Related Issues & PRs

- **Implementation PRs**: n/a
- **Related PRs**: n/a
- **Related Issues**: n/a

tanish111:
@alexhancock @felixweinberger Can I work on this?

alexhancock:
@tanish111 Yeah absolutely. Mostly just warning if a tool name doesn't conform to the SHOULD guidelines

tanish111:
@alexhancock I have presented a possible implementation in the lines of [typescript implementation](https://github.com/modelcontextprotocol/typescript-sdk/pull/900) in PR #551

PR #551, opened 2025-11-18
Implementation of SEP-986: Specify Format for Tool Names

tanish111:
## Motivation and Context
[modelcontextprotocol/modelcontextprotocol/issues/986](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986)
## How Has This Been Tested?
Manual testing in examples/servers: added name to existing #[tool] attributes in common/counter.rs with valid and invalid names. Rebuilding and running showed expected warnings. Unit tests (11) cover validation scenarios; integration tests are planned.

## Breaking Changes
None, but console warnings may appear for SDK users who have names that do not satisfy the SEP.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
All design decisions, names, and other aspects follow the [TypeScript implementation](https://github.com/modelcontextprotocol/typescript-sdk/pull/900) for consistency across SDKs.
Related issue #518

### DIFF
--- a/crates/rmcp/src/handler/server/tool_name_validation.rs
+++ b/crates/rmcp/src/handler/server/tool_name_validation.rs
@@ -10,22 +10,22 @@ use std::collections::HashSet;
 
 /// Result of tool name validation containing validation status and warnings.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ToolNameValidationResult {
+struct ToolNameValidationResult {
     /// Whether the tool name is valid according to the specification
-    pub is_valid: bool,
+    is_valid: bool,
     /// Array of warning messages about non-conforming aspects of the tool name
-    pub warnings: Vec<String>,
+    warnings: Vec<String>,
 }
 
 impl ToolNameValidationResult {
     /// Create a new validation result
-    pub fn new(is_valid: bool, warnings: Vec<String>) -> Self {
+    fn new(is_valid: bool, warnings: Vec<String>) -> Self {
         Self { is_valid, warnings }
     }
 }
 
 /// Validates a tool name according to the SEP specification.
-pub fn validate_tool_name(name: &str) -> ToolNameValidationResult {
+fn validate_tool_name(name: &str) -> ToolNameValidationResult {
     let mut warnings = Vec::new();
 
     // Check length
@@ -108,7 +108,7 @@ pub fn validate_tool_name(name: &str) -> ToolNameValidationResult {
 }
 
 /// Issues warnings for non-conforming tool names.
-pub fn issue_tool_name_warning(name: &str, warnings: &[String]) {
+fn issue_tool_name_warning(name: &str, warnings: &[String]) {
     tracing::warn!("Tool name validation warning for \"{}\":", name);
     for warning in warnings {
         tracing::warn!("  - {}", warning);


crates/rmcp/src/handler/server/tool_name_validation.rs
```diff
@@ -0,0 +1,312 @@
...
+//! Tool names are case-sensitive.
+//! Allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits
+//! (0-9), underscore (_), dash (-), and dot (.).
+//! Tool names SHOULD NOT contain spaces, commas, or other special characters.
+
+use std::collections::HashSet;
+
+/// Result of tool name validation containing validation status and warnings.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ToolNameValidationResult {
```
pbezglasny: I guess it would better to use enum here, like
```rust
pub enum ToolNameValidationResult{
 Ok,
 NonValid(Vec<String>)
}
```
In current implementation it possible create result with `is_valid=true` and non empty `warnings`. Or use simply use `std::Result<(), Vec<String>>`

tanish111: @pbezglasny I haven't seen any such case where `is_valid=true `and we have non empty `warnings`

tanish111: @pbezglasny Wouldn’t the enum approach make the Rust implementation inconsistent with the [TypeScript SDK](https://github.com/modelcontextprotocol/typescript-sdk/pull/900/files)? In TypeScript, the valid case still returns an empty list of warnings.

pbezglasny: Ah, I see that is possible variant when is_valid = true, but not empty warnings
https://github.com/modelcontextprotocol/typescript-sdk/blob/2e67eb552e2743b30c4efd551d95e9e49ef7c953/src/shared/toolNameValidation.ts#L41-L50
So it's allowed tools to contains whitespaces, but not recommended
`Tool names SHOULD NOT contain spaces, commas, or other special characters.`

crates/rmcp/src/handler/server/tool_name_validation.rs
```diff
@@ -0,0 +1,312 @@
...
+/// use rmcp::handler::server::tool_name_validation::validate_and_warn_tool_name;
+///
+/// let is_valid = validate_and_warn_tool_name("my_tool");
+/// assert!(is_valid);
+/// ```
+pub fn validate_and_warn_tool_name(name: &str) -> bool {
+    let result = validate_tool_name(name);
+
+    // Always issue warnings for any validation issues (both invalid names and warnings)
+    issue_tool_name_warning(name, &result.warnings);
```
alexhancock: I would check if there are warnings here and only call the method if there are warnings as opposed to having that check internally

tanish111: @alexhancock I have updated the code with this suggestion.
(This would definitely save unnecessary stack calls) 
Also I have remove most of comments just kept the one which are needed

tanish111: @alexhancock can I resolve this conversation ?

crates/rmcp/src/handler/server/router/tool.rs
```diff
@@ -219,7 +220,11 @@ where
     }
 
     pub fn add_route(&mut self, item: ToolRoute<S>) {
-        self.map.insert(item.attr.name.clone(), item);
+        let new_name = &item.attr.name;
+        // Validate tool name according to SEP specification
```
alexhancock: I'd remove these comments as it already seems clear based on method name

alexhancock (approved):

crates/rmcp/src/handler/server/router/tool.rs
tanish111: @alexhancock I have updated add_route fn

crates/rmcp/src/handler/server/tool_name_validation.rs
```diff
@@ -0,0 +1,261 @@
...
+
+impl ToolNameValidationResult {
+    /// Create a new validation result
+    pub fn new(is_valid: bool, warnings: Vec<String>) -> Self {
+        Self { is_valid, warnings }
+    }
+}
+
+/// Validates a tool name according to the SEP specification.
+pub fn validate_tool_name(name: &str) -> ToolNameValidationResult {
```
alexhancock: these don't need to be `pub` right?

only `validate_and_warn_tool_name`

tanish111: @alexhancock I kept it pub because I wanted to keep it consistent with [TS SDK](https://github.com/kentcdodds/typescript-sdk/blob/5c3ceb3d9e2c482e777b3c3e021e54f4efaf0e03/src/shared/toolNameValidation.ts#L21)

alexhancock: Given that the tool name validation is only for the purpose of logging warnings right now, I think we should keep the other things private

crates/rmcp/src/handler/server/tool_name_validation.rs
```diff
@@ -0,0 +1,261 @@
...
+//! Tool names are case-sensitive.
+//! Allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits
+//! (0-9), underscore (_), dash (-), and dot (.).
+//! Tool names SHOULD NOT contain spaces, commas, or other special characters.
+
+use std::collections::HashSet;
+
+/// Result of tool name validation containing validation status and warnings.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ToolNameValidationResult {
```
tanish111: @alexhancock should I change this to `private` as well as It's not used anywhere outside this file?(and all its fields and functions)

alexhancock: yes. only things which are used outside the module should be private IMO. easier to expose things later if we want vs take things private.

tanish111: > yes. only things which are used outside the module should be private IMO. easier to expose things later if we want vs take things private.

you mean only things which are used outside the module should be **public**

tanish111: @alexhancock fixed.

alexhancock: yes sorry typoed. it looks good to me now.

PR #556, opened 2025-11-21
feat: add support for custom client notifications

bolinfest:
MCP servers, particularly ones that offer "experimental" capabilities,
may wish to handle custom client notifications that are not part of the
standard MCP specification. This change introduces a new
`CustomClientNotification` type that allows a server to process
such custom notifications.

- introduces `CustomClientNotification` to carry arbitrary methods/params while
  still preserving meta/extensions; wires it into the `ClientNotification` union
  and `serde` so `params` can be decoded with `params_as`
- allows server handlers to receive custom notifications via a new
  `on_custom_notification` hook
- adds integration coverage that sends a custom client notification end-to-end
  and asserts the server sees the method and payload

Test:

```shell
cargo test -p rmcp --features client test_custom_client_notification_reaches_server
```

### DIFF
--- a/crates/rmcp/src/handler/server.rs
+++ b/crates/rmcp/src/handler/server.rs
@@ -89,6 +89,9 @@ impl<H: ServerHandler> Service<RoleServer> for H {
             ClientNotification::RootsListChangedNotification(_notification) => {
                 self.on_roots_list_changed(context).await
             }
+            ClientNotification::CustomClientNotification(notification) => {
+                self.on_custom_notification(notification, context).await
+            }
         };
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


alexhancock:
@bolinfest Mind taking a look at the failing checks? Thanks

bolinfest:
@alexhancock I believe I fixed the test and clippy issues, and I think the code coverage job failed due to the test failure, so could you please let CI run again?

bolinfest:
> It does leave me with a question why we wouldn't generalize it to also handle server -> client. What do you think?

Yes, I agree this should also be supported.

I'm happy to help with that, but I would like to move Codex off my custom branch of `rust-sdk` first:

https://github.com/openai/codex/blob/a421eba31f45d162e20b2c78ac899f78f559e3f6/codex-rs/Cargo.toml#L291

Issue #557, opened 2025-11-23
Missing resource-level metadata to listResources (lacks parity with readResource contents meta)


athola:
## Describe the bug
Resources returned by `resources/list` (`RawResource` / `Resource`) lack an extensible `_meta` field. This forces servers to embed provenance data (e.g., location=global/project/universal) into less suitable fields like `description` or URIs. Ultimately this creates inconsistency, as `readResource` allows `_meta` on `ResourceContents`, preventing clients that only list resources from accessing structured metadata.

## To Reproduce
1. Implement an MCP server that wants to tag each resource with `location`.
2. Return resources via `listResources`.
3. Observe there is no field to attach structured metadata; only `description` or name can be overloaded.

## Expected behavior
`listResources` should allow an optional structured metadata map (e.g., `_meta` / `meta: Meta`) on each resource, matching how `ResourceContents` supports `_meta`. Clients should receive structured fields like `{"location":"global"}` without string scraping.

## Logs
Not a runtime error; this is a protocol/schema limitation observed during server implementation.

## Additional context
- Adding optional `_meta` preserves backward compatibility (omit when empty).
- Aligns with other SDKs that allow resource metadata.
- Alternatives (description/URI query) are brittle, change resource identity, and require heuristic parsing.

## Related issues

- https://github.com/modelcontextprotocol/rust-sdk/issues/505 mentions adding `_meta` to `Tool` registration, but does not cover the need for resource-level `_meta` in `listResources`.
- https://github.com/modelcontextprotocol/rust-sdk/issues/61 describes allowing servers to attach arbitrary metadata to requests such that handlers are able to access client-related details (such as transport type or bearer token). It doesn't provide the ability to add `_meta` to `listResources` or resources. The scope with this issue is limited to request context rather than including resource metadata.

PR #558, opened 2025-11-24
Added _meta field to prompts, resources and paginated results

pbezglasny:
## Motivation and Context

MCP specifies _meta field for requested entities. Currently `_meta` field implemeted for tool, but missing for others.

This PR:
- adds `_meta` field for [Prompt](https://modelcontextprotocol.io/specification/2025-06-18/schema#prompt), [Resource](https://modelcontextprotocol.io/specification/2025-06-18/schema#resource)
- adds `_meta` field for PaginatedResult: ListToolsResult, ListPromptsResult, ListResourceTemplatesResult, ListPromptsResult.
- adds the `meta` attribute to tool_handler and prompt_handler macros
- updates Counter example to include example of usage new attributes(there is an issues with RustRover to parse attributes, but examples are working)

Format of meta attribute is the same as existed for tool. 
https://github.com/modelcontextprotocol/rust-sdk/blob/2b60f8af0ffccf6ec9f21ac8f95feecfc28dea18/crates/rmcp-macros/src/tool.rs#L127-L131
In this case you need to provide Meta object, but we can update to something like
```rust
        let meta = if let Some(meta) = meta {
            quote! { rmcp::model::Meta(rmcp::object!(#meta)) }
        } else {
            quote! { None }
        };
```
and use with item definition only with json, for instance
```rust
    #[prompt(
        name = "example_prompt",
        meta = {"meta_key": "meta_value"}
    )]
    async fn example_prompt(
```

## How Has This Been Tested?

- Code was tested via modelcontextprotocol/inspector.

## Breaking Changes

No

Closes https://github.com/modelcontextprotocol/rust-sdk/issues/557

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/examples/servers/src/common/counter.rs
+++ b/examples/servers/src/common/counter.rs
@@ -110,7 +110,10 @@ impl Counter {
 #[prompt_router]
 impl Counter {
     /// This is an example prompt that takes one required argument, message
-    #[prompt(name = "example_prompt")]
+    #[prompt(
+        name = "example_prompt",
+        meta = Meta(rmcp::object!({"meta_key": "meta_value"}))
+    )]
     async fn example_prompt(
         &self,
         Parameters(args): Parameters<ExamplePromptArgs>,
@@ -161,8 +164,8 @@ impl Counter {
     }
 }
 
-#[tool_handler]
-#[prompt_handler]
+#[tool_handler(meta = Meta(rmcp::object!({"tool_meta_key": "tool_meta_value"})))]
+#[prompt_handler(meta = Meta(rmcp::object!({"router_meta_key": "router_meta_value"})))]
 impl ServerHandler for Counter {
     fn get_info(&self) -> ServerInfo {
         ServerInfo {
@@ -188,6 +191,7 @@ impl ServerHandler for Counter {
                 self._create_resource_text("memo://insights", "memo-name"),
             ],
             next_cursor: None,
+            meta: None,
         })
     }
 
@@ -226,6 +230,7 @@ impl ServerHandler for Counter {
         Ok(ListResourceTemplatesResult {
             next_cursor: None,
             resource_templates: Vec::new(),
+            meta: None,
         })
     }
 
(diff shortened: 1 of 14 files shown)


examples/servers/src/common/counter.rs
```diff
@@ -30,7 +30,7 @@ pub struct ExamplePromptArgs {
 #[derive(Debug, serde::Serialize, serde::Deserialize, schemars::JsonSchema)]
 pub struct CounterAnalysisArgs {
     /// The target value you're trying to reach
-    pub goal: i32,
+    pub goal: String,
```
pbezglasny: Changed type to String, because in standard arguments are strings https://modelcontextprotocol.io/specification/2025-06-18/schema#prompts/get

athola:
LGTM. if there is a requirement to update associated tests or documentation I would encourage doing so but this appears to address my ticket here: https://github.com/modelcontextprotocol/rust-sdk/issues/557

examples/servers/src/common/counter.rs
jokemanfire: It means the args support prompt must be string? I think there must be some problem,  or we should convert very type to String in prompt macro, but not enforcing the args's type should be String?

pbezglasny: > It means the args support prompt must be string? 

I suppose, now there is error if using mcpinspector when calling `counter_analysis` tool, e.g. using `examples/counter_streamhttp` server. Screenshot attached

Image

> I think there must be some problem, or we should convert very type to String in prompt macro, but not enforcing the args's type should be String?

Using macro would be good solution, at least for primitive types, now it's possible to use any type with jsonschema. 
Frankly speeking, I changed type to string because I was testing with inspector, but now I thinks it would be better to revert change type and changes in different PR.

pbezglasny: I reverted field type change. In other sdks(typescript, python) is allowed to define not string arguments of prompt the same way as rust sdk, despite string type requirements. There was issue in inspector https://github.com/modelcontextprotocol/inspector/issues/869 that stands that client should send string arguments

jokemanfire (approved):

PR #561, opened 2025-11-25
docs: mark SSE transport examples as deprecated

domdomegg:
## Summary
- Moves SSE transport examples to a collapsed `<details>` section at the bottom of the example list
- Removes SSE examples from the 'How to Run' section
- Adds a note explaining that SSE has been removed from newer versions of the MCP spec and that streamable HTTP is preferred

SSE transport is less reliable than streamable HTTP and has been removed from newer versions of the MCP specification.

### DIFF
--- a/examples/servers/README.md
+++ b/examples/servers/README.md
@@ -11,23 +11,6 @@ A basic MCP server that communicates using standard input/output transport.
 - Provides a simple counter tool with increment, decrement, and get_value operations
 - Demonstrates basic tool implementation and stdio transport
 
-### Counter SSE Server (`counter_sse.rs`)
-
-A server that provides counter functionality using Server-Sent Events (SSE) transport.
-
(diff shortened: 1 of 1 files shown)


domdomegg:
> @domdomegg thoughts on timeline for full deletion?

@alexhancock I think we definitely could do this if you'd be open to it. SSE was removed from the spec quite some time ago now and I don't think anyone should be creating new SSE servers.

I can see it being useful for people maintaining older SSE servers. But there might just not really be any (from Claude.ai traffic I don't think there are any popular rust SSE servers AFAIK). Again, up to you!

Happy to put a PR up to remove if you think we can :)

alexhancock:
@domdomegg Yeah I am comfortable scorched-earthing SSE stuff now if you want - feels like time...

If someone really wants to do SSE in rust they can rely on an older version of the package until they switch to streamable http.

PR #562, opened 2025-11-25
feat!: remove SSE transport support

domdomegg:
SSE transport has been removed from the MCP specification in favor of streamable HTTP. This removes all SSE-specific transport code.

## Changes

- Remove `transport-sse-client` and `transport-sse-server` features
- Remove `SseClientTransport` and `SseServer` types
- Remove SSE-specific examples (`counter_sse`, `counter_sse_directly`)
- Migrate auth examples from SSE to streamable HTTP
- Update tests to remove SSE transport usage
- Update documentation

## Breaking changes

The following have been removed:
- `transport-sse-client` feature
- `transport-sse-client-reqwest` feature  
- `transport-sse-server` feature
- `SseClientTransport` type
- `SseServer` type
- `sse_client` and `sse_server` modules

Users should migrate to streamable HTTP transport. See `StreamableHttpClientTransport` and `StreamableHttpService` for the replacement APIs.

Ref: https://github.com/modelcontextprotocol/rust-sdk/pull/561#issuecomment-3576551699


jokemanfire:
I think removing SSE now is too early for SDK.

seuros:
SSE is super unstable ... There is no reason to promote it .

It not a bug, it just the wrong architecture.

SSE needs sticky sessions and it very hostile load balancers.  HTTP streamable is preferred way to connect remotely.

jokemanfire:
> SSE is super unstable ... There is no reason to promote it .
> 
> It not a bug, it just the wrong architecture.
> 
> SSE needs sticky sessions and it very hostile load balancers. HTTP streamable is preferred way to connect remotely.

This is not promoting it, but rather it existed in spec 2024-11-05 before, and we cannot start deleting it directly. Marking it as obsolete and deleting it in a later version would be more suitable for code evolution. As an SDK, we are not following the upper application layer, but rather adopting a gradual strategy based on specifications, especially for features that need to be abandoned.

seuros:
This is an SDK not a consumer API. 

You can drop and refactor as much you need to make the SDK cleaner and more secure. It won't force auto update on people repos.

This is a Rust context, that means if someone upgrade blindly, the compiler will not warn, it will refuse to obey. There is no (I shipped it, production blew because protocol was removed... javascript/ruby/python style)

They have 2 options: 

- Use new spec (will take them 2 minutes)
- Pin old version of the SDK forever.

---

Your thinking is not wrong, but this SDK case is an exception. The SSE spec was rushed and was afterthought, it never worked more than few minutes.

The STDIO in 2024 is still used in few abandoned repos, but SSE had no production implementation , zero serious user base.  

IDEs implemented remote MCP after HTTP streamable got released, and Claude code and other are supporting spec of June 2025+.

I will stand corrected if you show me 1 single serious app that still depends on 2024 SSE. 

--- 

The work is already done by @domdomegg. We just have to merge it.

alexhancock:
I reviewed and approved before seeing this conversation @jokemanfire @seuros

@jokemanfire I am interested to hear if you have concerns about simply removing it now

@domdomegg You may also be able to provide context on what you've seen other SDKs do, etc

jokemanfire:
I still prefer to mark n+1 as dropped and n+2 as deleted for the evolution of Rust code.But  @seuros  as mentioned, there's  zero serious user base, feel free to do this.

jokemanfire (approved):

alexhancock:
Yeah, I agree. Merging. Thanks @domdomegg

kzg001:
It is unreasonable for this SDK to eagerly remove support for a previously published specification. This forces projects that depend on this SDK to deal with additional compatibility issues. SDK evolution should strive to maintain backward compatibility. I cannot understand this decision. Other language SDKs for MCP maintain backward compatibility, yet the Rust version is being this aggressive.

seuros:
@cherishs001 Show a single serious project that used it and had a stable implementation.  The SSE is deprecated because it has a replacement streamable http.

The crate version was not yanked. 
You can pin forever the version prior the removal or fork the project and experience maintaining backward compatibility.

If you want SSE, copy all this code in the host app and use it there.

the protocol is transport agnostic, you can even use gprc or websocket if you wish.

alexhancock:
@cherishs001 Sorry you disagree with the decision to remove it but I agree with @seuros. The previous crate version that supported the former SSE Transport is still available, and many servers are now using the similar but better Streamable HTTP transport.

Is the Streamable HTTP transport suitable for a challenge that you have?

Issue #563, opened 2025-11-26
Handle JsonRpcError in initialize


wzxxing:
**Describe the bug**

When the MCP protocol initialization failed, the server can return a JSON RPC Error object, which is currently ignored by the rust SDK.

https://github.com/modelcontextprotocol/rust-sdk/blob/94428d543f9b97ff2d208800454131a8dc5ce1ea/crates/rmcp/src/service/client.rs#L128-L129

The Specification suggested that 

> Implementations SHOULD be prepared to handle these error cases:

https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#error-handling

**To Reproduce**

1. I found this when using the Kiro CLI which uses the rust-sdk. Plugging a MCP server that initialize failure. The error can only be found through `tracing`

```log
2025-11-26T11:05:21.291608Z  WARN rmcp::service::client: 129: Received unexpected message Error(JsonRpcError { jsonrpc: JsonRpcVersion2_0, id: Number(0), error: ErrorData { code: ErrorCode(-32001), message: "Your AWS session token has expired. Please refresh your credentials and try again. If you're using temporary credentials, you may need to re-authenticate to obtain new credentials.", data: None } })
```

`expect_response` ignores the mcp jsonrpc error, hence there was never a response received by the client.

https://github.com/modelcontextprotocol/rust-sdk/blob/94428d543f9b97ff2d208800454131a8dc5ce1ea/crates/rmcp/src/service/client.rs#L229-L235

Because my test-mcp server quits after the initialization error, I got a very generic "connection closed" message.

**Expected behavior**

My test server response to `initialization` with a synthesized JSON RPC error, I expect the JSON RPC error should be used as the `ClientInitializeError`.

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

PR #569, opened 2025-12-03
fix: use the json rpc error from the initialize response and bubble up to the client

tonyxwz:
## Motivation and Context

Returns the error when initialization response is a json rpc error. Otherwise the client would be waiting in the loop and hangs until timeout.

fixes: https://github.com/modelcontextprotocol/rust-sdk/issues/563

Before the change:
- If the server sent a JSON-RPC error during initialization, it fell through to the catch-all _ pattern
- The error was only logged with tracing::warn! and ignored
- The client would hang waiting for a valid response that never comes
- User gets a timeout or connection closed error instead of the actual error

After the change:
- JSON-RPC error is caught and converted to ClientInitializeError::JsonRpcError
- The error propagates up through the ? operator at line 241: .await?
- Returns immediately from serve_client_with_ct_inner with the error
- User receives Err(ClientInitializeError::JsonRpcError(error_data)) from .serve()
- User can see the actual error code and message from the server

## Breaking Changes

This is not a breaking change. But it should fix the issues where the client response with a default error message when the session connection failed.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -8,7 +8,7 @@ use crate::{
         ArgumentInfo, CallToolRequest, CallToolRequestParam, CallToolResult, CancelledNotification,
         CancelledNotificationParam, ClientInfo, ClientJsonRpcMessage, ClientNotification,
         ClientRequest, ClientResult, CompleteRequest, CompleteRequestParam, CompleteResult,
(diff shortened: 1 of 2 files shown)


alexhancock:
@tonyxwz Can you run `cargo fmt`? I should really add this to a precommit hook. Sorry!

tonyxwz:
Yes, will do.

alexhancock (approved):

PR #580, opened 2025-12-09
feat: add support for custom server notifications

bolinfest:
https://github.com/modelcontextprotocol/rust-sdk/pull/556 introduced support for
custom client notifications, so this PR makes the complementary change, adding
support for custom server notifications.

MCP clients, particularly ones that offer "experimental" capabilities,
may wish to handle custom server notifications that are not part of the
standard MCP specification. This change replaces `CustomClientNotification`
with a more general `CustomNotification`  type that can be used in both directions.

- `CustomNotification` can carry arbitrary methods/params while
  still preserving meta/extensions; wires it into the `ServerNotification` union
  and `serde` so `params` can be decoded with `params_as`
- allows client handlers to receive custom notifications via a new
  `on_custom_notification` hook
- adds integration coverage that sends a custom server notification end-to-end
  and asserts the client sees the method and payload

Test:

```shell
cargo test -p rmcp --features client test_custom_server_notification_reaches_client
```

### DIFF
--- a/crates/rmcp/src/handler/client.rs
+++ b/crates/rmcp/src/handler/client.rs
@@ -56,6 +56,9 @@ impl<H: ClientHandler> Service<RoleClient> for H {
             ServerNotification::PromptListChangedNotification(_notification_no_param) => {
                 self.on_prompt_list_changed(context).await
             }
+            ServerNotification::CustomNotification(notification) => {
+                self.on_custom_notification(notification, context).await
+            }
         };
         Ok(())
     }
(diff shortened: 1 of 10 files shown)


bolinfest:
@alexhancock hopefully this is relatively straightforward to review since it mirrors https://github.com/modelcontextprotocol/rust-sdk/pull/556

alexhancock (commented):
LGTM in terms of approach!

But can we unify the `CustomClientNotification` and `CustomServerNotification` types into a single `CustomNotification` type that can be sent either direction? Would cut down on a lot of the duplicated logic we need to implement for each type

bolinfest:
@alexhancock ah, so with the current implementation, we have these two things:

```rust
ts_union!(
    export type ClientNotification =
    | CancelledNotification
    | ProgressNotification
    | InitializedNotification
    | RootsListChangedNotification
    | CustomClientNotification;
);
```

```rust
ts_union!(
    export type ServerNotification =
    | CancelledNotification
    | ProgressNotification
    | LoggingMessageNotification
    | ResourceUpdatedNotification
    | ResourceListChangedNotification
    | ToolListChangedNotification
    | PromptListChangedNotification
    | CustomServerNotification;
);
```

I didn't realize it was possible to share types across the two (like `CancelledNotification`): will consolidate!

bolinfest:
@alexhancock please re-review now that:

- I consolidated things into one `CustomNotification` type
- Updated the PR body to reflect this change
- Tightened up some of the tests

PR #609, opened 2026-01-07
feat: provide blanket implementations for ClientHandler and ServerHandler traits

tninesling:
Adds blanket implementations for `Box<T>` and `Arc<T>` when `T` implements `ClientHandler` or `ServerHandler`.

## Motivation and Context

When building an MCP service, we may want to wrap it in a smart pointer before passing it into the tower service constructor. In my case, I want to spawn a background thread to do some work using some session state as requests are processed. However, introducing the `Arc` results in a compilation error:

```rust
use std::sync::Arc;
use rmcp::ServerHandler;
use rmcp::transport::StreamableHttpService;

struct MyService;

impl ServerHandler for MyService {}

fn start_server() {
    let my_service = Arc::new(MyService);

    // This gives "the trait bound `Arc<MyService>: ServerHandler` is not satisfied"
    let http_service = StreamableHttpService::new(
        move || Ok(my_service), 
        Default::default(),
        Default::default(),
    );
    // ...
}
```

## How Has This Been Tested?

Added integration tests and confirmed this fixes the compilation error in my project.

## Breaking Changes

I believe adding blanket implementations of this kind is typically considered a breaking change, although it should be a no-op for most users.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

## Additional context

I omitted blanket implementations for `&T` and `&mut T` due to the `'static` bound on the traits, and I omitted `Rc<T>` due to the `Sync` bound.

### DIFF
--- a/crates/rmcp/src/handler/client.rs
+++ b/crates/rmcp/src/handler/client.rs
@@ -1,4 +1,6 @@
 pub mod progress;
+use std::sync::Arc;
+
 use crate::{
     error::ErrorData as McpError,
     model::*,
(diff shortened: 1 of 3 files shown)


alexhancock:
@tninesling Nice! A good idea

I had a play with macros to see if it could dedupe some of the volume

Let me know what you think of https://github.com/modelcontextprotocol/rust-sdk/commit/0b54219c6bd7fb2b90fed14b2bf4c7d8b28b57c9

It's somewhat vibecoded so let me know if it does the job and if you like it

tninesling:
> Let me know what you think of [0b54219](https://github.com/modelcontextprotocol/rust-sdk/commit/0b54219c6bd7fb2b90fed14b2bf4c7d8b28b57c9)

Thanks for taking a look @alexhancock. As long as those tests pass, it fixes the issue I ran into. Although, the macros in that commit add a lot of indirection that make it hard to read. I agree with the direction of deduplicating the implementations though. I added some simpler macros in https://github.com/modelcontextprotocol/rust-sdk/pull/609/commits/a493962695b7683b3a906378c16bf324fccc3bed, which are slightly more verbose, but hopefully easier to read and maintain.

alexhancock:
@tninesling Can you run the formatter: `cargo fmt` and then also make the commit messages only have valid https://www.conventionalcommits.org? Then we can merge

alexhancock (approved):

Issue #610, opened 2026-01-08
OneshotTransport hangs due to race condition in stateless mode


madzarm:
**Describe the bug**
OneshotTransport can hang forever due to a race condition between send() and receive(). The issue is that tokio::sync::Notify::notify_waiters() only wakes tasks that are currently waiting. If notify_waiters() is called before receive() starts waiting (via notified().await), the notification is lost and receive() blocks indefinitely. 
This happens because send() returns a 'static future that runs independently, while receive() is called through tokio::select! which cancels and recreates futures between iterations. If the timing lines up wrong, you get a deadlock.

In our testing, this only occurred when the tool returned an error (e.g., 404 from downstream API). Successful responses never triggered the hang. This might be due to different code paths or timing characteristics when is_error: true.

**To Reproduce**
1. Configure stateless HTTP mode:
  let session_manager = Arc::new(LocalSessionManager::default());
  let config = StreamableHttpServerConfig {
      stateful_mode: false,
      ..Default::default()
  };

  2. Deploy to Lambda or similar serverless environment
  3. Send requests that result in tool errors (e.g., 404 from downstream)
  4. Send concurrent requests, especially during cold starts
  5. Observe some requests hanging for the full timeout duration (30s in Lambda)

  The race window is small but hits reliably under load. In our testing, ~3-5% of cold start requests would hang, always on error responses.

**Expected behavior**
After send() completes with a Response or Error message, the next call to receive() should return None promptly, allowing the serve loop to exit cleanly.

**Logs**
Successful request (no race):
  Looking up order: TEST
  serve finished quit_reason=Closed
  REPORT Duration: 841.46 ms

  Failed request (race condition hit - note: downstream returned 404):
  Looking up order: TEST
  Downstream API error (status 404)
  Response(JsonRpcResponse { ... is_error: Some(true) ... })
  REPORT Duration: 30000.00 ms  Status: timeout

  Note: no serve finished log - the serve loop never exited because receive() hung waiting for a notification that already fired.

**Additional context**
To make it work locally, i replaced Notify with Semaphore. Permits persist until acquired, so there's no race:

  // In struct
  termination: Arc<Semaphore>,

  // In new()
  termination: Arc::new(Semaphore::new(0)),

  // In send()
  if terminate {
      termination.add_permits(1);  // Persists even if no one waiting
  }

  // In receive()  
  if let Some(msg) = self.message.take() {
      return Some(msg);
  }
[...]

PR #611, opened 2026-01-09
fix: use Semaphore instead of Notify in OneshotTransport to prevent race condition during stateless mode

madzarm:
## Motivation and Context

  Fixes issue [#610](https://github.com/modelcontextprotocol/rust-sdk/issues/610)

 `OneshotTransport` uses `tokio::sync::Notify` to signal when the transport should close. However, `Notify::notify_waiters()` only wakes futures that are **currently awaiting** - if the notification fires before `receive()` starts waiting, the signal is lost forever.

  This creates a race condition in `serve_inner()` where `tokio::select!` can drop the `receive()` future between iterations:

  1. receive() returns the initial message
  2. send() completes with Response/Error, calls notify_waiters()
  3. select! picks join_next() branch, drops the receive() future
  4. Next iteration: receive() creates NEW notified() future
  5. This future waits forever - the notification already fired
  6. Serve loop hangs, stream never closes

  In stateless HTTP mode (Lambda, Cloud Functions), this causes requests to hang until timeout even though the tool executed successfully.

  ## Solution

  Replace `Notify` with `Semaphore`. Unlike notifications, **permits persist until acquired**:

  - `add_permits(1)` stores the permit even if no one is waiting
  - `acquire().await` returns immediately if a permit exists, or waits if not
  - Dropping an `acquire()` future doesn't consume the permit (cancel-safe)

  From [tokio docs](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#cancel-safety):
  > "This method is cancel safe. If acquire is used as the event in a tokio::select! statement and some other branch completes first, then no permits are acquired."

  ## How Has This Been Tested?

  - Deployed to AWS Lambda with API Gateway in stateless mode
  - Load tested with concurrent requests during cold starts
  - Specifically tested error paths (downstream 404s) which previously triggered the race ~3-5% of the time
  - After fix: zero hangs across thousands of requests

  ## Breaking Changes

  None. This is an internal implementation change. The `OneshotTransport` public API is unchanged.

  ## Types of changes

  - [ ] Documentation update

  ## Checklist

  ## Additional context

  The race window is small but hits reliably under load. In our production environment, ~3-5% of cold start requests would hang, always on error responses (possibly due to faster code paths when `is_error: true`).

  The fix is minimal - only changes the synchronization primitive from `Notify` to `Semaphore` with equivalent semantics but proper cancellation safety.

### DIFF
--- a/crates/rmcp/src/transport.rs
+++ b/crates/rmcp/src/transport.rs
@@ -173,7 +173,7 @@ where
 {
     message: Option<RxJsonRpcMessage<R>>,
     sender: tokio::sync::mpsc::Sender<TxJsonRpcMessage<R>>,
-    finished_signal: Arc<tokio::sync::Notify>,
+    termination: Arc<tokio::sync::Semaphore>,
 }
 
 impl<R> OneshotTransport<R>
@@ -188,7 +188,7 @@ where
             Self {
                 message: Some(message),
                 sender,
-                finished_signal: Arc::new(tokio::sync::Notify::new()),
+                termination: Arc::new(tokio::sync::Semaphore::new(0)),
             },
             receiver,
         )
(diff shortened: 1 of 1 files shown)


alexhancock (approved):
Nice catch and fix, thank you

madzarm:
Hi @alexhancock thanks for the quick response! I've shortened the commit message to fix the commit lint failure. I'd appreciate it if you could approve the workflow, thank you!

PR #618, opened 2026-01-14
fix: Don't treat non-success HTTP codes as transport errors

alexhancock:
Bringing back the remaining relevant change from https://github.com/modelcontextprotocol/rust-sdk/pull/486

cc @SteffenDE


alexhancock:
@jamadeo @jokemanfire Small one here redoing the still relevant parts of https://github.com/modelcontextprotocol/rust-sdk/pull/486

PR #619, opened 2026-01-14
chore: add pre-commit hook for conventional commit verification

alexhancock:
Automatically verify commits follow the https://www.conventionalcommits.org/en/v1.0.0/ format.

We get a lot of PRs with commits that do not follow the format, and it creates unnecessary cycles during code review. This PR installs a pre-commit hook on the first `cargo build` and then uses it when you commit.

Example:

```
13:09:41 ~/Development/rust-sdk alexhancock/pre-commit-lint* $ git commit -am "something invalid"
ERROR: Commit message does not follow conventional commit format.

Expected format: <type>(<scope>): <description>

Valid types: build, chore, ci, docs, feat, fix, perf, refactor, revert, style, test

Examples:
  feat: add new feature
  fix(parser): resolve parsing issue
  docs: update README

Your commit message:
  something invalid
```

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,0 +1,50 @@
+#!/bin/sh
+# Verify commit message follows conventional commit format
+# https://www.conventionalcommits.org/
+#
+# Uses npx commitlint if Node.js is available (same as CI),
+# otherwise falls back to basic shell regex validation.
+
+commit_msg_file="$1"
+commit_msg=$(cat "$commit_msg_file")
+
+# Skip merge commits
+if echo "$commit_msg" | grep -qE "^Merge "; then
+    exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


jamadeo (approved):
Nice, yes this will improve things.

Looks like CI calls `npx commitlint` -- should this do the same?

alexhancock:
@jamadeo I tried a version that does that but if we want to do local packages it would require adding `package.json`, assuming a `node` installation etc. It added a lot more boilerplate to the project.

There is maybe a version of that approach which just uses `npx` and doesn't do the linting if it's not available. I'll spike on that as it would be good to have the consistency.

alexhancock:
@jamadeo Came up with a better version which uses commitlint if node is available, and falls back to a more basic implementation if not.

PR #622, opened 2026-01-16
Add native-tls feature for reqwest dependency

anmatako:
This new feature avoids bringing the ring dependency that comes through even when using the reqwest-tls-no-provider feature

## Motivation and Context
This change adds a new feature that entirely avoids the transitive `ring` dependency. Ring is not allowed in Microsoft's internal repos and use of rmcp is flagged by compliance.

## How Has This Been Tested?
Tested in Codex agent from https://github.com/openai/codex

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -86,6 +86,8 @@ __reqwest = ["dep:reqwest"]
 
 reqwest = ["__reqwest", "reqwest?/rustls-tls"]
 
+reqwest-native-tls = ["__reqwest", "reqwest?/native-tls"]
+
 reqwest-tls-no-provider = ["__reqwest", "reqwest?/rustls-tls-no-provider"]
 
 server-side-http = [


alexhancock:
@anmatako This will need the commit message edited to conform to the https://www.conventionalcommits.org/en/v1.0.0/ format

I put in a pre-commit hook the other day to check this locally on commit, but perhaps you branched off from before? 

https://github.com/modelcontextprotocol/rust-sdk/pull/619

Sorry for the hassle!

anmatako:
Moving to v0.13 avoids pulling ring. Native TLS may still be needed as a feature, but not blocking at the moment

anmatako:
Alex,

I actually closed the PR since the latest version allows to avoid "ring"
without this native-tls feature. However, having the native-tls provider
may still be useful so if you think that's valid I can reinstate the PR and
amend the commit message.
--
Antonis Matakos

alexhancock:
Ok sounds good. Let's leave it at that with the solution in 0.13 then. Thanks!

PR #627, opened 2026-01-23
fix(tasks): #626 model task capabilities correctly

### LINKED ISSUE #626: in 0.13.0, TasksCapability structure has incorrect types
Comparing

https://github.com/modelcontextprotocol/rust-sdk/blob/e623f2acabd53e51a978d160f955c315bc16c220/crates/rmcp/src/model/capabilities.rs#L47

with

https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-11-25/schema.ts#L433

the fields should be maps, not booleans

alexhancock:
Fixes #626 

Capabilities for tasks take the following form.

**Client**
```
{
  "capabilities": {
    "tasks": {
      "list": {},
      "cancel": {},
      "requests": {
        "sampling": {
          "createMessage": {}
        },
        "elicitation": {
          "create": {}
        }
      }
    }
  }
}
```

**Server**
```
{
  "capabilities": {
    "tasks": {
      "list": {},
      "cancel": {},
      "requests": {
        "tools": {
          "call": {}
        }
      }
    }
  }
}
```

We were previously using booleans to model these instead of objects, causing errors during initialization between servers and clients that expressed these capabilities

### DIFF
--- a/crates/rmcp/src/model/capabilities.rs
+++ b/crates/rmcp/src/model/capabilities.rs
@@ -40,24 +40,121 @@ pub struct RootsCapabilities {
     pub list_changed: Option<bool>,
 }
 
-/// Task capability negotiation for SEP-1686.
+/// Task capabilities shared by client and server.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
 #[serde(rename_all = "camelCase")]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 pub struct TasksCapability {
-    /// Map of request category (e.g. "tools.call") to a boolean indicating support.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub requests: Option<TaskRequestMap>,
-    /// Whether the receiver supports `tasks/list`.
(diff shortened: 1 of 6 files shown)


crates/rmcp/src/model/capabilities.rs
```diff
@@ -368,4 +465,61 @@ mod test {
...
+        // Verify structure
+        assert!(tasks.supports_list());
+        assert!(tasks.supports_cancel());
+        assert!(tasks.supports_sampling_create_message());
+        assert!(tasks.supports_elicitation_create());
+        assert!(!tasks.supports_tools_call());
+        
+        // Verify serialization matches expected format
+        let json = serde_json::to_value(&tasks).unwrap();
+        assert_eq!(json["list"], serde_json::json!({}));
```
jamadeo, line 502: are these really testing all that much?

alexhancock: not terribly. the serialization one has value but these I agree don't add much. merging for not to not wait for another CI cycle but I will remove them on a future change.

PR #639, opened 2026-02-02
fix(tasks): avoid dropping completed task results during collection

### LINKED ISSUE #638: Completed tasks are erased from `OperationProcessor`
**Describe the bug**
The completed tasks are erased from `OperationProcessor` 

**To Reproduce**
1. `tasks/get` request returns `task: None` after task completion.
2. `tasks/result` request returns error after task completion.

**Expected behavior**
1. `tasks/get` request should return a `Task` object with `status: "completed"`.
2. `tasks/result` request should return the task result.

**Additional context**
I am using the latest commit (32a68aa) because the published version 0.14.0 does not include the enum-variant reordering fix.
```toml
rmcp = { git = "https://github.com/modelcontextprotocol/rust-sdk", features = ["macros", "client", "rmcp/transport-streamable-http-client-reqwest"] }
```

**Fix**

This is caused by `std::mem::take` from the `OperationProcessor::collect_completed_results`, which clears `OperationProcessor::completed_results`.

https://github.com/modelcontextprotocol/rust-sdk/blob/32a68aa2396aba6a9990d2b38164b2e74906e092/crates/rmcp/src/task_manager.rs#L197-L206

The solution would be simply to not return anything from this function. While this would be a breaking change, this method is only used in a single test code, and I doubt anyone was calling this method directly without using the `#[tool_handler]` macro.

coinmoles:
Remove the `std::mem::take` from `OperationProcessor::collect_completed_results`

## Motivation and Context
`std::mem::take` would clear `OperationProcessor::completed_results`, causing completed tasks to be dropped. This made `tasks/get` and `tasks/result` to fail for subsequent requests. 

## How Has This Been Tested?
- All existing tests passed
- Tested with a custom task server/client implementation (Will add this to examples if requested)

## Breaking Changes
This is technically a breaking change that changes the return type of `OperationProcessor::collect_completed_results` from `Vec<TaskResult>` to `()`. However, I doubt there was anyone calling the method directly without using the `#[tool_handler]` macro.

I did this to avoid introducing an unnecessary `clone` call. If this is a problem, I can change the method so it returns either:
- `self.completed_results.clone()`
-  An owned vector of newly collected items, which I think was the intended return value in the first place.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Fixes #638

### DIFF
--- a/crates/rmcp/src/task_manager.rs
+++ b/crates/rmcp/src/task_manager.rs
@@ -195,7 +195,7 @@ impl OperationProcessor {
     }
 
     /// Collect completed results from running tasks and remove them from the running tasks map.
-    pub fn collect_completed_results(&mut self) {
+    fn collect_completed_results(&mut self) {
         while let Ok(result) = self.task_result_receiver.try_recv() {
             self.running_tasks.remove(&result.descriptor.operation_id);
             self.completed_results.push(result);
@@ -204,6 +204,7 @@ impl OperationProcessor {
 
     /// Check for tasks that have exceeded their timeout and handle them appropriately.
     pub fn check_timeouts(&mut self) {
+        self.collect_completed_results();
         let now = std::time::Instant::now();
         let mut timed_out_tasks = Vec::new();
 
@@ -228,7 +229,8 @@ impl OperationProcessor {
     }
 
     /// Get the number of running tasks.
-    pub fn running_task_count(&self) -> usize {
+    pub fn running_task_count(&mut self) -> usize {
+        self.collect_completed_results();
         self.running_tasks.len()
     }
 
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/task_manager.rs
```diff
@@ -195,14 +195,13 @@ impl OperationProcessor {
     }
 
     /// Collect completed results from running tasks and remove them from the running tasks map.
-    pub fn collect_completed_results(&mut self) -> Vec<TaskResult> {
+    pub fn collect_completed_results(&mut self) {
```
alexhancock: thoughts on this being `pub` in the first place?

coinmoles: Right now, the function has to be public because it is called from the output of  `#[task_handler]` macro.

However, it would make sense to make this private and have `OperationProcessor::peek_completed` and `OperationProcessor::take_completed_result` call this before returning. Do you think that would be better?

alexhancock: Yes, something like that makes sense to me. Can you make the change?

coinmoles: Pushed the change.

One thing that is slightly awkward is that `peek_completed` (and a few other functions whose names suggest non-mutating behavior) now requires `&mut self`. I think it is fine, but it could be slightly confusing.

Also, while I was doing that I realized that `task_result_receiver` has no real reason to be an `Option` (The channel is initialized on startup and not set again) and made it required.

alexhancock (approved):

Issue #632, opened 2026-01-28
AthorizationManager discover_metadata stopped working for paths with suffix in 0.13 release


glicht:
**Describe the bug**
After upgrading to 0.13 (also tested with 0.14) metadata discovery stopped working for urls with a suffix. 

Example url: `https://auditlogs.mcp.cloudflare.com/mcp`

It seems that the discovery stopped checking `<base_domain>/.well-known/oauth-authorization-server`.

**To Reproduce**

Use following example program:
```rust
use anyhow::{Context, Result};
use rmcp::transport::AuthorizationManager;
use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};

#[tokio::main]
async fn main() -> Result<()> {
    tracing_subscriber::registry()
        .with(tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| "debug".into()))
        .with(tracing_subscriber::fmt::layer())
        .init();

    let url = std::env::args().nth(1).context("Usage: metadata_discovery <url>")?;
    tracing::info!("Discovering metadata for URL: {}", url);

    let metadata = AuthorizationManager::new(&url).await?.discover_metadata().await?;
    tracing::info!("Discovered metadata: {:#?}", metadata);

    Ok(())
}
```

Run with the following url for example: `https://auditlogs.mcp.cloudflare.com/mcp`

Output with rmcp 0.14 (fails):
```
2026-01-28T08:55:19.269795Z  INFO metadata_discovery: Discovering metadata for URL: https://auditlogs.mcp.cloudflare.com/mcp
2026-01-28T08:55:19.290356Z DEBUG rmcp::transport::auth: discovery url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("auditlogs.mcp.cloudflare.com")), port: None, path: "/.well-known/oauth-authorization-server/mcp", query: None, fragment: None }
2026-01-28T08:55:19.290959Z DEBUG reqwest::connect: starting new connection: https://auditlogs.mcp.cloudflare.com/
2026-01-28T08:55:19.293894Z DEBUG hyper_util::client::legacy::connect::http: connecting to 104.18.24.159:443
2026-01-28T08:55:19.307814Z DEBUG hyper_util::client::legacy::connect::http: connected to 104.18.24.159:443
[...]

PR #641, opened 2026-02-02
fix(auth): oauth metadata discovery

glicht:
Auth metadata discovery stopped checking <base_domain>/.well-known/oauth-authorization-server with the 0.13 release. Adding a fix.

## Motivation and Context

See #632 

## How Has This Been Tested?

Added unit test that failed before the fix and passes after the fix

## Breaking Changes

No
## Types of changes

- [ ] Documentation update

## Checklist

- [ x I have added or updated documentation as needed

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1804,11 +1804,12 @@ mod tests {
     fn test_discovery_urls_with_path_suffix() {
         // When the base URL has a path suffix (e.g., /mcp), the discovery should
         // eventually fall back to checking /.well-known/oauth-authorization-server
-        // at the root, not just /.well-known/oauth-authorization-server/{path}.        
+        // at the root, not just /.well-known/oauth-authorization-server/{path}.
         let base_url = Url::parse("https://mcp.example.com/mcp").unwrap();
         let urls = AuthorizationManager::generate_discovery_urls(&base_url);
 
-        let canonical_oauth_fallback = "https://mcp.example.com/.well-known/oauth-authorization-server";
+        let canonical_oauth_fallback =
+            "https://mcp.example.com/.well-known/oauth-authorization-server";
 
         assert!(
             urls.iter().any(|u| u.as_str() == canonical_oauth_fallback),


wdawson:
This was due to an [intentional correction](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/797) in the MCP [spec version 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery)

glicht:
> This was due to an [intentional correction](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/797) in the MCP [spec version 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery)

@wdawson see #632 

The change that was implemented was a breaking change. I don't think that was the purpose. The sdk stopped working with common mcp servers (such as cloudflare). 

Also from what I checked in the typescript-sdk it supports fetching from: `<base_domain>/.well-known/oauth-authorization-server` as a fallback.  So, I think we should maintain this support for backwards compatibility.

wdawson:
> > This was due to an [intentional correction](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/797) in the MCP [spec version 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery)
> 
> @wdawson see #632 
> 
> The change that was implemented was a breaking change. I don't think that was the purpose. The sdk stopped working with common mcp servers (such as cloudflare). 
> 
> Also from what I checked in the typescript-sdk it supports fetching from: `<base_domain>/.well-known/oauth-authorization-server` as a fallback.  So, I think we should maintain this support for backwards compatibility.

@glicht I think it's ok. Just wanted to point out why it was done. I think the bigger problem is that this SDK seems to not use the protected resource metadata location and tries to discover something first. Fixing that bug might fix this issue with Cloudflare

tanish111 (commented):
@glicht tested locally?

tanish111:
@alexhancock can you also look into it issue and fix once?

glicht:
> @glicht tested locally?

Yes. I've tested locally and the fix works as expected. Thanks

alexhancock:
@wdawson @tanish111 @glicht re:

> I think the bigger problem is that this SDK seems to not use the protected resource metadata location and tries to discover something first. Fixing that bug might fix this issue with Cloudflare

I feel like we've made a patchwork set of fixes/adjustments/tweaks over time. I get so confused by the ordering needed! 

Is anyone well positioned to take a holistic look at how the Rust SDK gets/uses resource metadata?

alexhancock:
@glicht for this one specifically, looks like it needs the code formatter run. `cargo fmt` should do it

glicht:
> @glicht for this one specifically, looks like it needs the code formatter run. `cargo fmt` should do it

Thanks. I've pushed a format fix.

wdawson:
> @wdawson @tanish111 @glicht re:
> 
> > I think the bigger problem is that this SDK seems to not use the protected resource metadata location and tries to discover something first. Fixing that bug might fix this issue with Cloudflare
> 
> I feel like we've made a patchwork set of fixes/adjustments/tweaks over time. I get so confused by the ordering needed! 
> 
> Is anyone well positioned to take a holistic look at how the Rust SDK gets/uses resource metadata?

@alexhancock I'm happy to take a pass. I know there are several issues and PRs in flight and I don't know the history. But I was a contributor to some of the authorization changes (hopefully the less confusing ones 😅) and should be able to get to a "correct" implementation if that would be a good starting point. Then would love help testing that against all the things people need/want to make sure there aren't regressions like @glicht mentioned. Let me know if that approach is useful. Alternatively I can review or advise another implementor. Happy to chat on discord or whatever.

alexhancock:
Yes @wdawson let's do a comprehensive change that addresses all edge cases in this space and makes the implementation of metadata discovery & handling correct.

Then when we have a branch where we believe everything functions 100% to spec, I can make a test build of [goose](https://github.com/block/goose) (the other project I work on) which uses the branch and we can test with various servers before a merge

Thanks for being willing to help!

PR #660, opened 2026-02-14
fix(streamable-http): prevent SSE reconnect loops with shadow channels

its-mash:
## Summary

Fixes infinite SSE reconnect loops caused by multiple competing EventSource connections in the Streamable HTTP server session layer. Tested with Cursor, VS Code, and Claude Desktop.

### Problem

When the server sends POST SSE responses with a `retry` field (priming event), the browser's EventSource API automatically reconnects via GET after the stream ends. This creates **multiple competing EventSource connections** — one from the initial standalone GET, and additional ones from completed POST responses (initialize, tools/list, etc.).

Each reconnecting GET called `resume()` which **replaced** `self.common.tx` (the common channel sender). Dropping the old sender closed the old receiver, terminating the OTHER EventSource's stream. Both reconnected every `sse_retry` seconds (3s), leapfrogging each other in an infinite loop.

### Root Cause Analysis

The connection lifecycle that causes the loop:

```
1. Client sends POST initialize → SSE response with priming (retry: 3000) → stream ends
2. Client sends GET (standalone) → becomes primary common channel
3. POST EventSource reconnects via GET (3s later) → replaces common.tx → kills stream from step 2
4. GET from step 2's EventSource reconnects → replaces common.tx → kills stream from step 3
5. Repeat every 3 seconds indefinitely
```

### Solution — Shadow Channels

Instead of always replacing the common channel sender on resume, the new `resume_or_shadow_common()` method checks whether the primary common channel is still active:

- **Primary dead** (`tx.is_closed()`) → Replace it. The new stream becomes the primary notification channel.
- **Primary alive** → Create a **shadow stream** — an idle SSE connection kept alive by SSE keep-alive pings (`:` comments) that does NOT receive notifications and does NOT replace the primary channel.

Shadow streams prevent competing EventSource connections from killing each other while keeping the HTTP connections alive (so EventSource doesn't reconnect again).

### Changes

| Commit | Description |
|--------|-------------|
| `8bd424e` | Initial 409 Conflict approach (returned error on duplicate standalone stream) |
| `0d03eb5` | Handle resume with completed request-wise channels (fall through to common) |
| `a7bb822` | Remove 409 Conflict — allow channel replacement per MCP spec ("client MAY remain connected to multiple SSE streams") |
| `7cf5406` | Skip cache replay (`sync`) when replacing active streams (replaying list_changed notifications caused notification loops) |
| `a7df58c` | **Shadow channels** — the final fix that prevents the leapfrog loop entirely |

### Key Design Decisions

1. **No cache replay on common channel resume**: Server-initiated notifications (tools/list_changed, resources/list_changed) are idempotent signals. Replaying cached ones causes clients to re-process old events, triggering unnecessary re-fetches. Missing one is harmless — the next real event will arrive naturally.

2. **Shadow streams are idle**: They don't receive notifications — only the primary common channel does. This is intentional: the reconnecting POST EventSources don't need notifications (their purpose — delivering the POST response — is already fulfilled). They just need to stay alive so EventSource stops reconnecting.

3. **Automatic cleanup**: Closed shadow senders (from client disconnection) are cleaned up via `retain(!is_closed())` on each `resume()` call and `clear()` on `close_sse_stream()`.

4. **Request-wise channels still sync**: Only the common channel skips replay. Request-wise channels (POST response streams) still use `sync()` for proper resumption, since those carry actual request/response data that may need replay.

### Files Changed

- `crates/rmcp/src/transport/streamable_http_server/session/local.rs`
  - Added `shadow_txs: Vec<Sender<ServerSseMessage>>` to `LocalSessionWorker`
  - New method `resume_or_shadow_common()` with primary-alive check
  - Updated `resume()` to use shadow logic for both direct common and request-wise fallback paths
  - Updated `close_sse_stream()` to clear shadow senders
  - Updated `create_local_session()` to initialize `shadow_txs`

## Test Plan

- [x] Cursor connects and initializes successfully (no 409/500 errors)
- [x] Cursor does NOT enter infinite GET reconnect loop after connection
- [x] Feature changes trigger exactly one batch of list_changed notifications
- [x] Cursor receives and processes notifications correctly (re-fetches tools/resources)
- [x] No notification replay loop (no repeated ResourceListChanged every 3s)
- [x] VS Code connects and works correctly (unaffected by changes)
- [x] `cargo check --workspace` passes

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -567,8 +567,14 @@ impl LocalSessionWorker {
         &mut self,
         last_event_index: usize,
     ) -> Result<StreamableHttpMessageReceiver, SessionError> {
-        let (tx, rx) = tokio::sync::mpsc::channel(self.session_config.channel_capacity);
-        if self.common.tx.is_closed() {
+        let is_replacing_dead_primary = self.common.tx.is_closed();
+        let capacity = if is_replacing_dead_primary {
+            self.session_config.channel_capacity
+        } else {
+            1 // Shadow streams only need keep-alive pings
+        };
+        let (tx, rx) = tokio::sync::mpsc::channel(capacity);
+        if is_replacing_dead_primary {
             // Primary common channel is dead — replace it.
             tracing::debug!("Replacing dead common channel with new primary");
             self.common.tx = tx;
@@ -580,6 +586,15 @@ impl LocalSessionWorker {
             // that stays alive via SSE keep-alive but doesn't receive
             // notifications. This prevents competing EventSource connections
             // from killing each other's channels.
+            const MAX_SHADOW_STREAMS: usize = 32;
+
+            if self.shadow_txs.len() >= MAX_SHADOW_STREAMS {
+                tracing::warn!(
+                    shadow_count = self.shadow_txs.len(),
+                    "Shadow stream limit reached, dropping oldest"
+                );
+                self.shadow_txs.remove(0);
+            }
             tracing::debug!(
                 shadow_count = self.shadow_txs.len(),
                 "Common channel active, creating shadow stream"
--- a/crates/rmcp/tests/test_sse_concurrent_streams.rs
+++ b/crates/rmcp/tests/test_sse_concurrent_streams.rs
@@ -743,3 +743,42 @@ async fn dead_primary_replacement_replays_cached_events() {
 
     ct.cancel();
 }
+
+// ─── Tests: Shadow stream limits ─────────────────────────────────────────────
+
+/// Opening more than 32 shadow streams should not crash or reject — the server
+/// drops the oldest shadow to stay within the limit. Primary still works.
+#[tokio::test]
+async fn shadow_stream_limit_drops_oldest() {
+    let ct = CancellationToken::new();
+    let trigger = Arc::new(Notify::new());
+    let url = start_test_server(ct.clone(), trigger.clone()).await;
+    let client = reqwest::Client::new();
+
+    let session_id = initialize_session(&client, &url).await;
+    send_initialized_notification(&client, &url, &session_id).await;
+    tokio::time::sleep(Duration::from_millis(200)).await;
+
+    // First GET — primary
+    let get1 = open_standalone_get(&client, &url, &session_id).await;
+    assert_eq!(get1.status(), 200);
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
+    // Open 35 shadow streams (exceeds MAX_SHADOW_STREAMS=32)
+    let mut shadows = Vec::new();
+    for i in 0..35 {
+        let shadow = open_standalone_get(&client, &url, &session_id).await;
+        assert_eq!(shadow.status(), 200, "Shadow #{i} should succeed");
+        shadows.push(shadow);
+    }
+
+    // Primary should still receive notifications despite shadow churn
+    trigger.notify_one();
+
+    assert!(
+        wait_for_sse_event(get1, "tools/list_changed", Duration::from_secs(3)).await,
+        "Primary should still work after exceeding shadow limit"
+    );
+
+    ct.cancel();
+}


crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -315,6 +321,8 @@ pub enum SessionError {
     SessionServiceTerminated,
     #[error("Invalid event id")]
     InvalidEventId,
+    #[error("Conflict: Only one standalone SSE stream is allowed per session")]
+    Conflict,
```
DaleSeo: Where is this used?

crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -513,36 +521,69 @@ impl LocalSessionWorker {
...
+                    // The client's EventSource is reconnecting after the POST SSE stream
+                    // ended. Fall through to common channel handling below.
+                    tracing::debug!(
+                        http_request_id,
+                        "Request-wise channel completed, falling back to common channel"
+                    );
+                    self.resume_or_shadow_common()
+                }
             }
+            None => self.resume_or_shadow_common(),
```
DaleSeo: The old `resume()` code replayed cached messages starting from the client's `Last-Event-ID`, but `resume_or_shadow_common()` completely drops `last_event_id.index`. Is that intentional?

crates/rmcp/Cargo.toml
```diff
@@ -241,3 +241,8 @@ required-features = [
   "transport-streamable-http-server",
 ]
 path = "tests/test_custom_headers.rs"
+
+[[test]]
+name = "test_sse_channel_replacement_bug"
```
DaleSeo: I suggest describing what the tests verify rather than what triggered them.

```suggestion
name = "test_sse_concurrent_streams"
```

alexhancock: Yes I agree with this

its-mash: Done in 4fd9ae7 — renamed both the Cargo.toml entry and the file.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
its-mash: You're right, dropping last_event_id.index entirely was an oversight. The common channel also carries server-initiated requests (e.g. sampling/createMessage, roots/list), not just idempotent list_changed signals — losing those on reconnect would be a real problem.

Fixed in 4fd9ae7: resume_or_shadow_common() now takes last_event_index and calls sync() when replacing a dead primary. Shadows still skip replay since they're idle by design. Added a test (dead_primary_replacement_replays_cached_events) to verify cached events are replayed on primary replacement.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
its-mash: Good catch — this was a leftover from the initial 409 Conflict approach (commit 8bd424e) that's no longer used after switching to shadow channels. Removed it along with the dead string-matching in tower.rs in 4fd9ae7.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -513,36 +519,77 @@ impl LocalSessionWorker {
...
+    ///
+    /// When the primary is still active, create a "shadow" stream — an idle SSE
+    /// connection kept alive by keep-alive pings. This prevents multiple
+    /// EventSource connections (e.g. from POST response reconnections) from
+    /// killing each other by repeatedly replacing the common channel sender.
+    async fn resume_or_shadow_common(
+        &mut self,
+        last_event_index: usize,
+    ) -> Result<StreamableHttpMessageReceiver, SessionError> {
+        let (tx, rx) = tokio::sync::mpsc::channel(self.session_config.channel_capacity);
```
DaleSeo: Shadow streams are idle by design. They don’t receive notifications, just SSE keep-alive pings. However, they still get the full `channel_capacity` here.

Since only the primary replacement path needs that full capacity, we might want to use a minimal buffer for shadows:

```suggestion
        let is_replacing_dead_primary = self.common.tx.is_closed();
        let capacity = if is_replacing_dead_primary {
            self.session_config.channel_capacity
        } else {
            1 // Shadow streams only need keep-alive pings
        };
        let (tx, rx) = tokio::sync::mpsc::channel(capacity);
```

crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -513,36 +519,77 @@ impl LocalSessionWorker {
...
+        } else {
+            // Primary common channel is still active. Create a shadow stream
+            // that stays alive via SSE keep-alive but doesn't receive
+            // notifications. This prevents competing EventSource connections
+            // from killing each other's channels.
+            tracing::debug!(
+                shadow_count = self.shadow_txs.len(),
+                "Common channel active, creating shadow stream"
+            );
+            self.shadow_txs.push(tx);
```
DaleSeo, line 602: The `shadow_txs: Vec<Sender<ServerSseMessage>>` vector can grow indefinitely. If a client misbehaves or has a bug, it could open hundreds of GET streams, each creating a new sender. The cleanup with `retain(!is_closed())` only happens at the beginning of `resume()`, so there's a period when all shadows are active at the same time. It would be helpful to add an upper limit.

```suggestion
            const MAX_SHADOW_STREAMS: usize = 32;

            if self.shadow_txs.len() >= MAX_SHADOW_STREAMS {
                tracing::warn!(
                    shadow_count = self.shadow_txs.len(),
                    "Shadow stream limit reached, dropping oldest"
                );
                self.shadow_txs.remove(0);
            }
            self.shadow_txs.push(tx);
```

its-mash: agree, updated

crates/rmcp/src/transport/streamable_http_server/session/local.rs
its-mash: agree, updated

alexhancock:
@its-mash Can you fix this minor lint issue

```
error: this import is redundant
  --> crates/rmcp/tests/test_sse_concurrent_streams.rs:18:1
   |
18 | use reqwest;
   | ^^^^^^^^^^^^ help: remove it entirely
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports
   = note: `-D clippy::single-component-path-imports` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::single_component_path_imports)]`
```

alexhancock:
> @its-mash Can you fix this minor lint issue
> 
> ```
> error: this import is redundant
>   --> crates/rmcp/tests/test_sse_concurrent_streams.rs:18:1
>    |
> 18 | use reqwest;
>    | ^^^^^^^^^^^^ help: remove it entirely
>    |
>    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports
>    = note: `-D clippy::single-component-path-imports` implied by `-D warnings`
>    = help: to override `-D warnings` add `#[allow(clippy::single_component_path_imports)]`
> ```

@its-mash I applied the required fix here and merged https://github.com/modelcontextprotocol/rust-sdk/pull/682

All your commits went in!

its-mash:
@alexhancock sorry didn't notice your last comment. Thanks for the help.

PR #664, opened 2026-02-16
fix(schema): remove AddNullable from draft2020_12 settings

### LINKED ISSUE #663: schema_for_type uses AddNullable which is non-compliant with JSON Schema 2020-12
## Problem

The `schema_for_type()` function in `crates/rmcp/src/handler/server/common.rs` uses `AddNullable::default()` as a transform when generating JSON Schema 2020-12 schemas:

```rust
let mut settings = SchemaSettings::draft2020_12();
settings.transforms = vec![Box::new(schemars::transform::AddNullable::default())];
```

However, the `nullable` keyword is an **OpenAPI 3.0 extension**, not part of JSON Schema 2020-12. This causes validation failures with strict JSON Schema validators.

## Background

PR #549 updated `schema_for_type()` to use `SchemaSettings::draft2020_12()` to align with the MCP spec update (modelcontextprotocol/modelcontextprotocol#655), but it retained the `AddNullable` transform from the previous Draft 7 implementation.

The `nullable` keyword originated from OpenAPI 3.0 and was never part of any JSON Schema draft. In JSON Schema 2020-12, nullable types should be represented using:
- `{"type": ["string", "null"]}` (type array with null)
- `{"anyOf": [{"type": "string"}, {"type": "null"}]}`

## Impact

- Strict JSON Schema 2020-12 validators (like mcporter) reject schemas containing `nullable`
- Generated tool schemas may fail validation in environments with strict schema checking
- The `` declaration claims 2020-12 compliance but includes non-2020-12 keywords

## Example

With the current implementation, an optional string field produces:
```json
{
  "nullable": true,
  "type": "string"
}
```

This should instead be:
```json
{
  "type": ["string", "null"]
}
```

## Suggested Fix

Remove or conditionally disable the `AddNullable` transform when using `SchemaSettings::draft2020_12()`:

```rust
let mut settings = SchemaSettings::draft2020_12();
// Do not add AddNullable - nullable is OpenAPI 3.0, not JSON Schema 2020-12
let generator = settings.into_generator();
```

## References

- PR #549: Switched to JSON Schema 2020-12 but retained AddNullable
- [JSON Schema 2020-12 Specification](https://json-schema.org/draft/2020-12/json-schema-core.html)
- [OpenAPI 3.0 nullable extension](https://swagger.io/docs/specification/data-models/data-types/#nullable)
- [schemars AddNullable documentation](https://docs.rs/schemars/latest/schemars/transform/struct.AddNullable.html)

mwotton:
## Problem

The `schema_for_type()` function uses `AddNullable::default()` as a transform when generating JSON Schema 2020-12 schemas. However, the `nullable` keyword is an **OpenAPI 3.0 extension**, not part of JSON Schema 2020-12. This causes validation failures with strict JSON Schema validators.

## Solution

Remove the `AddNullable` transform from the draft2020_12 settings. JSON Schema 2020-12 represents nullable types using:
- `{"type": ["string", "null"]}` (type array with null)
- `{"anyOf": [{"type": "string"}, {"type": "null"}]}`

## Before

An optional string field produced:
```json
{
  "nullable": true,
  "type": "string"
}
```

## After

An optional string field produces:
```json
{
  "type": ["string", "null"]
}
```

Fixes #663


alexhancock:
@mwotton Thanks for the fix. `test_complex_schema` needs a look

mwotton:
Added a second follow-up commit (`70011f9`) after the `test_complex_schema` fix because another test had the same stale nullable assumption.

`test_tool_macros::test_optional_field_schema_generation_via_macro` was still asserting OpenAPI-style `nullable: true` and explicitly rejecting `type: ["string", "null"]`, which conflicts with Draft 2020-12 output from `schema_for_type()`.

This was not visible in the original CI failure because `cargo test --all-features` failed fast on `test_complex_schema` first. The second commit aligns that macro test with Draft 2020-12 behavior.

DaleSeo:
Hey @mwotton, thanks for the fix! The code change looks good.

The CI failures aren't related to your changes. All five failing checks hit the same compilation error:

```
error[E0004]: non-exhaustive patterns: `Some(Protocol::Tlsv13)` not covered  
--> native-tls-0.2.17/src/imp/openssl.rs:61:22
```

This happened because your branch is based on an older main that still used reqwest 0.12, which depends on native-tls. Since then, PR #669 upgraded to reqwest 0.13, which uses rustls by default and doesn't pull in native-tls, so the main CI is green now.

To fix the CI, please rebase your branch onto the latest main. I think that should resolve all five failures. The code changes themselves look correct.

PR #670, opened 2026-02-18
Update SECURITY.md to use GitHub Security Advisories

localden:
Replaces Anthropic HackerOne VDP references with standard GitHub Security Advisory process.


alexhancock:
@localden FYI I rebased and amended the commit just to make it follow conventional commit std which we require in ci

PR #678, opened 2026-02-22
fix(auth): current_scopes read to async

glicht:
## Motivation and Context
I got a panick when performing token refresh and the token endpoint didn't return scopes. The cause is a call to `blocking_read()` which cannot be called from within a Tokio async runtime. 

Here is the message I received:
```
thread 'tokio-runtime-worker' (67286162) panicked at /Users/glicht/dev/binahm/rust-sdk/crates/rmcp/src/transport/auth.rs:1005:52:
Cannot block the current thread from within a runtime. This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

Bug seems to have been introduced recently here: https://github.com/modelcontextprotocol/rust-sdk/pull/651 
@wdawson FYI

## How Has This Been Tested?

Tested locally with a client which connects to an mcp server which doesn't return scopes. After performing the fix it works fine.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -999,10 +999,10 @@ impl AuthorizationManager {
             .await
             .map_err(|e| AuthError::TokenRefreshFailed(e.to_string()))?;
 
-        let granted_scopes: Vec<String> = token_result
-            .scopes()
-            .map(|scopes| scopes.iter().map(|s| s.to_string()).collect())
-            .unwrap_or_else(|| self.current_scopes.blocking_read().clone());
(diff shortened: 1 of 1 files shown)


michaelneale:
nice @glicht worth a unit test for that or too obscure?

glicht:
> nice @glicht worth a unit test for that or too obscure?

Thanks for the review.

I thought of adding a test but it requires using a mock server (such as wiremock) to reach this edge case. I didn't see any tests using mock servers, so I figured I'll leave it out.

PR #680, opened 2026-02-23
fix (auth): refresh token expiry

wdawson:
OAuth token expiration was not being handled correctly.

## Motivation and Context

Two bugs currently exist:

1. **Token expiry is never detected**: `expires_in()` from the oauth2 crate returns the original duration from the token response (e.g., 3600s), not the time remaining. So the check `if expires_in <= Duration::from_secs(0)` is always false
2. **Refresh failures don't trigger re-auth**: If `refresh_token()` fails (e.g., refresh token revoked), the error propagates as `TokenRefreshFailed` instead of `AuthorizationRequired`, so it's more difficult for the client to know when to re-prompt.

## How Has This Been Tested?

- Tested locally against an Arcade.dev MCP Gateway server
- Added unit tests for the new behavior

## Breaking Changes

Technically breaking, but should provide better behavior:

- The `AuthorizationRequired` error now propagates instead of `TokenRefreshFailed` to provide a clear, consistent contract for clients

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Also contains the fix from @glicht in #678 (thanks!). That should merge first for proper attribution for that fix.

AI Disclosure: AI assisted with implementation and test authoring. All changes were reviewed and guided by hand.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashMap, sync::Arc, time::Duration};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
 
 use async_trait::async_trait;
 use oauth2::{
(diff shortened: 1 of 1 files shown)


glicht:
I also saw this behavior with the expiry. I wasn't fully sure if it was by design or a bug. My way to get around this was to implement at the credential store's load function to decrement the expire_in to reflect the actual number of seconds left for expiration (or zero once expired).

PR #540, opened 2025-11-14
feat(streamable-http): support both SSE and JSON response formats

daixijun:
What this enables:

- Clients can accept either Server-Sent Events (SSE) or JSON responses
- Flexible content negotiation based on server preferences
- Improved interoperability with different MCP server implementations

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

Issue #526, opened 2025-11-07
Implement SEP-1442: Make MCP Stateless


felixweinberger:
This is a tracking issue for implementation of [SEP-1442](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442).

### Summary

This SEP focuses on enhancing MCP's ability to scale horizontally and operate in stateless or near-stateless modes. The Rust SDK needs to introduce patterns and protocol enhancements that allow servers to operate with minimal or no session state, enable seamless request distribution across multiple server instances, improve session handling mechanisms, optimize server startup performance, and provide guidelines for implementing stateless server architectures to support production deployments requiring high availability and horizontal scaling.

### Related Issues & PRs

- **Implementation PRs**: n/a
- **Related PRs**: n/a
- **Related Issues**: n/a

erenatas:
This didn't make it to last protocol release, nor it is approved, why is there a ticket on SDK for this? Same with other SEPs that have not made it to protocol

DaleSeo:
Superseded by #869

PR #683, opened 2026-02-24
feat(streamable-http): add json_response option for stateless server mode

thiagomendes:
## What this changes

Adds `json_response: bool` field to `StreamableHttpServerConfig`. When `true` and `stateful_mode` is false, the server returns `Content-Type: application/json` directly instead of `text/event-stream`.

## Why

While working on a multi-language MCP server benchmark comparing implementations across Rust, Go, Java and other runtimes under equivalent load, we noticed the Rust server (using this SDK) showed significantly lower throughput than the other runtimes, even when the business logic and external I/O were equivalent.

After profiling the request lifecycle, the bottleneck was not in the server logic itself but in the transport layer. In stateless mode, the SDK was wrapping every single response in a Server-Sent Events stream, even for straightforward request-response interactions that complete in a single round trip.

Looking at how other MCP SDKs handle this, the Go SDK already exposes this as a first-class option via `StreamableHTTPOptions`:

```go
mcp.NewStreamableHTTPHandler(func(r *http.Request) *mcp.Server {
    return server
}, &mcp.StreamableHTTPOptions{
    Stateless:    true,
    JSONResponse: true,
})
```

We looked for an equivalent in this SDK and found none. This PR adds it.

The MCP Streamable HTTP spec (2025-06-18) explicitly allows JSON responses as an alternative to SSE:

> "The server MUST reply with Content-Type: text/event-stream OR Content-Type: application/json"

In stateless mode, each request is handled via `OneshotTransport` with exactly one message in and one message out. Wrapping this in an SSE stream adds unnecessary overhead:

- Approximately 41ms fixed overhead per request, measured under load
- Keep-alive pings holding connections open unnecessarily
- Extra framing and parsing work on both client and server with no benefit

This change allows stateless servers to opt into the simpler and more efficient JSON response format that the spec already permits.

## Benchmark evidence

Measurement setup: 50 concurrent virtual users, 5 minutes sustained, 2 CPUs, 2 GB RAM limit, I/O-bound workload with external HTTP calls and Redis operations, representative of typical MCP tool implementations.

| | Before (SSE) | After (JSON) | Delta |
|---|---|---|---|
| RPS | 770 | 1139 | +48% |
| Tool avg latency (Redis + HTTP) | 41ms | 0.76ms | -98% |
| Tool avg latency (HTTP + pipeline) | 41ms | 0.55ms | -99% |
| Errors | 0 | 0 | none |

The overhead was traced to SSE framing on responses from tools that are fully stateless and return a single JSON result, exactly the case this flag addresses.

## Backward compatibility

- `json_response: false` (default) keeps SSE behaviour unchanged
- `stateful_mode: true` path is not touched at all
- All existing tests pass

## Relation to existing work

- Completes server-side JSON support (client-side added in #540)
- Contributes to SEP-1442 (#526) stateless server mode goals

## Types of changes

## Checklist

### DIFF
--- a/crates/rmcp/tests/test_sse_concurrent_streams.rs
+++ b/crates/rmcp/tests/test_sse_concurrent_streams.rs
@@ -84,6 +84,7 @@ async fn start_test_server(ct: CancellationToken, trigger: Arc<Notify>) -> Strin
             sse_keep_alive: Some(Duration::from_secs(15)),
             sse_retry: Some(Duration::from_secs(3)),
             cancellation_token: ct.child_token(),
+            ..Default::default()
         },
     );
 


crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -585,27 +591,51 @@ where
...
-                        tracing::info!(?message);
-                        ServerSseMessage {
-                            event_id: None,
-                            message: Some(Arc::new(message)),
-                            retry: None,
+                    if self.config.json_response {
+                        // JSON-direct mode: await the single response and return as
+                        // application/json, eliminating SSE framing overhead.
+                        // Allowed by MCP Streamable HTTP spec (2025-06-18).
+                        match receiver.recv().await {
```
DaleSeo: The SSE path sends `self.config.cancellation_token.child_token()` to `sse_stream_response`, making sure the stream stops when the server shuts down. However, this JSON path's `receiver.recv().await` could block forever without being aware of cancellations. If the handler gets stuck or the server is shutting down, this request will hang until the task finishes or fails. Should we think about wrapping it with `tokio::select!`?

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -585,27 +591,51 @@ where
...
+                            Some(message) => {
+                                let body = serde_json::to_vec(&message).map_err(|e| {
```
DaleSeo: I suggest adding `tracing::info!(?message)` here to match the SSE path. This will help operators maintain visibility over responses when JSON mode is enabled.

```suggestion
                            Some(message) => {
                                tracing::info!(?message);
                                let body = serde_json::to_vec(&message).map_err(|e| {
```

crates/rmcp/src/transport/streamable_http_server/tower.rs
thiagomendes: @DaleSeo 

Great catch! I've wrapped the receiver in a tokio::select! block to listen for self.config.cancellation_token and return None (UnexpectedEof) if the server shuts down before the message arrives, preventing any potential hangs. Fixed in the latest commit.

crates/rmcp/src/transport/streamable_http_server/tower.rs
thiagomendes: @DaleSeo 

Added! That makes total sense for parity with the SSE path and better observability. Fixed in the latest commit.

thiagomendes:
Thanks for the review @DaleSeo ! I have pushed a new commit addressing both points (cancellation awareness and observability parity). The local tests are passing and everything looks good.

DaleSeo:
Thanks for addressing my feedback, @thiagomendes. The code looks good, but there's one linting error. We should be good to merge the PR once this is fixed.

```sh
error[E0063]: missing field `json_response` in initializer of `rmcp::transport::StreamableHttpServerConfig`
  --> crates/rmcp/tests/test_sse_concurrent_streams.rs:82:9
   |
82 |         StreamableHttpServerConfig {
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `json_response`

For more information about this error, try `rustc --explain E0063`.
```

thiagomendes:
> Thanks for addressing my feedback, @thiagomendes. The code looks good, but there's one linting error. We should be good to merge the PR once this is fixed.
> 
> ```shell
> error[E0063]: missing field `json_response` in initializer of `rmcp::transport::StreamableHttpServerConfig`
>   --> crates/rmcp/tests/test_sse_concurrent_streams.rs:82:9
>    |
> 82 |         StreamableHttpServerConfig {
>    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `json_response`
> 
> For more information about this error, try `rustc --explain E0063`.
> ```

@DaleSeo Fixed, thanks for catching it! Pushed the correction. CI should be green now.

PR #699, opened 2026-02-26
fix: Commits on Feb 27, 2026 fix: downgrade logging of message to `TRACE` to avoid spamming logs

kristof-mattei:
## Motivation and Context

Reduces spam in the logs.

## How Has This Been Tested?

Yes, validated that the log is no longer being printed. 

## Breaking Changes

No.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -608,7 +608,7 @@ where
                             _ = cancel.cancelled() => None,
                         } {
                             Some(message) => {
-                                tracing::info!(?message);
+                                tracing::trace!(?message);
                                 let body = serde_json::to_vec(&message).map_err(|e| {
                                     internal_error_response("serialize json response")(e)
                                 })?;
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -608,7 +608,6 @@ where
                             _ = cancel.cancelled() => None,
                         } {
                             Some(message) => {
-                                tracing::info!(?message);
```
DaleSeo, line 611: > We could log them in both situations at a TRACE level?

Having some visibility into the messages flowing through stateless mode can really help with debugging. Since TRACE is filtered out by default, it won't clutter the logs.

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -627,14 +626,12 @@ where
                         }
                     } else {
                         // SSE mode (default): original behaviour preserved unchanged
-                        let stream = ReceiverStream::new(receiver).map(|message| {
-                            tracing::info!(?message);
```
DaleSeo, line 631: Same here.

kristof-mattei:
@DaleSeo addressed, moved to `trace!`.

Issue #379, opened 2025-08-18
WASM Support for rmcp


joulei:
Hi! I'm working on integrating `rmcp` into a project that requires WASM compatibility and noticed that rmcp currently doesn't compile for wasm32-unknown-unknown due to several WASM-specific blockers, mostly `Send` requirements.

### Current Issues:

1. Send trait bounds incompatibility
    - traits like`SseClient` requiree `Send + Sync + 'static`
    - Send is a no go in web:
        - JavaScript interop types contain `PhantomData<*const u8>` making them intentionally !Send
        - reqwest's WASM implementation uses `JsValue` types that are !Send by design

2. Transport trait architecture
    - All transport implementations must return `impl Future + Send`

3. Worker transport assumptions
    - Worker trait expects thread spawning capabilities
    - WASM is single-threaded, no `std::thread` support

4. Service layer threading
    - `ServiceRole`, `Service` traits all require `Send + Sync`
    - Future return types across the codebase assume `Send` bounds

### Questions:
  1. Is WASM support planned for rmcp in the near future?
  2. Is anyone currently working on this?

### Proposal:
I'm willing to work on this if there's interest (I might do either way in a fork, as I'd really like to avoid making my own rmcp). We've recently had to deal with similar issues making some cross-platform async Rust projects wasm-compatible, there are some patterns we've used that could work here:

1. Platform-conditional trait abstractions:
```rust
  cfg_if::cfg_if! {
      if #[cfg(target_arch = "wasm32")] {
          pub trait PlatformSend {}
          impl<T> PlatformSend for T {}
      } else {
          pub trait PlatformSend: Send {}
          impl<T> PlatformSend for T where T: Send {}
      }
  }

  pub trait SseClient: Clone + PlatformSend + 'static { ... }
  pub type BoxPlatformSendFuture<T> = Pin<Box<dyn Future<Output = T> + PlatformSend>>;
```

2. ThreadToken pattern for WASM:
  - If needed, use thread-local storage for non-Send types that need to cross Send boundaries
  - ^ safe access to JavaScript objects in single-threaded WASM context

3. Conditional spawn implementations:
```rust
  #[cfg(not(target_arch = "wasm32"))]
  fn spawn_worker(fut: impl Future + Send + 'static) { tokio::spawn(fut); }

  #[cfg(target_arch = "wasm32")]
  fn spawn_worker(fut: impl Future + 'static) { wasm_bindgen_futures::spawn_local(fut); }
```

[...]

PR #710, opened 2026-02-28
Remove dependency on tokio/rt and `tokio::spawn(...)`

### LINKED ISSUE #290: Remove dependency on `tokio/rt`
Since this project is a library, it is unfavorable that it includes its own async runtime from `tokio/rt`. If a project using the `async-std` runtime or some other runtime tries to use this project, the `tokio` runtime will now be needed to be running in addition to it, if there are no compatibility issues. Additionally, if a project uses a *different* version of `tokio` the runtimes may be incompatible.

Specifically, calls to `tokio::spawn` will panic if not called within a Tokio runtime. This is bad news for projects that include this and use a different runtime. 

Remove dependency on `tokio/rt` feature, and any calls to `tokio::spawn` in library code.

N/A

**Additional context**
A library crate bringing it's own async runtime is bad, and forces the consuming project of this crate to make more significant decisions about using the same runtime, having a larger binary size with two runtimes, or in the worst case, compatibility issues. 

Note that removing the `tokio/rt` feature means that some portions that rely on `tokio::task::JoinSet` can not compile. Consider refactoring these areas with `futures::stream::FuturesUnordered` from the `futures` crate. Additionally, some spots that use `tokio::spawn` and `JoinHandle` may be replaceable with plain `Pin<Box<dyn Future<Output = T> + Send + Sync + 'static>>`s. And in complex cases, you may even consider `futures::future::shared` to poll the same future from multiple spots. 

I understand this may be a large refactor, perhaps one for a major release. Async Rust also gets a lot harder when we can't just `spawn` tasks, adding to the complexity of this refactor.

AadamZ5:
These changes aim to remove most calls to `tokio::spawn(...)` and other runtime-intertwined calls. These changes are to support runtime-agnosticism.

**Heads Up:** There are a lot of changes, and I acknowledge that may make this PR unsuitable to merge.

## Motivation and Context

This will allow the project to be used in other environments where `tokio/rt` is not available or cannot be used. I originally raised this issue in #290, but some real demand for this particular change may be present in #379. 

Ultimately, the goal is to remove dependency on the Tokio runtime. However, as I progress, I see a few repeated smaller sub-issues I am trying to tackle to accomplish the larger goal:
 - Remove calls to `tokio::spawn(...)` (which mostly encompasses below)
 - Remove async side effects (via `tokio::spawn(...)`) in `Drop` implementations
 - Give control to the caller for how to run asynchronous background work
 - Make child-process adaptable to different async implementations

After my initial changes, there is *a lot* of the following refactor:
```rust
let client = ().serve(transport).await?;
// use client...
```
turns into
```rust
let (client, work) = ().serve(transport).await?;
// now the library user must explicitly spawn the work loop for the service. 
tokio::spawn(work);
// use client...
```

## How Has This Been Tested?

This refactor should be testable via the existing unit tests. As development continues, we can add more unit tests if needed. I also hope to perform real smoke-test, minimally using the already existing examples in this repo. 

## Breaking Changes

Yes, users will most likely need to update their code. This will very likely be a breaking change.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

This is an ongoing draft. Please feel free to interject with comments or concerns. 

Could eventually close #290


AadamZ5:
Next note for myself: `crates/rmcp/src/transport/child_process.rs` will need a lot of consideration about how to refactor to be runtime-agnostic or at least adaptable. We could include feature-guarded implementations of a new `ChildProcessRunner` trait, perhaps wrapped in a common struct to ease usage. 

Idea:
```rs
/// Underlying implementation and contract about what a command runner needs to provide
pub trait ChildProcessRunner {
   pub fn send_input(&self, input &[u8]) -> Result<...>;

   pub fn get_output_stream(&self) -> Result<PinnedStream<Vec<u8>>, ...>;
   pub fn get_exit_code(&self) -> Result<u8, ...>;
}

/// The `ChildProcess` struct would "control" the runner, and properly orchestrate methods calls (like
/// ensuring only one call to `get_output_stream(...)` etc)
///
/// This would simplify usage for the caller or library user
pub struct ChildProcess {
   inner: Box<dyn ChildProcessRunner>;
}
```

Then we could provide different implementations for the command runner, or provide a way for the library consumer to provide their own. 

All of the ChildProcess usage might need to be behind a feature-flag. 

**Edit:** I discovered the project is using the `process_wrap` crate, going to learn about that first before I make any major changes :)

AadamZ5:
The new `ChildProcess` API seems to be functional. We can later write an `async-std` implementation now without changing too many concrete types :grinning: I hope that refactor is not too disruptive from a user standpoint. I did try to update and run unit tests involving `ChildProcess` stuff. It could use a little more documentation about how to write one's own implementation for command running. 

Still need to continue eliminating calls to `tokio::spawn(...)`. There are plenty still, but a manageable amount! And fyi, I am not removing calls to `tokio::spawn(...)` in examples or tests. I think those are typically appropriate places to setup and use a runtime.

crates/rmcp/src/transport/child_process.rs
```diff
@@ -1,309 +1,6 @@
...
-#[cfg(unix)]
-#[cfg(test)]
-mod tests {
-    use tokio::process::Command;
-
-    use super::*;
-
-    #[tokio::test]
-    async fn test_tokio_child_process_drop() {
-        let r = TokioChildProcess::new(Command::new("sleep").configure(|cmd| {
```
AadamZ5, line 245: Oops, need to bring back some command tests.

crates/rmcp/src/transport/common/reqwest.rs
```diff
@@ -1,2 +1,2 @@
 #[cfg(feature = "transport-streamable-http-client-reqwest")]
-mod streamable_http_client;
```
AadamZ5, line 2: This did not need to be changed.

crates/rmcp/src/transport/common.rs
```diff
@@ -4,7 +4,7 @@ pub mod server_side_http;
 pub mod http_header;
 
 #[cfg(feature = "__reqwest")]
-mod reqwest;
```
AadamZ5, line 7: This did not need to be changed

AadamZ5:
Note for me:

Need to find all references to `ServiceExt::serve(...)` (in `crates/rmcp/src/service.rs`) and refactor into tuple unboxing and spawning work

AadamZ5:
We need to remove calls to `tokio::time::timeout(...)` since it cannot work if it is not inside of a Tokio runtime. 

There are a few portable candidates, but I will limit it to these two for now:
- [`futures-timeout`](https://crates.io/crates/futures-timeout): A very small and focused timeout library for futures. It spawns it's own thread to manage timers independently from any runtime. (wasm compatible)
- [`fastimer`](https://crates.io/crates/fastimer): Similar small library, focused on timing-out futures with it's own state managed in a separate thread. 

I have personally used `futures-timeout` before, and enjoy it's simple timeout API, and it is also wasm-compatible. It also has slightly more all-time downloads. I will go with that.

AadamZ5:
Okay, I think this is largely done. All async work loops are now returned at construction of services or clients. Timeouts have also been changed to `futures-timeout`.

Just merged in latest main into my dev branch and resolved any conflicts.

`cargo test --all-features` passes with no failures inside of the dev container 💪

I have not verified all examples yet, will try to do that on this weekend at the latest. They did receive a pass part-way through my refactoring, but not a final pass yet.

AadamZ5:
@alexhancock I see there are a few merge conflicts, so feel free to ping me if there is any more support needed here. Also, no hard feelings if this doesn't seem like what the team wants to merge in! I just took a swing at it

alexhancock:
Thanks for the earlier push on this @AadamZ5. It's a worthy goal but I agree that it's likely too painful to rebase and merge in current form.

But it is worth thinking through if there is an updated version we could do, and what the user-facing code changes required would look like for the case where a user wanted to continue using tokio or move to something new. If we can find something that doesn't introduce too much user facing pain (or could be handled by an agent skill) I am ok with a large change to the codebase to enable it.

AadamZ5:
@alexhancock Totally understand. I think #290 could use a little more presentation about possible solutions. While this one may have solved the issue, it definitely changed a lot of return types of methods, impacting any call sites. 

If I get time, I'd love to write up a better comment in the issue thread of the why, and the two major solutions I see. I can't commit to getting that in soon, but if you find this issue becomes a higher priority please ping me and maybe I can prioritize this!

PR #715, opened 2026-03-02
chore: add #[non_exhaustive] and mutation methods to improve compatibility

jamadeo:
In preparation for a 1.0 release, we should use #[non_exaustive] for most public structs/enums to help reduce backwards-incompatible changes.

While we don't foresee being strict with the semver guidelines, we should aim to keep backwards compatibility as much as possible between minor releases.

## Motivation and Context
This will prevent clients from writing code that will break and require updates in future versions.

## Types of changes

- [ ] Documentation update

## Checklist


alexhancock (approved):

crates/rmcp/src/model.rs
```diff
@@ -1201,6 +1444,27 @@ pub struct LoggingMessageNotificationParam {
...
+
+    /// Create with a logger name.
+    pub fn with_logger(level: LoggingLevel, logger: impl Into<String>, data: Value) -> Self {
+        Self {
+            level,
+            logger: Some(logger.into()),
+            data,
+        }
+    }
+}
```
DaleSeo, line 1466: The `with_` prefix means "take ownership of self, set a fieThe `with_` prefix means "take ownership of self, set a field, and return self." When callers look at the method list, they'll expect to see `new().with_logger("name")` chaining. I suggest we stick to the conventional build pattern here.

```suggestion
    pub fn with_logger(mut self, logger: impl Into<String>) -> Self {
        self.logger = Some(logger.into());
        self
    }
```

jamadeo: @DaleSeo I agree. I think most of them follow this pattern you're suggesting, and we should keep it consistent

DaleSeo: @jamadeo Quick follow-up: https://github.com/modelcontextprotocol/rust-sdk/pull/720

PR #720, opened 2026-03-03
fix: api ergonomics follow-up

DaleSeo:
## Motivation and Context

Just a quick follow-up on PR #715. This update fixes a few issues with the `with_*` builder methods before we finalize everything at 1.0.

The new methods now accept `Option<T>` instead of just `T`. This means you can use `.with_title(None)`, which compiles but doesn't do anything and will trigger a warning from Clippy. It's simpler to just skip the method if you don't have a value. Also, two methods, `with_logger` and `with_content`, had the `with_` prefix but were actually standalone constructors, which was confusing. The macro code in `rmcp-macros` has been updated to reflect this change, so it won't call the method if an attribute isn't set.

Since we're already making changes on the way to 1.0, it makes sense to include this now. After the stable release, we would need a whole new major version for what is really just a minor issue.

## How Has This Been Tested?

All existing tests pass without modification to their logic.

## Breaking Changes

These methods haven't been released yet, so in practice no consumers should be affected.

`with_*` methods that accepted `Option<T>` now take `T`:

```rust
// before
.with_title(Some("my tool".to_string()))
.with_stop_reason(Some(reason))

// after
.with_title("my tool")
.with_stop_reason(reason)
```

If you were passing `None`, just remove the call.

`with_logger` and `with_content` are now builder setters instead of constructors:

```rust
// before
LoggingMessageNotificationParam::with_logger(level, "my-logger", data)

// after
LoggingMessageNotificationParam::new(level, data).with_logger("my-logger")
```

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp-macros/src/tool.rs
+++ b/crates/rmcp-macros/src/tool.rs
@@ -3,7 +3,7 @@ use proc_macro2::{Span, TokenStream};
 use quote::{ToTokens, format_ident, quote};
 use syn::{Expr, Ident, ImplItemFn, LitStr, ReturnType, parse_quote};
 
-use crate::common::{extract_doc_line, none_expr};
+use crate::common::extract_doc_line;
 
 /// Check if a type is Json<T> and extract the inner type T
 fn extract_json_inner_type(ty: &syn::Type) -> Option<&syn::Type> {


crates/rmcp-macros/src/tool.rs
```diff
@@ -134,42 +134,37 @@ impl ResolvedToolAttribute {
...
-                .with_raw_output_schema(#output_schema)
-                .with_annotations(#annotations)
-                .with_execution(#execution)
-                .with_icons(#icons)
-                .with_meta(#meta)
+                #title_call
+                #output_schema_call
+                #icons_call
+                #meta_call;
+                if let Some(__annotations) = #annotations {
```
alexhancock: why are `annotations` and `execution` done this way instead of the same way as title/icons/meta?

DaleSeo: Good catch, @alexhancock! Refactored the code so all optional fields follow the same pattern and the generated code is a clean method chain again.

PR #723, opened 2026-03-04
fix: conformance syntax changes

alexhancock:
Required to get the conformance server & client running again after the `non_exhaustive` refactor

### DIFF
--- a/conformance/src/bin/client.rs
+++ b/conformance/src/bin/client.rs
@@ -5,8 +5,7 @@ use rmcp::{
     model::*,
     service::RequestContext,
     transport::{
-        AuthClient, AuthorizationManager, StreamableHttpClientTransport,
-        auth::{OAuthClientConfig, OAuthState},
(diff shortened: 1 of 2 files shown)


conformance/src/bin/client.rs
```diff
@@ -16,6 +15,7 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 // ─── Context parsed from MCP_CONFORMANCE_CONTEXT ────────────────────────────
 
 #[derive(Debug, Default, serde::Deserialize)]
+#[allow(dead_code)]
```
DaleSeo: Is it for backward compatibility? `ConformanceContext::name` (and possibly other fields) now appears to go unused after the `OAuthClientConfig` import was removed.

alexhancock: made a small change here to address and it needs new review

PR #726, opened 2026-03-05
chore: add release-plz config to include conformance server/client in considering changes

alexhancock:
This makes it so `release-plz` will track and manage the `conformance` server and client as if they were released packages without them actually being published to crates.io

### DIFF
--- a/conformance/Cargo.toml
+++ b/conformance/Cargo.toml
@@ -1,8 +1,10 @@
 [package]
 name = "mcp-conformance"
-version = "0.1.0"
-edition = "2021"
-publish = false
+version = { workspace = true }
+edition = { workspace = true }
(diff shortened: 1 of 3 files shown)


conformance/Cargo.toml
```diff
@@ -1,8 +1,10 @@
 [package]
 name = "mcp-conformance"
-version = "0.1.0"
-edition = "2021"
-publish = false
```
DaleSeo, line 5: I think we should keep this to avoid publishing this crate to crates.io. See https://github.com/modelcontextprotocol/rust-sdk/pull/701

alexhancock:
I don't think we need this any more. We've released a version that will tag the repo with a working version of the conformance server, which was my only goal.

PR #731, opened 2026-03-07
feat: include granted scopes in OAuth refresh token request

### LINKED ISSUE #624: Support passing scopes when doing a refresh token flow
  Refreshing access tokens currently sends only grant_type=refresh_token and refresh_token, with no scope. RFC 6749 allows scope on refresh and says it can be used (and if omitted, it defaults to the originally granted scope).  (datatracker.ietf.org (https://datatracker.ietf.org/doc/rfc6749)).  In practice, some providers require or expect scopes on refresh. In our case with Azure AD v2 (/oauth2/v2.0/token), the refresh call fails unless scopes are included, even though the initial authorization succeeds. The metadata we advertise  includes scopes like api://…/all and offline_access. Microsoft’s docs emphasize scopes in v2 token flows and that offline_access is required to receive refresh tokens. (learn.microsoft.com (https://learn.microsoft.com/EN-US/azure/
  active-directory/develop/scopes-oidc?utm_source=openai))

  Allow passing scopes when requesting a new token via refresh (e.g., reuse the scopes from the stored token response, or accept an explicit scopes parameter for refresh). This aligns with RFC 6749’s refresh grant, which allows the scope  parameter on refresh requests. (datatracker.ietf.org (https://datatracker.ietf.org/doc/rfc6749/)).  Concretely: when calling exchange_refresh_token, add scopes from the existing credential’s scope set (if present), or allow callers to override.

  - Re-authenticate instead of refresh (works but is disruptive).
  - Patch/fork to add scopes at the app level.
  - Require all servers to only use providers that tolerate scope-less refresh, which is not realistic.

**Additional context**
  RFC 6749 explicitly allows scope on refresh and states that if omitted it is treated as the originally granted scope;  this suggests clients may need to carry scopes forward. (datatracker.ietf.org (https://datatracker.ietf.org/doc/
  rfc6749/)).  OAuth.com’s summary of the refresh flow also notes scope as an optional parameter on refresh. (oauth.com (https://www.oauth.com/oauth2-servers/access-tokens/refreshing-access-tokens/))

DaleSeo:
Fixes #624

## Motivation and Context

[RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-6) allows sending a `scope` parameter with refresh token requests, and some providers, especially Azure AD v2, actually require it. If you don't include it, the refresh might succeed quietly, but it will return a token without any scopes, or the server might reject the request altogether. 

This fix keeps the full `StoredCredentials`, allowing us to add the previously granted scopes to the outgoing refresh request using `.add_scope()`. This follows the same pattern we already use in the authorization code and client credentials flows. If `granted_scopes` is empty, we won't send a `scope` parameter, which maintains the current behavior for providers that don't need it.

https://github.com/modelcontextprotocol/rust-sdk/blob/5c5a2e734d9c6b4872f4797ca8fdee6b2ac72f2e/crates/rmcp/src/transport/auth.rs#L913

https://github.com/modelcontextprotocol/rust-sdk/blob/5c5a2e734d9c6b4872f4797ca8fdee6b2ac72f2e/crates/rmcp/src/transport/auth.rs#L1766

## How Has This Been Tested?

Added unit tests

## Breaking Changes

None. Providers that don't accept a `scope` parameter on refresh requests will continue to work as before, since the parameter is only added when scopes were previously granted.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/OAUTH_SUPPORT.md
+++ b/docs/OAUTH_SUPPORT.md
@@ -127,7 +127,7 @@ cargo run -p mcp-client-examples --example clients_oauth_client
 6. **Authorization Request**: Build authorization URL with PKCE (S256) and RFC 8707 resource parameter
 7. **Authorization Code Exchange**: After user authorization, exchange code for access token (with resource parameter)
 8. **Token Usage**: Use access token for API calls via `AuthClient` or `AuthorizedHttpClient`
-9. **Token Refresh**: Automatically use refresh token to get new access token when current one expires
+9. **Token Refresh**: Automatically use refresh token to get new access token when current one expires; previously granted scopes are forwarded in the refresh request so providers that require them (e.g. Azure AD v2) work correctly
 10. **Scope Upgrade**: On 403 insufficient_scope, compute scope union and re-authorize with upgraded scopes
 
 ## Security Considerations
@@ -158,3 +158,4 @@ If you encounter authorization issues, check the following:
 - [RFC 8707: Resource Indicators for OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc8707)
 - [RFC 9728: OAuth 2.0 Protected Resource Metadata](https://datatracker.ietf.org/doc/html/rfc9728)
 - [RFC 7636: Proof Key for Code Exchange (PKCE)](https://datatracker.ietf.org/doc/html/rfc7636)
+- [RFC 6749 §6: Refreshing an Access Token](https://www.rfc-editor.org/rfc/rfc6749#section-6)


crates/rmcp/src/transport/auth.rs
```diff
@@ -3580,4 +3585,170 @@ mod tests {
...
+        manager.credential_store.save(stored).await.unwrap();
+
+        let err = manager.refresh_token().await.unwrap_err();
+        assert!(
+            matches!(err, AuthError::TokenRefreshFailed(_)),
+            "expected TokenRefreshFailed when no refresh token, got: {err:?}"
+        );
+    }
+
+    async fn start_token_server() -> (String, Arc<std::sync::Mutex<Option<String>>>) {
```
alexhancock, line 3653: we should `#[cfg(test)]` this yeah?

DaleSeo: This is a private helper inside the `#[cfg(test)]` mod tests block, so it should be already gated.

PR #735, opened 2026-03-09
fix(rmcp-macros): use re-exported serde_json path in task_handler

### LINKED ISSUE #487: `tool` macro expands to `serde_json` without re-export
**Describe the bug**
tool macro expands to `serde_json` [here](https://github.com/modelcontextprotocol/rust-sdk/blob/11093bc830d77717b8f24813d888238a12d7eefe/crates/rmcp-macros/src/tool.rs#L205) which causes compilation error in crates that don't have it as a dependency.

**To Reproduce**
1. cargo new asdf
2. cd asdf
3. cargo add rcmp -F macros
4. add basic tool call to src/main.rs
```rs
struct A;

#[rmcp::tool_router]
impl A {
    #[rmcp::tool]
    async fn foo() {}
}
```
5. error[E0433]: failed to resolve: use of unresolved module or unlinked crate `serde_json`

**Proposed solution**
[Other location](https://github.com/modelcontextprotocol/rust-sdk/blob/11093bc830d77717b8f24813d888238a12d7eefe/crates/rmcp-macros/src/tool.rs#L199) in the macro already assumes the downstream crate depends on `rmcp`, so expanding the macro to use the re-exported `::rmcp::serde_json::...` should work.

kakarot-dev:
## Summary

The `task_handler` macro generates code referencing `::serde_json::` directly. This causes compilation errors in crates that depend on `rmcp` but don't have `serde_json` as a direct dependency.

Changed `::serde_json::` to `::rmcp::serde_json::` in `crates/rmcp-macros/src/task_handler.rs` to use the re-exported path, consistent with how other generated code in the macro crate references rmcp dependencies.

Fixes #487

### DIFF
--- a/crates/rmcp-macros/src/task_handler.rs
+++ b/crates/rmcp-macros/src/task_handler.rs
@@ -193,7 +193,7 @@ pub fn task_handler(attr: TokenStream, input: TokenStream) -> syn::Result<TokenS
                                     if let Some(tool) = boxed.as_any().downcast_ref::<rmcp::task_manager::ToolCallTaskResult>() {
                                         match &tool.result {
                                             Ok(call_tool) => {
-                                                let value = ::rmcp::serde_json::to_value(call_tool).unwrap_or(::rmcp::serde_json::Value::Null);
+                                                let value = ::rmcp::serde_json::to_value(call_tool).unwrap_or_default();
                                                 return Ok(rmcp::model::GetTaskPayloadResult::new(value));
                                             }
                                             Err(err) => return Err(McpError::internal_error(


crates/rmcp-macros/src/task_handler.rs
```diff
@@ -193,7 +193,7 @@ pub fn task_handler(attr: TokenStream, input: TokenStream) -> syn::Result<TokenS
                                     if let Some(tool) = boxed.as_any().downcast_ref::<rmcp::task_manager::ToolCallTaskResult>() {
                                         match &tool.result {
                                             Ok(call_tool) => {
-                                                let value = ::serde_json::to_value(call_tool).unwrap_or(::serde_json::Value::Null);
+                                                let value = ::rmcp::serde_json::to_value(call_tool).unwrap_or(::rmcp::serde_json::Value::Null);
```
DaleSeo: This should be slightly more idiomatic:

```suggestion
                                                let value = ::rmcp::serde_json::to_value(call_tool).unwrap_or_default();
```

PR #740, opened 2026-03-09
feat: add local feature for !Send tool handler support

### LINKED ISSUE #728: Feature request: Support for !Send futures in #[tool] handlers (e.g. #[tool(local)])
## Problem

The `#[tool]` macro currently requires all tool handler futures to be `Send`. This `Send` bound propagates virally through the entire async call chain — not just the handler itself, but every function it awaits transitively.

This creates a fundamental incompatibility with popular async libraries that return `!Send` futures:

- **`sqlx::raw_sql()`** — returns a future holding `PhantomData<fn(&mut PgConnection)>` which is `!Send`. Cannot be used in any function that's transitively called from a `#[tool]` handler.
- **`sqlx` transactions** — transaction futures borrow `&mut PgConnection` across await points, making them `!Send` in many patterns.
- Other database libraries with similar patterns.

### Concrete Example

```rust
// This works (sqlx::query returns Send future):
#[tool]
async fn my_tool(&self) -> Result<String> {
    sqlx::query("SELECT 1").execute(&self.pool).await?;
    Ok("done".into())
}

// This fails to compile (sqlx::raw_sql returns !Send future):
#[tool]  
async fn my_tool(&self) -> Result<String> {
    sqlx::raw_sql("CREATE TABLE IF NOT EXISTS ...").execute(&self.pool).await?;
    Ok("done".into())
}
```

The error propagates even through indirect calls:
```
error[E0277]: `(dyn Any + Send + 'static)` cannot be sent between threads safely
```

### Current Workaround

We split SQL strings by `;` and execute each statement via `sqlx::query()` individually. This requires writing a custom SQL statement parser to handle semicolons inside string literals, comments, and dollar-quoted strings — significant complexity for what should be a simple migration runner.

## Proposed Solution

Add a `#[tool(local)]` attribute (or similar) that allows handlers to return `!Send` futures. Implementation options:

1. **`#[tool(local)]`** — use `tokio::task::LocalSet` for this specific handler
2. **`#[tool(spawn_blocking)]`** — run the handler in `spawn_blocking` context  
3. **Relax the Send bound globally** — if the MCP server doesn't actually need to send handler futures across threads (e.g., if it uses a single-threaded runtime or `LocalSet` internally)

## Environment

- rmcp version: 1.1.0
- sqlx version: 0.8+
- Rust: stable (edition 2024)

DaleSeo:
Closes #728

## Motivation and Context

Some MCP server use cases involve state that isn't `Send`. For example, this includes holding a raw pointer, an `Rc`, or a thread-local resource. The current service stack requires `Send + Sync` everywhere, which completely rules these out. This PR introduces a `local` Cargo feature that relaxes all `Send` and `Sync` bounds across the service stack. This change allows servers with `!Send` handlers to run on a single-threaded runtime using `tokio::task::LocalSet` and `spawn_local`.

## How Has This Been Tested?

The full test suite passes in both modes: 
- `cargo test --features server,client` (default, no behavior change)
- `cargo test --features server,client,local` (new path). 

## Breaking Changes

None. Without the `local` feature the effective trait bounds are identical to before — `MaybeSend` resolves to `Send + Sync` and `MaybeSendFuture` resolves to `Send` via blanket impls. The feature is strictly opt-in.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/README.md
+++ b/crates/rmcp/README.md
@@ -52,7 +52,7 @@ The transport layer is pluggable. Two built-in pairs cover the most common cases
 | | Client | Server |
 |:-:|:-:|:-:|
 | **stdio** | [`TokioChildProcess`](crate::transport::TokioChildProcess) | [`stdio`](crate::transport::stdio) |
-| **Streamable HTTP** | [`StreamableHttpClientTransport`](crate::transport::StreamableHttpClientTransport) | [`StreamableHttpService`](crate::transport::StreamableHttpService) |
+| **Streamable HTTP** | [`StreamableHttpClientTransport`](crate::transport::StreamableHttpClientTransport) | `StreamableHttpService` |
 
 Any type that implements the [`Transport`](crate::transport::Transport) trait can be used. The [`IntoTransport`](crate::transport::IntoTransport) helper trait provides automatic conversions from:
 
--- a/crates/rmcp/src/handler/client.rs
+++ b/crates/rmcp/src/handler/client.rs
@@ -4,7 +4,9 @@ use std::sync::Arc;
 use crate::{
     error::ErrorData as McpError,
     model::*,
-    service::{NotificationContext, RequestContext, RoleClient, Service, ServiceRole},
+    service::{
+        MaybeSendFuture, NotificationContext, RequestContext, RoleClient, Service, ServiceRole,
+    },
 };
 
 impl<H: ClientHandler> Service<RoleClient> for H {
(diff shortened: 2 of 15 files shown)


alexhancock (approved):
It LGTM as an opt-in thing.

Perhaps good to get @Stranmor to look as well to double confirm it addresses the need?

Stranmor:
This PR fully addresses the use case I described in #728. The global `local` feature approach works well for us — MCP servers are typically single-threaded, so relaxing Send+Sync bounds across the board is a practical trade-off. Appreciate the clean implementation with `MaybeSend`/`MaybeSyncSend` traits. Thanks @DaleSeo for picking this up!

PR #764, opened 2026-03-20
fix(http): add host check

jokemanfire:
This is a security problem ,there is a lack of host access verification here, which may be exploited by malicious attacks.

## Motivation and Context

Adds Host validation to the Streamable HTTP server to mitigate DNS rebinding risks. Introduces configurable allowed hosts,  and adds integration tests for both valid localhost traffic and malicious request scenarios.

## How Has This Been Tested?

Add the it test
## Breaking Changes

Yes, the sec change may change the mcp's server feature which has been deployed.
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -2,7 +2,7 @@ use std::{convert::Infallible, fmt::Display, sync::Arc, time::Duration};
 
 use bytes::Bytes;
 use futures::{StreamExt, future::BoxFuture};
-use http::{Method, Request, Response, header::ALLOW};
+use http::{HeaderMap, Method, Request, Response, header::ALLOW};
 use http_body::Body;
 use http_body_util::{BodyExt, Full, combinators::BoxBody};
 use tokio_stream::wrappers::ReceiverStream;
@@ -29,8 +29,8 @@ use crate::{
     },
 };
 
-#[derive(Debug, Clone)]
 #[non_exhaustive]
+#[derive(Debug, Clone)]
 pub struct StreamableHttpServerConfig {
     /// The ping message duration for SSE connections.
     pub sse_keep_alive: Option<Duration>,
@@ -49,6 +49,16 @@ pub struct StreamableHttpServerConfig {
     /// When this token is cancelled, all active sessions are terminated and
     /// the server stops accepting new requests.
     pub cancellation_token: CancellationToken,
+    /// Allowed hostnames or `host:port` authorities for inbound `Host` validation.
+    ///
+    /// By default, Streamable HTTP servers only accept loopback hosts to
+    /// prevent DNS rebinding attacks against locally running servers. Public
+    /// deployments should override this list with their own hostnames.
+    /// examples:
+    ///     allowed_hosts = ["localhost", "127.0.0.1", "0.0.0.0"]
+    /// or with ports:
+    ///     allowed_hosts = ["example.com", "example.com:8080"]
+    pub allowed_hosts: Vec<String>,
 }
 
 impl Default for StreamableHttpServerConfig {
@@ -59,11 +69,24 @@ impl Default for StreamableHttpServerConfig {
             stateful_mode: true,
             json_response: false,
             cancellation_token: CancellationToken::new(),
+            allowed_hosts: vec!["localhost".into(), "127.0.0.1".into(), "::1".into()],
         }
     }
 }
 
 impl StreamableHttpServerConfig {
+    pub fn with_allowed_hosts(
+        mut self,
+        allowed_hosts: impl IntoIterator<Item = impl Into<String>>,
+    ) -> Self {
+        self.allowed_hosts = allowed_hosts.into_iter().map(Into::into).collect();
+        self
+    }
+    /// Disable allowed hosts. This will allow requests with any `Host` header, which is NOT recommended for public deployments.
+    pub fn disable_allowed_hosts(mut self) -> Self {
+        self.allowed_hosts.clear();
+        self
+    }
     pub fn with_sse_keep_alive(mut self, duration: Option<Duration>) -> Self {
         self.sse_keep_alive = duration;
         self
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -102,6 +124,99 @@ fn validate_protocol_version_header(headers: &http::HeaderMap) -> Result<(), Box
...
+    };
+    let Some(scheme) = uri.scheme_str() else {
+        return Err(forbidden_response("Forbidden: Invalid Origin header"));
+    };
+    if !matches!(scheme, "http" | "https") {
+        return Err(forbidden_response("Forbidden: Invalid Origin header"));
+    }
+
+    Ok(Some(normalize_host(authority.host())))
+}
```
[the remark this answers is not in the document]
jokemanfire: Great

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -102,6 +124,99 @@ fn validate_protocol_version_header(headers: &http::HeaderMap) -> Result<(), Box
...
+    let host = parse_host_header(headers)?;
+    if let Some(host) = host.as_deref() {
+        if !host_is_allowed(host, &config.allowed_hosts) {
+            return Err(forbidden_response("Forbidden: Host header is not allowed"));
+        }
+    }
+
+    let origin = parse_origin_host(headers)?;
+    if let Some(origin) = origin.as_deref() {
+        if !host_is_allowed(origin, &config.allowed_hosts) {
```
DaleSeo: Is it intentional to checking `Origin` against `allowed_hosts` here? `allowed_hosts` answers "what hostnames is this server known as?" while `Origin` represents the source page that initiated the request. These are different concepts. 

For example, consider an MCP server at `mcp.example.com` being called by a frontend at `app.example.com`:

```
allowed_hosts: ["mcp.example.com"]

Host: mcp.example.com           // ✅ this IS the server
Origin: http://app.example.com  // ❌ rejected, but this is a legitimate caller
```

From my understanding, restricting which origins can call the server is a [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) concern. For DNS rebinding protection, validating only the `Host` header should be sufficient.

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -102,6 +124,99 @@ fn validate_protocol_version_header(headers: &http::HeaderMap) -> Result<(), Box
...
+            return Err(forbidden_response(
+                "Forbidden: Origin header is not allowed",
+            ));
+        }
+        if let Some(host) = host.as_deref() {
+            if origin != host {
```
DaleSeo: The `Host` header identifies the destination server while `Origin` identifies the source page. They will naturally differ in any cross-origin scenario.

Would you consider removing the `Origin` validation entirely from this DNS rebinding guard and leaving origin-based restrictions to CORS middleware where they belong?

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -102,6 +124,99 @@ fn validate_protocol_version_header(headers: &http::HeaderMap) -> Result<(), Box
...
+        .status(http::StatusCode::FORBIDDEN)
+        .body(Full::new(Bytes::from(message.into())).boxed())
+        .expect("valid response")
+}
+
+fn normalize_host(host: &str) -> String {
+    host.trim_matches('[')
+        .trim_matches(']')
+        .to_ascii_lowercase()
+}
```
DaleSeo, line 167: `normalize_host` does not preserve the port, and `parse_host_header` uses `authority.host()`, which also strips the port. As a result, both `localhost:8000` and `localhost:9999` normalize to `localhost` and pass validation equally. This is significant because multiple services can run on different ports of the same host. Without port validation, a DNS rebinding attack could target a different service on `localhost:9999`, and the check would pass since `localhost` is in the allow list. Would you consider adding similar port-aware validation here?

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -102,6 +124,99 @@ fn validate_protocol_version_header(headers: &http::HeaderMap) -> Result<(), Box
...
+            ));
+        }
+        if let Some(host) = host.as_deref() {
+            if origin != host {
+                return Err(forbidden_response("Forbidden: Origin does not match Host"));
+            }
+        }
+    }
+
+    Ok(())
```
DaleSeo, line 244: Would it make sense to reject requests that are missing the `Host` header when `allowed_hosts` is non-empty?

TypeScript SDK's DNS rebinding guard rejects requests with missing headers as well: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/middleware/hostHeaderValidation.ts

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -48,6 +48,12 @@ pub struct StreamableHttpServerConfig {
     /// When this token is cancelled, all active sessions are terminated and
     /// the server stops accepting new requests.
     pub cancellation_token: CancellationToken,
+    /// Allowed hostnames for inbound `Host` / `Origin` validation.
+    ///
+    /// By default, Streamable HTTP servers only accept loopback hosts to
+    /// prevent DNS rebinding attacks against locally running servers. Public
+    /// deployments should override this list with their own hostnames.
+    pub allowed_hosts: Vec<String>,
```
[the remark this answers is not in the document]
DaleSeo: This is especially important given that PR #715 was recently merged to prepare for 1.0 stable release. `StreamableHttpServerConfig` may have been missed in that effort. Adding `#[non_exhaustive]` here would be consistent with that direction and prevent this class of breakage going forward.

DaleSeo (commented):
Thank you for tackling DNS rebinding protection, @jokemanfire. This is an important security hardening that the MCP Streamable HTTP transport was missing. 

I actually implemented a similar Host validation feature on the consumer side in apollographql/apollo-mcp-server#602, so I very much welcome this change. Once this lands in the SDK, I'd love to remove our application-layer middleware and rely on this built-in protection instead.

I have a few suggestions based on what I learned building the consumer-side implementation.

crates/rmcp/src/transport/streamable_http_server/tower.rs
jokemanfire: Well ,thanks ,I will take some time to resolve this.

crates/rmcp/src/transport/streamable_http_server/tower.rs
jokemanfire: Yes , It should be considerd.

crates/rmcp/src/transport/streamable_http_server/tower.rs
jokemanfire: Keep same with ts.

crates/rmcp/src/transport/streamable_http_server/tower.rs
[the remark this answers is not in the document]
jokemanfire: I think this security update should be introduced in 1.0 version , and I will add the #[non_exhaustive].

crates/rmcp/src/transport/streamable_http_server/tower.rs
jokemanfire: I consider CORS in other pr to enhance the sec.

DaleSeo (approved):
Thanks for addressing my feedback, @jokemanfire. Looks good!

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -130,6 +153,97 @@ fn validate_protocol_version_header(headers: &http::HeaderMap) -> Result<(), Box
...
+        .status(http::StatusCode::BAD_REQUEST)
+        .header(http::header::CONTENT_TYPE, "text/plain; charset=utf-8")
+        .body(body)
+        .expect("failed to build bad request response")
+}
+
+fn parse_host_header(headers: &HeaderMap) -> Result<NormalizedAuthority, BoxResponse> {
+    let Some(host) = headers.get(http::header::HOST) else {
+        return Err(bad_request_response("Bad Request: missing Host header"));
+    };
```
layou233, line 225: I got this error when sending requests via HTTP/2 with axum nesting service. H2 does not use `Host` header but use the `:authority` pseudo-header instead. Will this work? Or am I missing something? 🤔

DaleSeo: Sorry @layou233, I saw your comment late here. I've opened PR #827, which falls back to `request.uri().authority()` when the Host header is absent, allowing HTTP/2

PR #766, opened 2026-03-22
feat: add "theme" to Icon

glicht:
## Motivation and Context

Add the "theme" field to Icon struct as defined in the MCP Spec: https://modelcontextprotocol.io/specification/2025-11-25/schema#icon . Clients and Servers can use this field now.

## How Has This Been Tested?

Updated unit tests to include the theme field. All tests pass.

## Breaking Changes

No

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -895,7 +895,9 @@ impl Default for ClientInfo {
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[non_exhaustive]
 pub enum IconTheme {
+    /// Indicates the icon is designed to be used with a light background
     Light,
+    /// Indicates the icon is designed to be used with a dark background
     Dark,
 }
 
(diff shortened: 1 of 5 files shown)


crates/rmcp/src/model.rs
```diff
@@ -889,6 +889,15 @@ impl Default for ClientInfo {
     }
 }
 
+/// Icon themes supported by the MCP specification
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Copy)]
+#[serde(rename_all = "lowercase")] //match spec
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub enum IconTheme {
```
DaleSeo, line 897: The MCP spec could add more themes in the future (e.g., `high-contrast`). This would prevent downstream `match` arms from breaking when new variants are added.

```suggestion
#[non_exhaustive]
pub enum IconTheme {
```

crates/rmcp/src/model.rs
```diff
@@ -889,6 +889,15 @@ impl Default for ClientInfo {
     }
 }
 
+/// Icon themes supported by the MCP specification
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Copy)]
```
DaleSeo: Adding these would let consumers use it as a `HashMap` key or in `HashSet` without friction.

```suggestion
#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash, Copy)]
```

glicht:
> Thanks for adding this, @glicht! I have a couple of minor comments.

Thanks. I've committed your suggestions. Also improved the docs for the icon theme.

PR #774, opened 2026-03-25
feat(transport): add which_command for cross-platform executable resolution

### LINKED ISSUE #456: The abstract method uses which under the hood.
# 🪟 Windows Path Resolution Problem in `tokio::process::Command`

On **Windows**, launching a process like this:

```rust
use rmcp::{ServiceExt, transport::{TokioChildProcess, ConfigureCommandExt}};
use tokio::process::Command;

#[tokio::main]
    let client = ().serve(TokioChildProcess::new(Command::new("npx").configure(|cmd| {
        cmd.arg("-y").arg("@modelcontextprotocol/server-everything");
    }))?).await?;
    Ok(())
}
```

can **fail**, because:

* Unlike Linux or macOS, **Windows does not reliably resolve executables** via the `PATH` environment variable in `tokio::process::Command`.
* Many tools like `npx` are **`.cmd` shim scripts**, often stored in locations like:

  ```
  C:\Users\<User>\AppData\Roaming\npm\npx.cmd
  ```
* Without the **full executable path**, you’ll get runtime errors like:

  ```
  The system cannot find the file specified. (os error 2)
  ```

---

## Why Use `[which](https://docs.rs/which/latest/which/)`

The [`which`](https://docs.rs/which/latest/which/) crate resolves the **absolute path** to an executable:

* On Linux/macOS, it works like the native `which` command in a shell.
* On Windows, it handles `.cmd` / `.exe` resolution and searches `PATH` correctly.

By wrapping this logic in a **high-level abstraction**, your library can work **seamlessly across platforms** without requiring users to handle these quirks manually.

---

# 💡 Solution — A Builder That Hides `which` Internally

Below is a minimal `CmdBuilder` that:

1. Automatically resolves the executable path using `which` under the hood.
2. Provides a **fluent API** with `.arg(...)` chaining for adding arguments.
3. Offers `.configure(...)` for **low-level full control** over the `Command`.
4. Returns a ready-to-use `TokioChildProcess`.

```rust
use which::which;
use rmcp::{ServiceExt, transport::{TokioChildProcess, ConfigureCommandExt}};
use tokio::process::Command;
use std::error::Error;

struct CmdBuilder {
    command: Command,
}

impl CmdBuilder {
    /// Creates the builder and resolves the full program path (important on Windows).
    fn new(name: &str) -> Result<Self, Box<dyn Error>> {
        let path = which(name)?;
        Ok(Self { command: Command::new(path) })
    }

    /// Adds an argument (chainable).
    fn arg(mut self, arg: &str) -> Self {
        self.command.arg(arg);
        self
    }

    /// Allows full, low-level access to `Command`.
    fn configure<F>(mut self, f: F) -> Self
    where
        F: FnOnce(&mut Command),
    {
        f(&mut self.command);
        self
    }

    /// Finalizes and returns a `TokioChildProcess`.
    fn build(self) -> TokioChildProcess<Command> {
        TokioChildProcess::new(self.command)
    }
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    let client = ().serve(
        CmdBuilder::new("npx")?
            .arg("-y")
            .arg("@modelcontextprotocol/server-everything")
            .configure(|cmd| {
                // Optional: add environment variables or other settings
                cmd.env("MY_ENV_VAR", "123");
            })
            .build(),
    ).await?;

    Ok(())
}
```

---

## ⚠️ Disclaimer

This code is **provisional** and **not a production-ready solution**. It’s only meant to illustrate an idea for solving cross-platform executable path issues.

**Cheers and good luck !**

kakarot-dev:
## Summary

Adds a `which_command()` helper function that resolves executable paths via the [`which`](https://docs.rs/which) crate before constructing a `tokio::process::Command`. This fixes Windows failures where `.cmd` shim scripts (e.g. `npx.cmd`) are not found by `Command::new()` without a fully-qualified path.

**Changes:**
- Added `which_command(name) -> io::Result<Command>` to `child_process.rs`
- Added `which` v7 as an optional dependency, gated behind `transport-child-process` feature
- Re-exported `which_command` from `transport` module
- Added tests for known binary resolution and nonexistent binary error handling

**Usage:**
```rust
use rmcp::transport::{which_command, ConfigureCommandExt};

let cmd = which_command("npx")?
    .configure(|cmd| {
        cmd.arg("-y").arg("@modelcontextprotocol/server-everything");
    });
```

Closes #456

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -136,6 +136,9 @@ transport-child-process = [
   "transport-async-rw",
   "tokio/process",
   "dep:process-wrap",
+]
+which-command = [
+  "transport-child-process",
   "dep:which",
 ]
 transport-streamable-http-server = [
--- a/crates/rmcp/src/transport/child_process.rs
+++ b/crates/rmcp/src/transport/child_process.rs
@@ -242,7 +242,7 @@ impl ConfigureCommandExt for tokio::process::Command {
 ///
 /// # Example
 /// ```rust,no_run
-/// use rmcp::transport::child_process::{which_command, ConfigureCommandExt};
+/// use rmcp::transport::{which_command, ConfigureCommandExt};
 ///
 /// # fn example() -> std::io::Result<()> {
 /// let cmd = which_command("npx")?
@@ -252,6 +252,7 @@ impl ConfigureCommandExt for tokio::process::Command {
 /// # Ok(())
 /// # }
 /// ```
+#[cfg(feature = "which-command")]
 pub fn which_command(
     name: impl AsRef<std::ffi::OsStr>,
 ) -> std::io::Result<tokio::process::Command> {
@@ -260,6 +261,7 @@ pub fn which_command(
     Ok(tokio::process::Command::new(resolved))
 }
 
+#[cfg(feature = "which-command")]
 #[cfg(test)]
 mod tests_which {
     #[test]
--- a/crates/rmcp/src/transport.rs
+++ b/crates/rmcp/src/transport.rs
@@ -83,8 +83,10 @@ pub use worker::WorkerTransport;
 
 #[cfg(feature = "transport-child-process")]
 pub mod child_process;
+#[cfg(feature = "which-command")]
+pub use child_process::which_command;
 #[cfg(feature = "transport-child-process")]
-pub use child_process::{ConfigureCommandExt, TokioChildProcess, which_command};
+pub use child_process::{ConfigureCommandExt, TokioChildProcess};
 
 #[cfg(feature = "transport-io")]
 pub mod io;


crates/rmcp/Cargo.toml
```diff
@@ -133,6 +136,7 @@ transport-child-process = [
   "transport-async-rw",
   "tokio/process",
   "dep:process-wrap",
+  "dep:which",
```
DaleSeo, line 142: The `which` crate is now included for everyone using `transport-child-process`, even on Linux and macOS where `Command::new` already resolves `PATH` correctly. Since this is a library crate, it might be a good idea to make it a separate feature flag so users can opt in explicitly.
```toml
which-command = [
  "transport-child-process",
  "dep:which",
]
```

crates/rmcp/src/transport/child_process.rs
```diff
@@ -233,6 +233,54 @@ impl ConfigureCommandExt for tokio::process::Command {
...
+/// Resolve the absolute path to an executable using the system `PATH`,
+/// then return a [`tokio::process::Command`] pointing at it.
+///
+/// This is especially useful on Windows where `.cmd` / `.exe` shim scripts
+/// (e.g. `npx.cmd`) are not reliably found by [`tokio::process::Command`]
+/// without a fully-qualified path.
+///
+/// # Example
+/// ```rust,no_run
+/// use rmcp::transport::child_process::{which_command, ConfigureCommandExt};
```
DaleSeo: These are re-exported.

```suggestion
/// use rmcp::transport::{which_command, ConfigureCommandExt};
```

crates/rmcp/Cargo.toml
kakarot-dev: Good call — moved `which` behind a separate `which-command` feature flag so it's fully opt-in. Users on Linux/macOS won't pull it in unless they need it. Thanks for the review!

crates/rmcp/src/transport/child_process.rs
kakarot-dev: Fixed — updated to use the re-exported path.

Issue #330, opened 2025-07-23
HTTP 401 Unauthorized errors and connection drops when running MCP server behind Azure Load Balancer


louis030195:
Will add more context but this is what i got, wondering if this is problem on my infra side or could be anything related to the `rmcp` lib?:

### **Environment:**
- **MCP Server**: `terminator-mcp-agent@latest` (HTTP transport mode)
- **Command**: `npx -y terminator-mcp-agent@latest -t http -p 8080 --host 0.0.0.0`
- **Load Balancer**: Azure Standard Load Balancer (Round-robin distribution)
- **Client**: `terminator mcp run --url http://IP:8080/mcp`
- **Infrastructure**: 3x Windows Server 2022 VMs in Azure VM Scale Set

### **Problem Description:**
The MCP server works fine for individual connections but becomes unreliable when accessed through a load balancer, experiencing frequent **HTTP 401 Unauthorized** errors and connection drops during workflow execution.

### **Current Setup:**
```bash
# Health endpoint works consistently
curl http://x.x.x.x:8080/health
# Returns: {"status":"ok"}

# Load balancer configuration
- LoadDistribution: Default (Round-robin)
- IdleTimeout: 4 minutes  
- Protocol: TCP
- Frontend/Backend Port: 8080
```

#### **1. 401 Unauthorized During Execution:**
```
❌ MCP command error: Transport send error: Client error: 
HTTP status client error (401 Unauthorized) for url (http://x.x.x.x:8080/mcp)
```

#### **2. Connection Initialization Failures:**
```
rmcp::transport::worker: worker quit with fatal: Client error: 
HTTP status client error (404 Not Found) for url (http://x.x.x.x:8080/), 
when send initialize request
```

#### **3. Mid-Execution Drops:**
```
✅ MCP Connected: http://x.x.x.x:8080/mcp
✅ Protocol Initialized: MCP protocol handshake completed
⚡ Workflow Started: [executes some steps]
❌ Transport send error: 401 Unauthorized [randomly fails]
```

### **Observed Behaviors:**

1. **✅ Direct connections work perfectly**: Connecting directly to individual VMs is stable
2. **❌ Load balancer introduces instability**: Same code fails randomly when routed through LB
3. **🔄 Intermittent nature**: Sometimes works for several steps, then fails
4. **📊 Pattern**: Errors seem more frequent under concurrent load or longer workflows

#### **Tried Session Persistence:**
```terraform
[...]

PR #775, opened 2026-03-25
feat: optional session store (resumabillity support)

glicht:
## Motivation and Context

In horizontally-scaled deployments or after restarting a server, MCP clients use `Mcp-Session-Id` to resume sessions. When routed to a different server instance or after a restart, the session is unknown and clients get a 404, forcing re-initialization.

This PR adds an optional SessionStore trait that persists InitializeRequestParams after a successful handshake. When a request arrives at an instance without a matching in-memory session, the store is consulted and the session is transparently restored by replaying initialize. 

The `SessionManager` trait gains a `restore_session` method with a default no-op implementation. Custom session manager implementations can override it to integrate with their own logic if needed. The built-in `LocalSessionManager` has an implementation that re-creates the in-memory session worker. 

Additionally to provide indication to `ServerHandler` implementation that a call to initialize and on_initialized is as a result of a restore, a marker `SessionRestoreMarker` is added to the context.extensions so implementors can act appropriately when a session is restored. 

The feature is opt-in. Configurations without a session_store configured are unaffected.

Note: for full resumability and multi-server (behind a load balancer) support there is need to also implement an Event store so events aren't lost. This is being discussed at: #330 

## How Has This Been Tested?

Added a new integration test suite at: `test_streamable_http_session_store.rs` with three test scenarios.

## Breaking Changes

None. Existing logic remains if session_store is not configured (defaults to None). 

## Types of changes

- [ ] Documentation update

## Checklist

- [ x I have added or updated documentation as needed

## Additional context

Related to #330

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -78,19 +78,9 @@ pub struct StreamableHttpServerConfig {
     pub session_store: Option<Arc<dyn SessionStore>>,
 }
 
-impl std::fmt::Debug for StreamableHttpServerConfig {
+impl std::fmt::Debug for dyn SessionStore {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("StreamableHttpServerConfig")
-            .field("sse_keep_alive", &self.sse_keep_alive)
-            .field("sse_retry", &self.sse_retry)
-            .field("stateful_mode", &self.stateful_mode)
-            .field("json_response", &self.json_response)
-            .field("cancellation_token", &self.cancellation_token)
-            .field(
-                "session_store",
-                &self.session_store.as_ref().map(|_| "<SessionStore>"),
-            )
-            .finish()
+        f.write_str("<SessionStore>")
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -237,15 +289,249 @@ where
...
+                tokio::spawn(async move {
+                    if let Some(tx) = pending_restores.write().await.remove(&session_id) {
+                        let _ = tx.send(Some(result));
+                    }
+                });
+            }
+        };
+
+        // --- Step 3: load from external store ---
+        let state = match store.load(session_id.as_ref()).await {
```
rockdaboot, line 567: If store.load is cancelled (client connection closes/drops), how is `pending_restores` cleaned up?

glicht: @rockdaboot good point. I'll take a look how this can be  protected. Thanks for the feedback.

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -29,7 +34,7 @@ use crate::{
     },
 };
 
-#[derive(Debug, Clone)]
+#[derive(Clone)]
```
BrentGarnett:
```suggestion
#[derive(Debug, Clone)]
```

Would be be less risk to impl on `dyn SessionStore` only?

```
impl std::fmt::Debug for dyn SessionStore{
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str("<SessionStore>")
    }
}
```

glicht: I've implemented your suggestion. Thanks

crates/rmcp/src/transport/streamable_http_server/tower.rs
glicht: I've changed the code to use a guard.

glicht:
@alexhancock any feedback on this PR? Would really like to progress this forward.

alexhancock (commented):
In theory this will only be needed until June/July when https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575 is in the official spec, but the implementation looks good and will likely need to live longer than SEP-2575 is included official and used!

alexhancock (approved):

PR #778, opened 2026-03-26
feat(auth): add StoredCredentials::new() constructor

### LINKED ISSUE #777: StoredCredentials needs a constructor — #[non_exhaustive] without new() makes it unconstructable from external crates
`StoredCredentials` was marked `#[non_exhaustive]` in [#715](https://github.com/modelcontextprotocol/rust-sdk/pull/715) / [#768](https://github.com/modelcontextprotocol/rust-sdk/pull/768), but unlike `OAuthClientConfig` (which got a `new()` in the same PR), `StoredCredentials` doesn't have a constructor or `Default` impl. External crates that implement `CredentialStore` and need to construct `StoredCredentials` in their `save()` path don't have a compile-safe way to do so.

The workaround we landed on is a serde JSON roundtrip:

```rust
let credentials: StoredCredentials = serde_json::from_value(serde_json::json!({
    "client_id": client_id,
    "token_response": token_response,
    "granted_scopes": granted_scopes,
    "token_received_at": timestamp,
}))?;
```

This bypasses `#[non_exhaustive]` via `Deserialize` but turns what should be a compile-time error (if a new required field is added) into a silent runtime failure during OAuth.

We ran into this in [block/goose](https://github.com/block/goose) after bumping to rmcp 1.3.0 for Unix socket transport support ([#749](https://github.com/modelcontextprotocol/rust-sdk/pull/749)).

**Suggested fix:** A constructor matching the pattern used for `OAuthClientConfig`:

```rust
impl StoredCredentials {
    pub fn new(
        client_id: String,
        token_response: Option<OAuthTokenResponse>,
        granted_scopes: Vec<String>,
        token_received_at: Option<u64>,
    ) -> Self {
        Self {
            client_id,
            token_response,
            granted_scopes,
            token_received_at,
        }
    }
}
```

wpfleger96:
## Summary

- Add a `new()` constructor to `StoredCredentials`

`StoredCredentials` was marked `#[non_exhaustive]` in #715 / #768, but unlike other types in that sweep (e.g. `OAuthClientConfig`), it didn't get a constructor. External crates implementing `CredentialStore` that need to construct `StoredCredentials` in their `save()` path currently have to resort to a serde JSON roundtrip, which turns compile-time field mismatches into silent runtime failures.

Fixes #777

## Test plan

- `just check` passes
- `just test` passes — all existing tests unaffected (they construct `StoredCredentials` with struct literals inside the crate, which still works)

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -84,6 +84,23 @@ impl std::fmt::Debug for StoredCredentials {
     }
 }
 
+impl StoredCredentials {
+    /// Create a new `StoredCredentials` instance.
+    pub fn new(
+        client_id: String,
+        token_response: Option<OAuthTokenResponse>,
+        granted_scopes: Vec<String>,
+        token_received_at: Option<u64>,
+    ) -> Self {
+        Self {
+            client_id,
+            token_response,
+            granted_scopes,
+            token_received_at,
+        }
+    }
+}
+
 /// Trait for storing and retrieving OAuth2 credentials
 ///
 /// Implementations of this trait can provide custom storage backends


wpfleger96:
thank you @alexhancock !!

PR #780, opened 2026-03-27
fix: default session keep_alive to 5 minutes

DaleSeo:
## Motivation and Context

When an HTTP/2 connection silently drops (e.g., an Envoy sidecar sending RST_STREAM with NO_ERROR), the `LocalSessionManager` still holds the session handle, keeping the worker's event channel open. The session worker never detects the disconnect and runs indefinitely as a zombie. Over time these zombies accumulate and can block servers that iterate over sessions during notifications, eventually causing the server to become unresponsive.

This changes the default `SessionConfig::keep_alive` from `None` (infinite) to 5 minutes. After 5 minutes of inactivity, the session worker exits, the transport closes, and downstream servers can detect the peer as closed. Users can still set `keep_alive: None` to restore the old behavior if needed.

## How Has This Been Tested?

2. Verify the default is applied by checking the `SessionConfig::default()` value:

```rust
use rmcp::transport::streamable_http_server::session::local::SessionConfig;
let config = SessionConfig::default();
assert_eq!(config.keep_alive, Some(std::time::Duration::from_secs(300)));
```

3. To manually verify zombie cleanup, start an MCP server using `LocalSessionManager::default()`, initialize a session via curl, Ctrl+C the curl (creating a zombie), and wait 5 minutes. The server logs should show `"keep alive timeout after 300000ms"` and the session worker should exit.

## Breaking Changes

None. No API signatures changed.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -1069,19 +1069,30 @@ impl Worker for LocalSessionWorker {
 pub struct SessionConfig {
     /// the capacity of the channel for the session. Default is 16.
     pub channel_capacity: usize,
-    /// if set, the session will be closed after this duration of inactivity.
+    /// The session will be closed after this duration of inactivity.
+    ///
+    /// This serves as a safety net for cleaning up sessions whose HTTP
+    /// connections have silently dropped (e.g., due to an HTTP/2
+    /// `RST_STREAM`). Without a timeout, such sessions become zombies:
+    /// the session worker keeps running indefinitely because the session
+    /// handle's sender is still held in the session manager, preventing
+    /// the worker's event channel from closing.
+    ///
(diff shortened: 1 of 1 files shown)


alexhancock (approved):

synapsis1:
Was the intention that it would log as ERROR?  I am seeing:

`ERROR rmcp::transport::worker: worker quit with fatal: keep alive timeout after 300000ms, when poll next session event`

The good news is that I didn't realize this was happening with my servers (session worker zombies) and hadn't overridden the default of None, but isn't this more of a client error (the client left without saying goodbye)?  Perhaps this is more of a WARN, or maybe INFO (in the case the worker was terminated because the client dropped silently).  ERROR generates questions, and this is normal behavior.  

I didn't want to open up a new issue before understanding if this was the intent. 

There is a secondary error that shows up as well, that is simply because of the first:

`ERROR rmcp::transport::streamable_http_server::tower: Failed to close session <uuid>: Session error: Session service terminated`

PR #788, opened 2026-04-03
fix(server): remove initialized notification gate to support Streamable HTTP

### LINKED ISSUE #783: Server fatally rejects any request before initialized notification — breaks Streamable HTTP clients
**Describe the bug**

After sending the `InitializeResult`, the server enters a loop waiting for `notifications/initialized`. This loop only tolerates `logging/setLevel` and `ping` — any other message (e.g. `tools/list`, `resources/list`) causes a fatal `ServerInitializeError::ExpectedInitializedNotification` error that kills the entire session.

This is the problematic code in [`crates/rmcp/src/service/server.rs` lines 276-280](https://github.com/modelcontextprotocol/rust-sdk/blob/52c93e9/crates/rmcp/src/service/server.rs#L276-L280):

```rust
other => {
    return Err(ServerInitializeError::ExpectedInitializedNotification(
        Some(other),
    ));
}
```

PR #730 (fixing #704) added tolerance for `SetLevelRequest` and `PingRequest`, but any other request type still fatally terminates the session.

**To Reproduce**

1. Configured OpenClaw to run with mcp-remote. 
2. Client sends `initialize`, receives `InitializeResult`, then sends both `notifications/initialized` and `tools/list` as separate HTTP POSTs
3. If `tools/list` POST arrives before `notifications/initialized` POST (which is common — there's no ordering guarantee), the server crashes

**Error:**
```
Failed to serve session: expect initialized notification, but received: 
Some(Request(JsonRpcRequest { ... request: ListToolsRequest(...) }))
```

The session worker exits, the transport drops, and all subsequent requests return HTTP 500.

**Expected behavior**

The server should process requests arriving before `notifications/initialized` normally, rather than fatally rejecting them. The `initialized` notification should not gate request handling.

**Why this should be fixed**

**1. MCP spec uses SHOULD NOT, not MUST NOT**

From the [MCP Specification (2025-11-25), Lifecycle](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle):

> "The client **SHOULD NOT** send requests other than pings before the server has responded to the `initialize` request."
>
> "The server **SHOULD NOT** send requests other than pings and logging before receiving the `initialized` notification."

Per RFC 2119, SHOULD NOT means "there may exist valid reasons in particular circumstances when the behavior is acceptable." The spec constrains what each side **sends** — it says nothing about **rejecting** messages received during this phase.

**2. Streamable HTTP has no ordering guarantee**

From the [MCP Specification (2025-11-25), Transports](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports):

> "Every JSON-RPC message sent from the client **MUST** be a new HTTP POST request to the MCP endpoint."

Each POST is independent. Even if a well-behaved client sends `initialized` before `tools/list`, the HTTP layer may deliver them in any order.

**3. The TypeScript SDK (reference implementation) does not gate on `initialized`**

The official TypeScript SDK handles this differently — it has no initialization gate and processes all requests immediately after `initialize`:

- **Protocol layer** ([`protocol.ts` line 542](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core/src/shared/protocol.ts#L542)): `_onrequest()` dispatches immediately to the registered handler — no `_initialized` check exists anywhere in this file.

- **Server class** ([`server.ts` line 136](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/server.ts#L136)): The `notifications/initialized` handler merely fires an optional `oninitialized?.()` callback. There is no logic that rejects or queues requests arriving before `initialized`.

- **StreamableHTTP transport** ([`streamableHttp.ts` line 678](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/streamableHttp.ts#L678)): `_initialized` is set to `true` on the `initialize` **request** (not the `initialized` notification). `validateSession()` ([line 852](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/streamableHttp.ts#L852)) only rejects if `_initialized` is `false` — i.e., before the `initialize` request, not before `notifications/initialized`.

The sequence `initialize → tools/list → notifications/initialized` works perfectly in the TypeScript SDK. This was confirmed in [typescript-sdk#578](https://github.com/modelcontextprotocol/typescript-sdk/issues/578), which was closed with the clarification that responding to client requests before `initialized` does not violate the spec.

**Proposed fix**

Remove the `initialized` notification wait loop entirely. After sending `InitializeResult`, enter `serve_inner` immediately. The `initialized` notification is then handled as a regular notification by the service's normal dispatch — matching the TypeScript SDK behavior.

This is a ~40-line deletion with no new code needed, since `serve_inner` already handles all message types including notifications.

I have a working implementation and can open a PR if preferred.

anara123:
## Summary

- Remove the ~40-line `initialized` notification wait loop that fatally rejected any request arriving before `notifications/initialized`
- Enter `serve_inner` immediately after sending `InitializeResult`, letting the main service loop handle all messages including `initialized`
- Remove the now-unreachable `ExpectedInitializedNotification` error variant from `ServerInitializeError`

## Motivation

Streamable HTTP requires each JSON-RPC message as a separate POST request. The transport layer cannot guarantee delivery order, so `tools/list` can easily arrive before `notifications/initialized`, causing a fatal HTTP 500 error.

The MCP spec uses **SHOULD NOT** (RFC 2119), not **MUST NOT**, for pre-initialized messages — allowing exceptions in specific circumstances like HTTP's lack of ordering guarantees.

This aligns with the TypeScript SDK's behavior, which processes requests immediately after `initialize` without gating on `initialized` (validated in modelcontextprotocol/typescript-sdk#578).

## Test plan

- [x] Updated `server_init_set_level_response_is_empty_result` to handle notifications from `serve_inner` dispatch
- [x] Updated `server_init_succeeds_after_set_level_before_initialized` similarly
- [x] Replaced `server_init_rejects_unexpected_message_before_initialized` with `server_init_buffers_request_before_initialized` — verifies `tools/list` before `initialized` is processed successfully
- [x] Added `server_init_buffers_multiple_requests_before_initialized` — verifies multiple pre-init messages are all processed

Closes #783

### DIFF
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -53,6 +53,13 @@ pub enum ServerInitializeError {
     #[error("expect initialized request, but received: {0:?}")]
     ExpectedInitializeRequest(Option<ClientJsonRpcMessage>),
 
+    #[deprecated(
+        since = "1.4.0",
+        note = "The server no longer gates on the initialized notification. This variant is never constructed and will be removed in a future major release."
+    )]
+    #[error("expect initialized notification, but received: {0:?}")]
+    ExpectedInitializedNotification(Option<ClientJsonRpcMessage>),
+
     #[error("connection closed: {0}")]
     ConnectionClosed(String),
 
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -53,9 +53,6 @@ pub enum ServerInitializeError {
     #[error("expect initialized request, but received: {0:?}")]
     ExpectedInitializeRequest(Option<ClientJsonRpcMessage>),
 
-    #[error("expect initialized notification, but received: {0:?}")]
-    ExpectedInitializedNotification(Option<ClientJsonRpcMessage>),
-
     #[error("connection closed: {0}")]
     ConnectionClosed(String),
 
@@ -243,49 +240,12 @@ where
             ServerInitializeError::transport::<T>(error, "sending initialize response")
         })?;
 
-    // Wait for initialized notification. The MCP spec permits logging/setLevel and ping
-    // before initialized; VS Code sends setLevel immediately after the initialize response.
-    let notification = loop {
-        let msg = expect_next_message(&mut transport, "initialize notification").await?;
-        match msg {
-            ClientJsonRpcMessage::Notification(n)
-                if matches!(
-                    n.notification,
-                    ClientNotification::InitializedNotification(_)
-                ) =>
-            {
-                break n.notification;
-            }
-            ClientJsonRpcMessage::Request(req)
-                if matches!(
-                    req.request,
-                    ClientRequest::SetLevelRequest(_) | ClientRequest::PingRequest(_)
-                ) =>
-            {
-                transport
-                    .send(ServerJsonRpcMessage::response(
-                        ServerResult::EmptyResult(EmptyResult {}),
-                        req.id,
-                    ))
-                    .await
-                    .map_err(|error| {
-                        ServerInitializeError::transport::<T>(error, "sending pre-init response")
-                    })?;
-            }
-            other => {
-                return Err(ServerInitializeError::ExpectedInitializedNotification(
-                    Some(other),
-                ));
-            }
-        }
-    };
-    let context = NotificationContext {
-        meta: notification.get_meta().clone(),
-        extensions: notification.extensions().clone(),
-        peer: peer.clone(),
-    };
-    let _ = service.handle_notification(notification, context).await;
-    // Continue processing service
+    // Enter the main service loop immediately after sending InitializeResult.
+    // The initialized notification will be handled as a regular notification by serve_inner.
+    // This matches the TypeScript SDK behavior: no init gate, no waiting for initialized.
+    // Streamable HTTP has no ordering guarantee between POSTs, and the MCP spec uses
+    // SHOULD NOT (not MUST NOT) for pre-initialized messages, so any request arriving
+    // before initialized is processed normally.
     Ok(serve_inner(service, transport, peer, peer_rx, ct))
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ __pycache__/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+node_modules/
+.DS_Store
--- a/crates/rmcp/tests/test_server_initialization.rs
+++ b/crates/rmcp/tests/test_server_initialization.rs
@@ -6,7 +6,6 @@ use common::handlers::TestServer;
 use rmcp::{
     ServiceExt,
     model::{ClientJsonRpcMessage, ServerJsonRpcMessage, ServerResult},
-    service::ServerInitializeError,
     transport::{IntoTransport, Transport},
 };
 
@@ -54,7 +53,7 @@ async fn do_initialize(client: &mut impl Transport<rmcp::RoleClient>) {
     let _response = client.receive().await.unwrap();
 }
 
-// Server responds with EmptyResult to setLevel received before initialized.
+// Server handles setLevel sent before initialized notification (processed by serve_inner).
 #[tokio::test]
 async fn server_init_set_level_response_is_empty_result() {
     let (server_transport, client_transport) = tokio::io::duplex(4096);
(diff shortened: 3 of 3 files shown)


anara123:
@DaleSeo @alexhancock can anyone review this please.

crates/rmcp/src/service/server.rs
```diff
@@ -53,9 +53,6 @@ pub enum ServerInitializeError {
     #[error("expect initialized request, but received: {0:?}")]
     ExpectedInitializeRequest(Option<ClientJsonRpcMessage>),
 
-    #[error("expect initialized notification, but received: {0:?}")]
-    ExpectedInitializedNotification(Option<ClientJsonRpcMessage>),
```
DaleSeo, line 57: Since `ServerInitializeError` is public, removing its variant is a breaking change. Since the behavior changed intentionally, I suggest keeping this variant as deprecated. This way, we can maintain compatibility and avoid bumping the major version in the next release.

Here's the output from the semver checks:

```text
$ cargo semver-checks check-release
    Building rmcp v1.3.0 (current)
       Built [  17.264s] (current)
     Parsing rmcp v1.3.0 (current)
      Parsed [   0.069s] (current)
    Building rmcp v1.3.0 (baseline)
       Built [   3.207s] (baseline)
     Parsing rmcp v1.3.0 (baseline)
      Parsed [   0.066s] (baseline)
    Checking rmcp v1.3.0 -> v1.3.0 (no change; assume patch)
     Checked [   0.082s] 221 checks: 220 pass, 1 fail, 0 warn, 24 skip

--- failure enum_variant_missing: pub enum variant removed or renamed ---

Description:
A publicly-visible enum has at least one variant that is no longer available under its prior name. It may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.46.0/src/lints/enum_variant_missing.ron

Failed in:
  variant ServerInitializeError::ExpectedInitializedNotification, previously in file /private/var/folders/v2/5j2zr9qs7c76ph0yxn53d_9r0000gp/T/cursor-sandbox-cache/ef50fef5e15a857ee43162e39176828f/cargo-target/semver-checks/git-8e22aa2de28df5a285eed87c11cd89bf15fa90d3/257a621cff1beb7eafa5cd4a260f415999422844/crates/rmcp/src/service/server.rs:57

     Summary semver requires new major version: 1 major and 0 minor checks failed
    Finished [  22.930s] rmcp
```

DaleSeo (commented):
Thanks for the clear problem statement and rationale, @anara123! The lifecycle direction for Streamable HTTP makes sense and aligns with the spec language. I have one suggestion about backward compatibility.

crates/rmcp/src/service/server.rs
anara123: Thanks for the review, good catch. I reverted this variant.

DaleSeo: Thanks for addressing this, @anara123! Now I see it will bump the minor version. 👍

```text
$ cargo semver-checks check-release
    Building rmcp v1.3.0 (current)
       Built [   3.137s] (current)
     Parsing rmcp v1.3.0 (current)
      Parsed [   0.072s] (current)
    Building rmcp v1.3.0 (baseline)
       Built [   2.607s] (baseline)
     Parsing rmcp v1.3.0 (baseline)
      Parsed [   0.067s] (baseline)
    Checking rmcp v1.3.0 -> v1.3.0 (no change; assume patch)
     Checked [   0.074s] 221 checks: 220 pass, 1 fail, 0 warn, 24 skip

--- failure enum_variant_marked_deprecated: enum variant #[deprecated] added ---

Description:
An enum variant is now #[deprecated]. Downstream crates will get a compiler warning when using this variant.
        ref: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.46.0/src/lints/enum_variant_marked_deprecated.ron

Failed in:
  variant rmcp::service::ServerInitializeError::ExpectedInitializedNotification in /Users/dale.seo/.cursor/worktrees/rust-sdk/review-pr-788-qd1/crates/rmcp/src/service/server.rs:61

     Summary semver requires new minor version: 0 major and 1 minor checks failed
    Finished [   9.655s] rmcp
```

DaleSeo (approved):

DaleSeo:
@anara123 We're having an issue with the release-plz checks in main. I'll fix that first before merging this.

PR #790, opened 2026-04-06
fix(http): drain SSE stream for connection reuse

### LINKED ISSUE #789: StreamableHttpClientTransport has 10x higher latency on subsequent calls vs Python SDK
**Describe the bug**

`StreamableHttpClientTransport` in `rmcp v1.3.0` exhibits significantly higher latency on subsequent tool calls compared to the Python MCP SDK client. The first call is fast (~2ms), but every subsequent call takes ~42ms — roughly 10x slower than Python's subsequent calls (~3ms). The degradation is consistent and repro across all subsequent calls.

| Client | Call 1 | Call 2 | Call 3 | Call 4 | Call 5 |
|--------|--------|--------|--------|--------|--------|
| **Rust (rmcp)** | 1.7ms | 41.2ms | 42.0ms | 42.9ms | 41.9ms |
| **Python** | 12.0ms | 3.0ms | 2.9ms | 3.3ms | 3.2ms |

**To Reproduce**

1. Start the `servers_counter_streamhttp` server from `rmcp v1.3.0`.
2. Run the Rust client:
   ```rust
   use rmcp::transport::StreamableHttpClientTransport;
   use rmcp::{ServiceExt, model::*};
   use std::time::Instant;

   #[tokio::main]
       let transport = StreamableHttpClientTransport::from_uri("http://localhost:8000/mcp");
       let client = ClientInfo::default().serve(transport).await?;

       for i in 1..=5 {
           let start = Instant::now();
           let _res = client.call_tool(CallToolRequestParams::new("say_hello")).await.unwrap();
           println!("Call {}: {:.1}ms", i, start.elapsed().as_secs_f32() * 1000.0);
       }
       client.cancel().await?;
       Ok(())
   }
   ```
3. Run the equivalent Python client:
   ```python
   import asyncio, time
   from mcp import ClientSession
   from mcp.client.streamable_http import streamablehttp_client

   async def main():
       async with streamablehttp_client(url="http://localhost:8000/mcp") as (read, write, _):
           async with ClientSession(read, write) as session:
               await session.initialize()
               for i in range(1, 6):
                   start = time.perf_counter()
                   await session.call_tool("say_hello", arguments={})
                   print(f"Call {i}: {(time.perf_counter() - start) * 1000:.1f}ms")

   asyncio.run(main())
   ```

**Expected behavior**

Subsequent tool calls should have similar or lower latency than the first call, as the session is already established. The Python client demonstrates this pattern (12ms → 3ms). The Rust client should not regress to ~42ms on every call after the first.

**Logs**

Rust client output:
```
Call 1: 1.7ms
Call 2: 41.2ms
Call 3: 42.0ms
Call 4: 42.9ms
Call 5: 41.9ms
```

Python client output:
```
Call 1: 12.0ms
Call 2: 3.0ms
Call 3: 2.9ms
Call 4: 3.3ms
Call 5: 3.2ms
Session termination failed: 202
```

**Additional context**

- `rmcp` version: `1.3.0`
- Rust edition: `2024`
- Server: `servers_counter_streamhttp` from `rmcp v1.3.0` samples
- The issue is consistent across all post-initialization calls (not just a one-time spike)
- Fedora 43, x86_64

DaleSeo:
Fixes #789

## Motivation and Context

Subsequent tool calls via `StreamableHttpClientTransport` took ~42ms each on Linux, roughly 10x slower than the Python SDK. The root cause is twofold: the server kept per-request SSE channels open after sending a response, and the client dropped POST SSE response bodies without consuming them. This left stale connections in the pool that reqwest would stall on (~40ms TCP Delayed ACK) before falling back to a fresh connection.

This PR fixes both sides. On the server, request-wise channels now close immediately after sending a Response or Error, and the associated resource_router entries are cleaned up to prevent leaks. On the client, per-request POST SSE streams use a lightweight `raw_sse_to_jsonrpc` adapter instead of `SseAutoReconnectStream`, and `execute_sse_stream` drains remaining bytes after receiving a response so the connection returns to the pool cleanly. The default reqwest client also sets `pool_max_idle_per_host(0)` because TCP Delayed ACK on Linux still prevents reliable connection reuse even with the drain. Users who provide a custom client via `with_client()` are unaffected. Also fixed `is_response` to match `Error` in addition to `Response`, since JSON-RPC errors are equally terminal for request-wise streams.

## How Has This Been Tested?

1. Start the example server in one terminal:

```
cargo run --release --manifest-path examples/servers/Cargo.toml \
  --example counter_hyper_streamable_http
```

2. In another terminal, create a small benchmark client. Save this as `bench.rs` somewhere outside the workspace (e.g. `/tmp/bench/src/main.rs`) with a `Cargo.toml` pointing at the local rmcp crate:

```toml
[package]
name = "bench"
version = "0.1.0"
edition = "2024"

[dependencies]
rmcp = { path = "<path-to-this-repo>/crates/rmcp", features = ["client", "transport-streamable-http-client", "transport-streamable-http-client-reqwest"] }
tokio = { version = "1", features = ["full"] }
```

```rust
use rmcp::transport::StreamableHttpClientTransport;
use rmcp::{ServiceExt, model::*};
use std::time::Instant;

#[tokio::main]
    let transport = StreamableHttpClientTransport::from_uri("http://[::1]:8080/");
    let client = ClientInfo::default().serve(transport).await?;
    for i in 1..=5 {
        let start = Instant::now();
        let _res = client.call_tool(CallToolRequestParams::new("say_hello")).await?;
        println!("Call {}: {:.1}ms", i, start.elapsed().as_secs_f64() * 1000.0);
    }
    client.cancel().await?;
    Ok(())
}
```

3. Run it and verify calls 2-5 complete under 1ms:

```
cargo run --release
```

On Linux before this fix, calls 2-5 consistently show ~42ms. After the fix, all calls should be under 1ms. The issue does not reproduce on macOS due to different TCP Delayed ACK behavior.

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -339,3 +339,16 @@ required-features = [
 ]
 path = "tests/test_streamable_http_stale_session.rs"
 
+[[test]]
+name = "test_streamable_http_connection_reuse"
+required-features = [
+  "server",
+  "client",
+  "macros",
+  "schemars",
+  "transport-streamable-http-server",
+  "transport-streamable-http-client",
+  "transport-streamable-http-client-reqwest",
+]
+path = "tests/test_streamable_http_connection_reuse.rs"
+
--- a/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
@@ -262,7 +262,7 @@ impl StreamableHttpClientTransport<reqwest::Client> {
     /// This method requires the `transport-streamable-http-client-reqwest` feature.
     pub fn from_uri(uri: impl Into<Arc<str>>) -> Self {
         StreamableHttpClientTransport::with_client(
-            reqwest::Client::default(),
+            Self::default_http_client(),
             StreamableHttpClientTransportConfig {
                 uri: uri.into(),
                 auth_header: None,
@@ -277,7 +277,21 @@ impl StreamableHttpClientTransport<reqwest::Client> {
     ///
     /// * `config` - The config to use with this transport
     pub fn from_config(config: StreamableHttpClientTransportConfig) -> Self {
-        StreamableHttpClientTransport::with_client(reqwest::Client::default(), config)
+        StreamableHttpClientTransport::with_client(Self::default_http_client(), config)
+    }
+
+    /// Build the default reqwest client for this transport.
+    ///
+    /// Disables idle connection pooling because POST SSE response bodies
+    /// cannot be fully consumed before the next request starts, causing
+    /// reqwest to stall on pool checkout (~40 ms).  The long-lived GET
+    /// SSE stream already holds its own connection, so there is no
+    /// benefit to pooling POST connections with HTTP/1.1.
+    fn default_http_client() -> reqwest::Client {
+        reqwest::Client::builder()
+            .pool_max_idle_per_host(0)
+            .build()
+            .expect("failed to build default reqwest client")
     }
 }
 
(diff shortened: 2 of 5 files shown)


sco3:
[logs_64271381458.zip](https://github.com/user-attachments/files/26655414/logs_64271381458.zip)
Works as expected on macos-14 and ubuntu-22.04, one can check the logs from the workflow.

alexhancock (approved):

PR #803, opened 2026-04-10
 fix(macros): respect `local` feature in `#[prompt]` macro — omit `+ Send` bound

### LINKED ISSUE #801: #[prompt]` macro doesn't respect `local` feature — unconditionally emits `+ Send` unlike `#[tool]
Summary

The `#[tool]` macro correctly omits `+ Send` on generated handler futures when the `local` crate feature is enabled or `#[tool(local)]` is used. The `#[prompt]` macro does not — it unconditionally emits `+ Send` on all generated futures, making it impossible to use `!Send` types in prompt handlers even with the `local` feature.

## Context

The `local` feature was added in #740 to support `!Send` tool handlers (e.g., single-threaded platform APIs like Apple's EventKit on macOS). The `#[tool]` macro was updated to check for the feature:

**`rmcp-macros/src/tool.rs:339-341`:**
```rust
// 2. make return type: `std::pin::Pin<Box<dyn std::future::Future<Output = #ReturnType> + Send + '_>>`
//    (omit `+ Send` when the `local` crate feature is active or `#[tool(local)]` is used)
let omit_send = cfg!(feature = "local") || attribute.local;
```

The `#[prompt]` macro was not updated and unconditionally emits `+ Send`:

**`rmcp-macros/src/prompt.rs:141`:**
```rust
quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = ()> + Send + #lt>> }
```

**`rmcp-macros/src/prompt.rs:144`:**
```rust
quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = #ty> + Send + #lt>> }
```

## Reproduction

Enable the `local` feature and define a server struct holding a `!Send` type with both tool and prompt handlers:

```rust
use rmcp::*;
use std::rc::Rc; // !Send

struct MyServer {
    data: Rc<String>, // !Send
}

#[tool_router]
impl MyServer {
    #[tool(name = "greet")]
    async fn greet(&self) -> Result<CallToolResult, McpError> {
        // ✅ Compiles with `local` feature — future is not required to be Send
        Ok(CallToolResult::text(format!("Hello: {}", self.data)))
    }
}

#[prompt_router]
impl MyServer {
    #[prompt(name = "greeting")]
    async fn greeting(&self) -> Result<GetPromptResult, McpError> {
        // ❌ Fails to compile — future is required to be Send because
        // #[prompt] unconditionally emits `+ Send`
        Ok(GetPromptResult::new(vec![PromptMessage::new_text(
            PromptMessageRole::User,
            format!("Greeting: {}", self.data),
        )]))
    }
}
```

## Speculative Fix

Apply the same `omit_send` logic from `tool.rs` to `prompt.rs`. In `rmcp-macros/src/prompt.rs`, around line 130-145 where the return type is generated:

```rust
// Add this (matching tool.rs:339-341)
let omit_send = cfg!(feature = "local");

// Then change lines 141 and 144 from:
quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = ()> + Send + #lt>> }
quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = #ty> + Send + #lt>> }

// To:
if omit_send {
    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = ()> + #lt>> }
    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = #ty> + #lt>> }
} else {
    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = ()> + Send + #lt>> }
    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = #ty> + Send + #lt>> }
}
```

Optionally, also support the per-handler `#[prompt(local)]` attribute for parity with `#[tool(local)]`.

## Use Case

[eventkit-rs](https://crates.io/crates/eventkit-rs) is an MCP server wrapping Apple's EventKit framework. The underlying Objective-C managers (`EKEventStore`, `EKReminderStore`) are `!Send`. Currently every tool and prompt call must allocate a fresh manager because the `local` feature can't be used — the `#[prompt]` handlers reference `&self` and the macro forces `Send` on the future. With this fix, managers could be stored directly on the server struct and reused across calls.

WeekendSuperhero:
# Motivation and Context

Fixes #801

The `#[tool]` macro correctly omits `+ Send` on generated handler futures when the `local` crate feature is enabled (tool.rs:341). The `#[prompt]` macro was not updated when `local` support was added in #740 — it unconditionally emits `+ Send` on all generated futures (prompt.rs:141,144).

This makes it impossible to use `!Send` types in prompt handlers even with the `local` feature enabled, despite tool handlers working correctly. The real-world use case is MCP servers wrapping platform APIs with thread-affine types (e.g., Apple EventKit's `EKEventStore`).

## How Has This Been Tested?

- Existing tests pass locally (`cargo test --workspace`)
- Verified that a server struct holding `Rc<String>` (`!Send`) with both `#[tool]` and `#[prompt]` handlers compiles with `features = ["local"]` after this change (previously only `#[tool]` compiled)

## Breaking Changes

None. This only changes behavior when the `local` feature is explicitly enabled. Default behavior (without `local`) is unchanged — `+ Send` is still emitted.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix mirrors the existing pattern in `tool.rs:339-341`:

```rust
let omit_send = cfg!(feature = "local") || attribute.local;
```

Applied the same `cfg!(feature = "local")` check to `prompt.rs` for both `ReturnType::Default` and `ReturnType::Type` match arms.

### DIFF
--- a/crates/rmcp-macros/src/prompt.rs
+++ b/crates/rmcp-macros/src/prompt.rs
@@ -20,6 +20,9 @@ pub struct PromptAttribute {
     pub icons: Option<Expr>,
     /// Optional metadata for the prompt
     pub meta: Option<Expr>,
+    /// When true, the generated future will not require `Send`. Useful for `!Send` handlers
+    /// (e.g. single-threaded database connections). Also enabled globally by the `local` crate feature.
+    pub local: bool,
 }
 
 pub struct ResolvedPromptAttribute {
@@ -78,6 +81,7 @@ pub fn prompt(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream>
     };
     let mut fn_item = syn::parse2::<ImplItemFn>(input.clone())?;
     let fn_ident = &fn_item.sig.ident;
+    let omit_send = cfg!(feature = "local") || attribute.local;
 
     let prompt_attr_fn_ident = format_ident!("{}_prompt_attr", fn_ident);
 
@@ -124,6 +128,7 @@ pub fn prompt(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream>
     if fn_item.sig.asyncness.is_some() {
         // 1. remove asyncness from sig
         // 2. make return type: `std::pin::Pin<Box<dyn std::future::Future<Output = #ReturnType> + Send + '_>>`
+        //    (omit `+ Send` when the `local` crate feature is active or `#[prompt(local)]` is used)
         // 3. make body: { Box::pin(async move { #body }) }
         let new_output = syn::parse2::<ReturnType>({
             let mut lt = quote! { 'static };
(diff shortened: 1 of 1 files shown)


crates/rmcp-macros/src/prompt.rs
```diff
@@ -138,10 +138,18 @@ pub fn prompt(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream>
             }
             match &fn_item.sig.output {
                 syn::ReturnType::Default => {
-                    quote! { -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = ()> + Send + #lt>> }
+                    if omit_send {
```
DaleSeo, line 146: `omit_send` isn't defined, so it won't compile!

WeekendSuperhero: resolved. I added the bool along with a test.

DaleSeo (approved):

PR #809, opened 2026-04-15
feat(router): support runtime disabling of tools

### LINKED ISSUE #477: Custom disabling of specific tools
Is there an easy workflow for disabling certain tools when the MCP server starts?

Currently the best way I've been able to figure this out is by manually implementing the `ServerHandler` for my server however it would be nice to just be able to use the `#[tool_handler]` macro and maybe add in my custom disable logic within the `get_info` function which needs to be manually implemented (just an idea). 

Maybe I'm missing something and this is already possible. Thanks for the great work!

lutz-grex:
Add methods to disable/enable tools at runtime.
Disabled tools are hidden from listing, lookup,
and execution, including in composed routers.

Closes #477

## Motivation and Context

   Users of the `#[tool]` + `#[tool_handler]` macro system had no way to disable specific tools at runtime. The only workaround was manually implementing `ServerHandler`, losing all macro convenience, or using `remove_route` which permanently destroys the tool.
   This change adds reversible disable/enable support directly to `ToolRouter`, composing naturally with the existing macro workflow:

   ```rust
   let mut router = Self::tool_router();
   router.disable_route("dangerous_tool");
   // later...
   router.enable_route("dangerous_tool");
   ```

   ## How Has This Been Tested?

   - 9 new unit tests in `test_tool_routers.rs` covering:
     disable, enable, builder pattern, merge preservation,
     remove cleanup, pre-disable before add, iterator
     filtering, and full invisibility across all query methods
   - All 261 existing tests pass (`cargo test -p rmcp`)
   - Before/after validation: confirmed `disable_route` and
     `enable_route` do not compile on the baseline and work
     correctly after the change
   - `cargo clippy --all-targets --all-features` clean
   - `cargo +nightly fmt --all` clean

   ## Breaking Changes

   `has_route` now returns `false` for disabled tools (previously
   it only checked map membership). Code relying on `has_route`
   to test structural presence regardless of enabled state should
   use `is_disabled` instead. The `ToolRouter` struct is
   `#[non_exhaustive]`, so adding the private `disabled` field is
   not a breaking change.

   ## Types of changes

   - [ ] Documentation update

   ## Checklist

   ## Additional context

   Three files changed:
   - `crates/rmcp/src/handler/server/router/tool.rs` — core
     feature: `disabled` HashSet field, disable/enable/is_disabled
     /with_disabled methods, updated list_all/call/get/has_route
     /merge/remove_route/IntoIterator
   - `crates/rmcp/src/handler/server/router.rs` — one-line fix:
     `transparent_when_not_found` dispatch now checks `is_disabled`
     so disabled tools cannot leak through to the inner service
   - `crates/rmcp/tests/test_tool_routers.rs` — 9 new tests

### DIFF
--- a/crates/rmcp/src/handler/server/router/tool.rs
+++ b/crates/rmcp/src/handler/server/router/tool.rs
@@ -298,7 +298,6 @@ where
         self
     }
 }
-#[derive(Debug)]
 #[non_exhaustive]
 pub struct ToolRouter<S> {
     #[allow(clippy::type_complexity)]
@@ -307,6 +306,22 @@ pub struct ToolRouter<S> {
     pub transparent_when_not_found: bool,
 
     disabled: std::collections::HashSet<Cow<'static, str>>,
+
+    notifier: Option<Arc<dyn Fn() + Send + Sync>>,
+}
+
+impl<S> std::fmt::Debug for ToolRouter<S> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ToolRouter")
+            .field("map", &self.map)
+            .field(
+                "transparent_when_not_found",
+                &self.transparent_when_not_found,
+            )
+            .field("disabled", &self.disabled)
+            .field("notifier", &self.notifier.as_ref().map(|_| "..."))
+            .finish()
+    }
 }
 
 impl<S> Default for ToolRouter<S> {
@@ -315,15 +330,18 @@ impl<S> Default for ToolRouter<S> {
             map: std::collections::HashMap::new(),
             transparent_when_not_found: false,
             disabled: std::collections::HashSet::new(),
+            notifier: None,
         }
     }
 }
+
 impl<S> Clone for ToolRouter<S> {
     fn clone(&self) -> Self {
         Self {
             map: self.map.clone(),
             transparent_when_not_found: self.transparent_when_not_found,
             disabled: self.disabled.clone(),
+            notifier: self.notifier.clone(),
         }
     }
 }
@@ -346,11 +364,7 @@ where
     S: MaybeSend + 'static,
 {
     pub fn new() -> Self {
-        Self {
-            map: std::collections::HashMap::new(),
-            transparent_when_not_found: false,
-            disabled: std::collections::HashSet::new(),
-        }
+        Self::default()
     }
     pub fn with_route<R, A>(mut self, route: R) -> Self
     where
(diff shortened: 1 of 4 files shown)


crates/rmcp/src/handler/server/router/tool.rs
```diff
@@ -394,24 +403,64 @@ where
...
+
+    /// Re-enable a previously disabled tool.
+    pub fn enable_route(&mut self, name: &str) {
+        self.disabled.remove(name);
+    }
+
+    /// Returns `true` if the tool exists in the router but is currently
+    /// disabled.
+    pub fn is_disabled(&self, name: &str) -> bool {
+        self.map.contains_key(name) && self.disabled.contains(name)
```
DaleSeo, line 472: What happens if a tool is disabled before it's added? Is this supposed to return false?

crates/rmcp/src/handler/server/router/tool.rs
```diff
@@ -394,24 +403,64 @@ where
...
+
+    /// Disable a tool by name so it is hidden from `list_all`, `get`, and
+    /// rejected by `call`. The tool remains in the router and can be
+    /// re-enabled later with [`enable_route`](Self::enable_route).
+    ///
+    /// The name is recorded even if no matching route exists yet, so routes
+    /// added later (via [`add_route`](Self::add_route) or
+    /// [`merge`](Self::merge)) will inherit the disabled state.
+    pub fn disable_route(&mut self, name: &str) {
+        self.disabled.insert(Cow::Owned(name.to_owned()));
```
DaleSeo: `with_disabled` accepts `impl Into<Cow<'static, str>>`, but `disable_route` accepts `&str`. Can we align these two APIs so that callers don't have to choose between builder and mutable styles based on how memory is allocated?

DaleSeo (commented):
@lutz-grex How should clients learn about runtime disable/enable? I'm worried that the `list_tools` results will go stale until the client re-lists. I believe the MCP spec has `notifications/tools/list_changed` for this.

https://modelcontextprotocol.io/specification/2025-11-25/server/tools#list-changed-notification

crates/rmcp/src/handler/server/router/tool.rs
lutz-grex: Addressed - disable_route now records the name even without a matching route. Routes added later via add_route or merge inherit the disabled state. is_disabled() returns false until the route is registered. Added tests for this lifecycle.

crates/rmcp/src/handler/server/router/tool.rs
lutz-grex: Aligned - disable_route now takes impl Into<Cow<'static, str>>, matching with_disabled. Both return bool indicating whether the set changed.

DaleSeo:
Thanks for addressing my inline comments, @lutz-grex but I think you might've missed my concern in https://github.com/modelcontextprotocol/rust-sdk/pull/809#pullrequestreview-4125012020 I'm curious about your thoughts on the `notifications/tools/list_changed` side.

lutz-grex:
Yes, I overlooked these - good catch. I’ve now added automatic tools/list_changed notifications.

When using Router, this is handled transparently: Router::new() registers a deferred notifier that activates after InitializedNotification. From that point on, enable_route and disable_route emit a notification whenever the visible tool set actually changes. The router also advertises tools.listChanged: true in its capabilities.

crates/rmcp/src/handler/server/router/tool.rs
```diff
@@ -394,24 +417,135 @@ where
...
+    /// The name is recorded even if no matching route exists yet, so routes
+    /// added later will inherit the disabled state.
+    pub fn disable_route(&mut self, name: impl Into<Cow<'static, str>>) -> bool {
+        let name = name.into();
+        let was_visible = self.map.contains_key(&name) && !self.disabled.contains(&name);
+        let newly_disabled = self.disabled.insert(name.clone());
+        if was_visible && newly_disabled {
+            self.notify_if_visible(&name);
+        }
+        newly_disabled
```
DaleSeo: For simplicity

```suggestion
        let name = name.into();
        let was_visible = self.map.contains_key(&name) && !self.disabled.contains(&name);
        if was_visible {
            self.notify_if_visible(&name);
        }
        self.disabled.insert(name)
```

crates/rmcp/src/handler/server/router/tool.rs
```diff
@@ -394,24 +417,135 @@ where
...
+    }
+
+    /// Re-enable a previously disabled tool. Returns `true` if the name
+    /// was in the disabled set.
+    pub fn enable_route(&mut self, name: &str) -> bool {
```
DaleSeo, line 460: How about we also align this API with the other two?

crates/rmcp/src/handler/server/router/tool.rs
```diff
@@ -394,24 +417,135 @@ where
...
+                });
+            }
+        };
+        (notifier, peer_slot)
+    }
+
+    fn notify_if_visible(&self, name: &str) {
+        if self.map.contains_key(name) {
+            if let Some(notifier) = &self.notifier {
+                (notifier)();
```
DaleSeo: auto-deref

```suggestion
                notifier();
```

DaleSeo (dismissed):
Thanks, @lutz-grex! I just have minor suggestions.

crates/rmcp/src/handler/server/router/tool.rs
lutz-grex: Thanks for the note. The split is intentional: disable_route and with_disabled take impl Into<Cow<'static, str>> because they store the route name in the disabled set, whereas enable_route, remove_route, and has_route only take &str since they don’t store it. I find that this distinction makes it clearer where ownership is actually required, so I’d prefer to keep it as is.

DaleSeo (approved):

PR #811, opened 2026-04-17
fix: add init_timeout for streamable-http sessions

### LINKED ISSUE #808: Sessions stuck in pre-init state have no timeout, leak indefinitely
**Describe the bug**
rmcp 1.4.0
Sessions created by `LocalSessionManager` that never receive an initialize request remain alive indefinitely. The `session_config.keep_alive` timeout is only enforced inside `LocalSessionWorker::run()` after initialization - sessions waiting in the `get initialize request` phase have no timeout and are only cleaned up on server shutdown. In production, any HTTP POST to `/mcp` that creates a session but never follows up with initialize (e.g. a health probe, a load balancer preflight, a client that disconnects immediately, or a test harness bug) causes a permanent session leak. Over time this exhausts memory.

**To Reproduce**
1. Start a `StreamableHttpService` with `LocalSessionManager`.
2. Send an HTTP POST to `/mcp` with a valid JSON-RPC request that triggers session creation (e.g. `initialize`), but drop the TCP connection before sending the follow-up initialized notification - or send a non-initialize request that still creates a session
3. Observe that the session remains in memory indefinitely
4. Only server shutdown cleans it up

**Expected behavior**
Sessions waiting for initialize should have a configurable timeout (e.g. init_timeout, defaulting to 30-60 seconds). If no initialize request arrives within that window, the session should be terminated.

**Logs**
Server running with `session_config.keep_alive = 1200s` (20 min). Four sessions were created at 10:54:10 that never received `initialize`. They remained alive for 48 minutes until the server was shut down at 11:42:22, at which point they terminated with:
```log
ERROR rmcp::transport::worker: worker quit with fatal: transport terminated, when get initialize request
```
Meanwhile, sessions that did initialize were correctly reaped after 20 minutes of inactivity.

**Additional context**
Add a timeout in the pre-init phase of LocalSessionWorker, something like:
```rust
let init_timeout = self.session_config.init_timeout.unwrap_or(Duration::from_secs(60));
tokio::select! {
    req = self.wait_for_initialize() => { /* proceed */ },
    _ = tokio::time::sleep(init_timeout) => {
            LocalSessionWorkerError::InitTimeout(init_timeout),
            "get initialize request"
        ))
    }
}
```

DaleSeo:
Fixes #808. 

## Motivation and Context

The `LocalSessionWorker::run()` function waited for the first `initialize` request using just `self.event_rx.recv().await` without a timeout. Any HTTP POST that created a session but didn’t follow up could keep a worker task running indefinitely. The only way to free it was through a server shutdown, which eventually led to memory exhaustion.

The post-init loop already had a `keep_alive` timeout in place, but there was an issue with the pre-init path that created an asymmetric gap.

## How Has This Been Tested?

Added two integration tests

## Breaking Changes

No API breaks. Both `SessionConfig` and `LocalSessionWorkerError` are `#[non_exhaustive]`.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -916,6 +916,8 @@ pub enum LocalSessionWorkerError {
     FailToHandleMessage(SessionError),
     #[error("keep alive timeout after {}ms", _0.as_millis())]
     KeepAliveTimeout(Duration),
+    #[error("init timeout after {}ms", _0.as_millis())]
+    InitTimeout(Duration),
     #[error("Transport closed")]
     TransportClosed,
     #[error("Tokio join error {0}")]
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -945,13 +947,24 @@ impl Worker for LocalSessionWorker {
...
             FromHandler(WorkerSendRequest<LocalSessionWorker>),
         }
-        // waiting for initialize request
-        let evt = self.event_rx.recv().await.ok_or_else(|| {
-            WorkerQuitReason::fatal(
-                LocalSessionWorkerError::TransportTerminated,
-                "get initialize request",
-            )
-        })?;
+        let init_timeout = self.session_config.init_timeout.unwrap_or(Duration::MAX);
```
alexhancock, line 950: I don't understand why the `_or(Duration::MAX)` here. Aren't we setting a default init timeout of 60s?

alexhancock (approved):

crates/rmcp/src/transport/streamable_http_server/session/local.rs
DaleSeo: @alexhancock You're right. The default is 60s but users can set it to `None` to disable the timeout. `unwrap_or(Duration::MAX)` handles that opt-out case. This mirrors how `keep_alive` is handled a few lines below.

alexhancock: Right, but I was thinking we should not have 

```
pub init_timeout: Option<Duration>
```

```
pub init_timeout: Duration
```

seems more clear to me in this situation where it has a builtin default that is always set. It would make it so if the user wants to make it really long they could do the `Duration::MAX`

PR #536, opened 2025-11-11
feat(task): add task support (SEP-1686)

jokemanfire:
Support task manager to control the task , it will be very useful for longtime and concurrence env .
#528 
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #816, opened 2026-04-20
feat(task): route server-originated tasks/* to ClientHandler (SEP-1686)

TheUnderdev:
## Summary

Tasks are bidirectional per SEP-1686: either party can be the requestor or the receiver. The \`ServerHandler\` side is already wired for the client→server task flow (\`tools/call\` augmentation, merged in #536). This PR mirrors the same wiring on the client side so servers that initiate task-augmented requests — notably \`sampling/createMessage\` and \`elicitation/create\` — can follow up with \`tasks/get\`, \`tasks/list\`, \`tasks/result\`, and \`tasks/cancel\` directed at the client.

## Motivation and Context

The 2025-11-25 spec allows a server to task-augment its own \`sampling/createMessage\` or \`elicitation/create\` requests. When it does, the client returns a \`CreateTaskResult\` immediately and the server polls the client for completion. Today this is impossible because \`ServerRequest\` (the typed union of requests a client handler dispatches on) doesn't include the four \`tasks/*\` variants, and \`ClientHandler\` has no methods for them.

Concretely, this means a client cannot honestly advertise any of:

- \`capabilities.tasks.requests.sampling.createMessage\`
- \`capabilities.tasks.requests.elicitation.create\`
- client-side \`capabilities.tasks.list\` / \`capabilities.tasks.cancel\`

…because the server has no typed way to poll it. A well-behaved server sees those caps absent and falls back to synchronous request/response, which blocks its connection/thread for the duration of whatever the client is doing (user reply, slow LLM inference, etc.). Closing this gap lets clients offer the same deferred-result ergonomics the spec already gives servers.

## Changes (purely additive)

- **\`ServerRequest\`** (\`crates/rmcp/src/model.rs\`): add \`GetTaskInfoRequest | ListTasksRequest | GetTaskResultRequest | CancelTaskRequest\` variants; add a \`ServerRequest::method()\` accessor mirroring \`ClientRequest::method()\`; extend the \`variant_extension!\` invocation so the existing \`GetExtensions\`/\`GetMeta\` impls cover the new variants.
- **\`ClientResult\`**: add \`ListTasksResult | GetTaskResult | GetTaskPayloadResult | CancelTaskResult\` response variants. \`GetTaskPayloadResult\` retains its existing custom-\`Deserialize\`-fails behavior so payloads still surface as \`CustomResult\` on the wire, matching the server-side pattern.
- **\`ClientHandler\`** (\`crates/rmcp/src/handler/client.rs\`): add \`list_tasks\`, \`get_task_info\`, \`get_task_result\`, \`cancel_task\` methods with default \`-32601\` \`Method not found\` impls mirroring the server-side signatures; propagate via the \`Box\`/\`Arc\` wrapper macro; dispatch all four from the \`handle_request\` match.
- **Golden schemas**: \`crates/rmcp/tests/test_message_schema/*_json_rpc_message_schema{,_current}.json\` regenerated with \`UPDATE_SCHEMA=1\` to include the new \`ServerRequest\` and \`ClientResult\` variants.

## How Has This Been Tested?

New integration test file: \`crates/rmcp/tests/test_task_client_receiver.rs\` (5 tests).

It stands up a real rmcp client↔server pair over \`tokio::io::duplex\` and verifies the full RPC roundtrip for each of the four methods: the server sends a \`ServerRequest::*\` → the client dispatches into the matching \`ClientHandler\` method → the handler returns the appropriate result variant → the server receives it as the right \`ClientResult\` variant. There's also a default-impl test asserting that the unit \`()\` client returns \`-32601\` for \`tasks/get\`.

Two edge cases are documented in the tests rather than "fixed":

- \`tasks/result\`: \`GetTaskPayloadResult\` has an existing custom \`Deserialize\` that always fails (so that \`CustomResult\` can catch any JSON in the untagged enum). The test asserts the response arrives as \`ClientResult::CustomResult\`, matching the server-side convention.
- \`tasks/cancel\`: \`CancelTaskResult\` and \`GetTaskResult\` share the same JSON shape (both are \`Result + flattened Task\`); serde's untagged enum picks the first match. Callers distinguish by knowing which request they sent. This mirrors existing behavior on the server side and is out of scope for this PR.

All existing rmcp tests still pass (\`cargo test -p rmcp --all-features\`, 48 green test groups, 0 failures).

## Breaking Changes

None intended. The new \`ServerRequest\` / \`ClientResult\` variants are additive, and the new \`ClientHandler\` trait methods have default implementations that preserve today's \`Method not found\` behavior for unaware clients. Unknown methods still route to \`on_custom_request\` exactly as before.

Two observations worth noting:

- Clients that had been extracting task methods from \`CustomRequest\` via a workaround will no longer see them there — they now arrive as the typed variants. Any such code should migrate to the new \`ClientHandler\` methods.
- The \`ClientResult\` enum gains four variants, which will widen JSON Schema definitions for anyone consuming \`ClientJsonRpcMessage\` via \`schemars\`. The \`test_message_schema\` golden files are updated accordingly.

## Types of changes

- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] New and existing tests pass locally (\`cargo test -p rmcp --all-features\`)
- [x] I have added appropriate error handling (handlers fall back to \`-32601\`)
- [ ] I have added or updated documentation as needed — no user-facing docs touched; the trait-level doc comments on each new \`ClientHandler\` method explain when to override.

## Additional context

Related: #528 (Implement SEP-1686: Tasks), #536 (server-side half of SEP-1686).

This PR doesn't change any send-side behavior (the four \`tasks/*\` variants were already in \`ClientRequest\`); it only closes the receive-side gap on \`RoleClient\`.


crates/rmcp/src/handler/client.rs
```diff
@@ -191,6 +207,68 @@ pub trait ClientHandler: Sized + Send + Sync + 'static {
...
+    fn get_task_result(
+        &self,
+        request: GetTaskResultParams,
+        context: RequestContext<RoleClient>,
+    ) -> impl Future<Output = Result<GetTaskPayloadResult, McpError>> + MaybeSendFuture + '_ {
+        let _ = (request, context);
+        std::future::ready(Err(McpError::method_not_found::<GetTaskResultMethod>()))
+    }
+
+    /// Handle a `tasks/cancel` request from a server. Only relevant when
```
alexhancock: `tasks/cancel` is not in the SEP as I see it https://modelcontextprotocol.io/seps/1686-tasks

`tasks/delete` I think is perhaps what is meant?

https://modelcontextprotocol.io/seps/1686-tasks#3-6-deleting-tasks

alexhancock (commented):
Thanks for closing this gap. I left an action required item about `tasks/cancel` vs `tasks/delete`

alexhancock:
@TheUnderdev Do you want to continue on this and address https://github.com/modelcontextprotocol/rust-sdk/pull/816#discussion_r3197318915 ?

crates/rmcp/src/handler/client.rs
TheUnderdev: Good catch — fixed in 6d45614.

Renamed across the board: method string is now `tasks/delete`, types are `DeleteTaskMethod` / `DeleteTaskParams` / `DeleteTaskRequest` / `DeleteTaskResult`, and the trait methods on `ClientHandler` / `ServerHandler` are `delete_task`. The terminal `cancelled` task *status* stays (it's named that way in the SEP).

A couple of related fixes that fell out of the rename:

- `DeleteTaskResult` now carries only `_meta`, matching the §3.6 wire response (the old `CancelTaskResult` flattened a `Task` in, which wasn't in the spec).
- Gave `DeleteTaskResult` the same always-fail `Deserialize` impl used for `GetTaskPayloadResult` so its near-empty shape doesn't shadow `EmptyResult` / `CustomResult` in the untagged `ClientResult` / `ServerResult` enums. On the wire, delete responses now fall through to `EmptyResult`.
- `TaskManager` got a new `delete_task` helper that aborts a running task AND removes any stored completed result — the actual delete semantics. The `#[task_handler]` macro now uses it.

TheUnderdev:
@alexhancock yep — addressed in 6d45614, reply on the review thread with the details. tl;dr renamed everything to `tasks/delete` per SEP §3.6 and aligned the result shape (empty `_meta` only, instead of the old flattened `Task`). Ready for another look when you have a moment.

alexhancock:
@TheUnderdev Sorry for the delay on this. I was traveling for a bit. If you rebase it and address the red checks we can merge it

TheUnderdev:
@alexhancock rebased onto current `main` and addressed the red checks.

**Fixed (now green):**
- **Run Tests / Code Coverage** — these were failing on the `test_message_schema` golden files. Regenerated them so they reflect both the upstream changes pulled in by the rebase and this PR's new `ServerRequest` / `ClientResult` task variants. `cargo test --all-features`, `fmt`, and `clippy -D warnings` all pass locally.
- Resolved a rebase conflict in `handler/server.rs` by combining the new SEP-2164 resource-not-found remapping with the `tasks/delete` rename.

**SemVer Check — needs a call from you.** This one can't go fully green without a decision, and I didn't want to silently work around it:

- I added `[package.metadata.cargo-semver-checks.lints] enum_variant_added = "allow"` to `crates/rmcp/Cargo.toml`. The `ts_union!` protocol enums (`ClientRequest`, `ServerRequest`, `ClientResult`, …) are intentionally exhaustive and gain a variant every spec revision; those additions are non-breaking for downstream consumers (untagged serde + generated `From` impls), so the default major-break classification is just noise for this crate. Happy to drop this if you'd rather handle it differently.
- The **remaining 3 failures are the deliberate `tasks/cancel` → `tasks/delete` removal** (the rename you requested in [r3197318915](https://github.com/modelcontextprotocol/rust-sdk/pull/816#discussion_r3197318915)). Those names shipped in v1.5.0, so removing them is a genuine breaking change that `--release-type minor` (hardcoded in the semver job) will always flag.

Three ways forward, your preference:
1. Accept the break (major bump / override the semver gate for this PR) — keeps the clean removal.
2. I re-add `cancel_task` / `CancelTask*` / the `CancelTaskRequest`/`CancelTaskResult` variants as `#[deprecated]` aliases pointing at the new delete API — fully green, non-breaking, at the cost of keeping the off-spec names around as deprecated.
3. Some other CI policy you'd prefer.

Let me know and I'll turn it around quickly.

alexhancock:
@TheUnderdev Sorry I dropped replying to this again. Another rebase and do option 1 from https://github.com/modelcontextprotocol/rust-sdk/pull/816#issuecomment-4778886668 and we can merge.  We're already in a window where we are going to ship a bunch of breaking changes, so option 1 is good.

alexhancock:
Closing this one out for the moment as many things have changed with the v3 changes. Can re-open if we want to rebase and do a clean version without conflicts.

PR #824, opened 2026-04-30
Fix/issue 817 idle timeout log level

### LINKED ISSUE #817: LocalSessionWorker idle keep-alive timeout emits tracing::error! for normal zombie-session reaping
## Summary

In `rmcp` 1.4.0, when `LocalSessionWorker` hits its configured keep-alive timeout (default 5 min), it returns `WorkerQuitReason::Fatal { error: LocalSessionWorkerError::KeepAliveTimeout(..), context: \"poll next session event\" }`. The outer transport worker then logs this at `tracing::error!` level, even though the docstring explicitly frames this branch as *normal* zombie-session cleanup. That turns every expected idle reap into a red log line for downstream consumers.

## Versions / context

- `rmcp` 1.4.0 (from crates.io)
- Feature set in use: `[\"server\", \"schemars\", \"transport-streamable-http-server\"]`
- Typical deployment: a long-running server, where long-lived agents wake up, perform one MCP tool call, go idle, and create a fresh session next wakeup — so idle reaping fires on every natural agent cycle, not just on error.

## Where it comes from

1. `LocalSessionWorker::run` has the keep-alive branch:

   ```rust
   // src/transport/streamable_http_server/session/local.rs (≈ lines 952–971)
   let keep_alive = self.session_config.keep_alive.unwrap_or(Duration::MAX);
   loop {
       let keep_alive_timeout = tokio::time::sleep(keep_alive);
       let event = tokio::select! {
           // …other arms…
           _ = keep_alive_timeout => {
                   LocalSessionWorkerError::KeepAliveTimeout(keep_alive),
                   \"poll next session event\",
               ))
           }
       };
       // …
   }
   ```

2. The generic worker transport logs all `Fatal` exits at ERROR:

   ```rust
   // src/transport/worker.rs (≈ lines 122–134)
   .inspect_err(|e| match e {
       WorkerQuitReason::Cancelled
       | WorkerQuitReason::TransportClosed
       | WorkerQuitReason::HandlerTerminated => {
           tracing::debug!(\"worker quit with reason: {:?}\", e);
       }
       WorkerQuitReason::Join(e) => {
           tracing::error!(\"worker quit with join error: {:?}\", e);
       }
       WorkerQuitReason::Fatal { error, context } => {
           tracing::error!(\"worker quit with fatal: {error}, when {context}\");
       }
   })
   ```

3. The `SessionConfig::keep_alive` doc itself describes this as a safety net for silently-dropped HTTP connections, not an error condition:

   ```rust
   // src/transport/streamable_http_server/session/local.rs (≈ lines 1060–1089)
   /// The session will be closed after this duration of inactivity.
   ///
   /// This serves as a safety net for cleaning up sessions whose HTTP
   /// connections have silently dropped (e.g., due to an HTTP/2
   /// `RST_STREAM`). Without a timeout, such sessions become zombies […]
   ///
   /// Defaults to 5 minutes. […]
   pub keep_alive: Option<Duration>,
   ```

So the documented-as-normal path produces an ERROR log on every idle reap.

## Concrete symptom

Every ~5 minutes of agent idleness, downstream logs show:

```
ERROR rmcp::transport::worker: worker quit with fatal: keep alive timeout after 300000ms, when poll next session event
```

This is indistinguishable, at the log level, from a real transport fatal (e.g. peer reset mid-stream). Operators who wire ERROR-level alerts get paged on something that is explicitly normal by design.

## Why this matters

- Mixes normal lifecycle events with real transport failures in a single severity bucket.
- Forces every consumer to add target-level filters (`rmcp::transport::worker=warn`) to silence idle reaps, which also silences *real* fatals. No easy downstream fix preserves the original signal.
- Observability best practice: ERROR should be reserved for conditions requiring human attention. Scheduled idle cleanup of a zombie-prevention timer does not qualify.

## Proposal

Split the idle-reap exit from true transport fatals so the logger can tell them apart. Two options, in ascending invasiveness:

**Option A — add a dedicated `WorkerQuitReason` variant (preferred).**
`WorkerQuitReason` is already `#[non_exhaustive]`, so this is additive. Add e.g.:

```rust
pub enum WorkerQuitReason<E> {
    // existing variants…
    #[error(\"Worker idle timeout ({})ms\", _0.as_millis())]
    IdleTimeout(Duration),
}
```

Have `LocalSessionWorker::run` return `WorkerQuitReason::IdleTimeout(keep_alive)` instead of `Fatal { KeepAliveTimeout }`, and in `worker.rs` log it at `tracing::debug!` (or `tracing::info!`) alongside the other expected-exit arms. True transport failures continue to flow through `Fatal` and keep their ERROR severity.

**Option B — widen the `Fatal` arm's logger with a severity hint.**
Keep `Fatal` as-is but add a method on the inner error (e.g. `is_expected_shutdown() -> bool` on the `Worker::Error` trait, defaulting to `false`) that `LocalSessionWorkerError::KeepAliveTimeout` overrides to `true`. The `inspect_err` arm downgrades to `debug!` when `is_expected_shutdown()` returns true. This avoids expanding the enum but couples severity to error identity, which is harder to reason about downstream.

Either way, the invariant should become: **ERROR-level log ⇒ operator intervention likely needed**. Idle reap of a zombie-prevention timer should not meet that bar.

## Happy to follow up with a PR

If the maintainers prefer Option A, I can open a small PR (≤ 50 lines) that:
- adds the new `WorkerQuitReason::IdleTimeout(Duration)` variant,
- returns it from the `keep_alive_timeout` arm in `LocalSessionWorker::run`,
- logs it at `tracing::debug!` in `WorkerTransport::spawn_with_ct`,
- adds a regression test that asserts no ERROR-level records are emitted on idle reap (via `tracing-subscriber::Registry` + a test collector).

Let me know which direction you'd like and I'll send it.

lutz-grex:
Idle keep-alive timeout is normal zombie-session cleanup, not a transport failure. This PR stops it from producing `ERROR`-level log lines that are indistinguishable from real transport fatals.

 ## Motivation and Context

 When `LocalSessionWorker` hits its configured keep-alive timeout (default 5 min), it returns `WorkerQuitReason::Fatal`, which the transport worker logs at `tracing::error!`. The `SessionConfig::keep_alive` docstring explicitly
 frames this as a safety net for silently-dropped HTTP connections — normal lifecycle cleanup, not an error condition.

 This causes every idle reap to emit:
 ```
 ERROR rmcp::transport::worker: worker quit with fatal: keep alive timeout after 300000ms, when poll next session event
 ```

 Operators who wire ERROR-level alerts get paged on something that is normal by design. Downstream consumers cannot filter idle reaps without also silencing real transport fatals.

 Additionally, the post-exit cleanup path in `close_session` could fail with `SessionServiceTerminated` when the worker had already exited due to idle timeout, producing a second spurious error.

 Closes #817

 ## How Has This Been Tested?

 - Added an integration test (`test_keep_alive_timeout_does_not_emit_error_log`) that spins up a Streamable HTTP server with a 200ms keep-alive, connects a client, waits for idle reap, and asserts: no ERROR-level logs are
 emitted, and a DEBUG log with `IdleTimeout` is present.
 - Added a second integration test (`test_explicit_close_on_live_session_succeeds`) that verifies `close_session` on a still-alive worker succeeds without error.
 - Existing tests pass locally (`cargo test`).

 ## Breaking Changes

 None. `WorkerQuitReason` is `#[non_exhaustive]`, so the new `IdleTimeout` variant is additive. The removed `LocalSessionWorkerError::KeepAliveTimeout` variant was not part of the public API.

 ## Types of changes

 - [ ] Documentation update

 ## Checklist

 ## Additional context

 This implements Option A from #817: a dedicated `WorkerQuitReason::IdleTimeout(Duration)` variant that is logged at `debug` level alongside the other expected-exit arms (`Cancelled`, `TransportClosed`, `HandlerTerminated`). True
  transport failures continue to flow through `Fatal` and keep their ERROR severity.

 Changes across two commits:
 1. **`d41d3c8`** — Add `WorkerQuitReason::IdleTimeout`, return it from the keep-alive arm, log at debug. Remove unused `KeepAliveTimeout` error variant.
 2. **`2463369`** — Swallow `SessionServiceTerminated` in `close_session` when the worker has already exited, preventing a spurious error during post-exit cleanup.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -928,8 +928,6 @@ pub enum LocalSessionWorkerError {
     FailToSendInitializeRequest(SessionError),
     #[error("fail to handle message: {0}")]
     FailToHandleMessage(SessionError),
-    #[error("keep alive timeout after {}ms", _0.as_millis())]
-    KeepAliveTimeout(Duration),
     #[error("Transport closed")]
     TransportClosed,
     #[error("Tokio join error {0}")]
@@ -1008,7 +1006,7 @@ impl Worker for LocalSessionWorker {
                     return Err(WorkerQuitReason::Cancelled)
                 }
                 _ = keep_alive_timeout => {
-                    return Err(WorkerQuitReason::fatal(LocalSessionWorkerError::KeepAliveTimeout(keep_alive), "poll next session event"))
+                    return Err(WorkerQuitReason::IdleTimeout(keep_alive))
                 }
             };
             match event {
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -66,9 +66,16 @@ impl SessionManager for LocalSessionManager {
         Ok(response)
     }
     async fn close_session(&self, id: &SessionId) -> Result<(), Self::Error> {
-        let mut sessions = self.sessions.write().await;
-        if let Some(handle) = sessions.remove(id) {
-            handle.close().await?;
+        let handle = {
+            let mut sessions = self.sessions.write().await;
+            sessions.remove(id)
+        };
+        if let Some(handle) = handle {
+            match handle.close().await {
+                // Worker already exited — nothing left to clean up.
+                Ok(()) | Err(SessionError::SessionServiceTerminated) => {}
+                Err(e) => return Err(e.into()),
+            }
         }
         Ok(())
     }
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -935,6 +935,9 @@ pub enum LocalSessionWorkerError {
     FailToSendInitializeRequest(SessionError),
     #[error("fail to handle message: {0}")]
     FailToHandleMessage(SessionError),
+    #[deprecated(note = "idle timeout now surfaces as WorkerQuitReason::IdleTimeout")]
+    #[error("keep alive timeout after {}ms", _0.as_millis())]
+    KeepAliveTimeout(Duration),
     #[error("Transport closed")]
     TransportClosed,
     #[error("Tokio join error {0}")]
--- a/crates/rmcp/src/transport/worker.rs
+++ b/crates/rmcp/src/transport/worker.rs
@@ -1,4 +1,4 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, time::Duration};
 
 use tokio_util::sync::CancellationToken;
 use tracing::{Instrument, Level};
@@ -22,6 +22,8 @@ pub enum WorkerQuitReason<E> {
     TransportClosed,
     #[error("Handler terminated")]
     HandlerTerminated,
+    #[error("Worker idle timeout ({}ms)", _0.as_millis())]
+    IdleTimeout(Duration),
 }
 
 impl<E: std::error::Error + Send + 'static> WorkerQuitReason<E> {
@@ -122,7 +124,8 @@ impl<W: Worker> WorkerTransport<W> {
                 .inspect_err(|e| match e {
                     WorkerQuitReason::Cancelled
                     | WorkerQuitReason::TransportClosed
-                    | WorkerQuitReason::HandlerTerminated => {
+                    | WorkerQuitReason::HandlerTerminated
+                    | WorkerQuitReason::IdleTimeout(_) => {
                         tracing::debug!("worker quit with reason: {:?}", e);
                     }
                     WorkerQuitReason::Join(e) => {
--- a/crates/rmcp/src/transport/worker.rs
+++ b/crates/rmcp/src/transport/worker.rs
@@ -22,7 +22,7 @@ pub enum WorkerQuitReason<E> {
     TransportClosed,
     #[error("Handler terminated")]
     HandlerTerminated,
-    #[error("Worker idle timeout ({}ms)", _0.as_millis())]
+    #[error("Worker idle timeout after {}ms", _0.as_millis())]
     IdleTimeout(Duration),
 }
 
--- a/crates/rmcp/tests/test_streamable_http_idle_timeout_log.rs
+++ b/crates/rmcp/tests/test_streamable_http_idle_timeout_log.rs
@@ -10,15 +10,14 @@ use std::{
 };
 
 use rmcp::transport::streamable_http_server::{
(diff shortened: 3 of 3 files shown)


crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -928,8 +935,6 @@ pub enum LocalSessionWorkerError {
     FailToSendInitializeRequest(SessionError),
     #[error("fail to handle message: {0}")]
     FailToHandleMessage(SessionError),
-    #[error("keep alive timeout after {}ms", _0.as_millis())]
-    KeepAliveTimeout(Duration),
```
DaleSeo, line 932: > The removed LocalSessionWorkerError::KeepAliveTimeout variant was not part of the public API.

I think it's reachable via fully pub module chain:
`rmcp::transport::streamable_http_server::session::local::LocalSessionWorkerError::KeepAliveTimeout`

Could we keep the variant with `#[deprecated]` to avoid a breaking change?

crates/rmcp/src/transport/worker.rs
```diff
@@ -22,6 +22,8 @@ pub enum WorkerQuitReason<E> {
     TransportClosed,
     #[error("Handler terminated")]
     HandlerTerminated,
+    #[error("Worker idle timeout ({}ms)", _0.as_millis())]
```
DaleSeo:
```suggestion
    #[error("Worker idle timeout after {}ms", _0.as_millis())]
```

crates/rmcp/tests/test_streamable_http_idle_timeout_log.rs
```diff
@@ -0,0 +1,246 @@
...
+            .await
+            .unwrap(),
+        "session should have been removed after idle reap"
+    );
+
+    let captured = events.lock().unwrap();
+
+    let error_events: Vec<_> = captured
+        .iter()
+        .filter(|e| e.level == tracing::Level::ERROR)
```
DaleSeo: How about we also filter errors by target? Any unrelated subsystem that logs at `ERROR` would flake this test.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
lutz-grex: Good catch

crates/rmcp/tests/test_streamable_http_idle_timeout_log.rs
lutz-grex: Added target field and scoped both filters to starts_with("rmcp")

DaleSeo (approved):
Thanks for addressing my feedback, @lutz-grex!

PR #829, opened 2026-05-01
chore(rmcp): remove dependency on chrono default features

### LINKED ISSUE #828: Consider disabling chrono default features to avoid unnecessary iana-time-zone/libbe dependency on Haiku
## Summary

`rmcp` currently depends on `chrono` with default features enabled. On Haiku, that can pull in `iana-time-zone-haiku`, which links Haiku's `libbe.so`. This creates an avoidable platform-library dependency, and it can interact badly with other crates that vendor libraries also used by Haiku system DSOs. In my case, this contributed to a final-link failure involving `zstd-sys` and Haiku `libbe.so`.

## Environment

- OS: Haiku x86_64 r1beta5 hrev57937+129
- Rust/Cargo: cargo 1.89.0
- rmcp: 1.3.0
- chrono: 0.4.44
- iana-time-zone: 0.1.65
- iana-time-zone-haiku: 0.1.2

## Current Dependency Behavior

In rmcp 1.3.0, the non-wasm chrono dependency is effectively:

```toml
[target.'cfg(not(all(target_family = "wasm", target_os = "unknown")))'.dependencies.chrono]
version = "0.4.38"
features = ["serde"]
```

Because `default-features = false` is not set, this enables chrono's default features.

chrono's default features include:

```toml
default = ["clock", "std", "oldtime", "wasmbind"]
clock = ["winapi", "iana-time-zone", "now"]
```

On Haiku, iana-time-zone pulls iana-time-zone-haiku, whose build script links libbe, libbe.so itself depends on system libzstd.so.1.

This means that using rmcp on Haiku can introduce libbe and system libzstd into the final link even if the application only needs UTC timestamps. In my case, the final binary also used Rust zstd / zstd-sys, which by default vendors zstd and hides zstd C symbols. The final link then failed on Haiku with:

```
error: linking with `cc` failed: exit status: 1
  = note: /boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/.../ld:
          <binary>: hidden symbol `ZSTD_compressStream` in
          target/debug/deps/libzstd_sys-...rlib(...-zstd_compress.o)
          is referenced by DSO
          final link failed: bad value
          collect2: error: ld returned 1 exit status
```

This is not primarily an rmcp bug, but rmcp appears to be pulling local-timezone support when it may only need UTC time.

From a quick inspection, rmcp appears to use:

- chrono::DateTime<Utc>
- chrono::Utc::now()
- chrono::Duration
- RFC3339 formatting

I did not see obvious usage of chrono::Local or local timezone discovery. 

## Proposed Change

Would it be possible to change the non-wasm chrono dependency to disable default features and enable only the needed features?

For example:

```toml
[target.'cfg(not(all(target_family = "wasm", target_os = "unknown")))'.dependencies.chrono]
version = "0.4.38"
default-features = false
features = ["serde", "now"]
```

This could avoid linking Haiku libbe.so unnecessarily, and reduce dependency footprint on all platforms.

## Workaround Downstream

Downstreams can sometimes work around the specific linker failure by making zstd-sys use system zstd on Haiku

```bash
ZSTD_SYS_USE_PKG_CONFIG=1 cargo build
```

or

```toml
[target.'cfg(target_os = "haiku")'.dependencies]
zstd = { version = "0.13", features = ["pkg-config"] }
```

But it would still be cleaner if rmcp did not pull local timezone support unless it actually needs it.

## Related Upstream issue

[Haiku: vendored hidden zstd symbols conflict with libbe.so dependency on system libzstd](https://github.com/gyscos/zstd-rs/issues/363)

chitao1234:
Disable `chrono` default features for non-wasm `rmcp` builds and enable only the features `rmcp` actually uses. This avoids pulling unnecessary local-timezone support such as `iana-time-zone` / `iana-time-zone-haiku`, which can introduce avoidable platform library dependencies on Haiku. Fixes #828.

## Motivation and Context

Issue #828 reports that `rmcp`’s native `chrono` dependency leaves default features enabled, which pulls in local-timezone support that `rmcp` does not use. On Haiku, that can bring in `iana-time-zone-haiku` and `libbe.so`, contributing to downstream link failures when combined with vendored `zstd`.

This change narrows the native `chrono` dependency to `default-features = false` with only `["serde", "now"]` enabled. That keeps the functionalities `rmcp` uses while avoiding the extra dependency footprint from `chrono` defaults.

## How Has This Been Tested?

`cargo test -p rmcp`, `cargo test -p rmcp --all-features`  and `cargo test --workspace --all` under Linux x86_64

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

See #828.

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -102,7 +102,7 @@ hyper-util = { version = "0.1", features = ["tokio"], optional = true }
 # macro
 rmcp-macros = { workspace = true, optional = true }
 [target.'cfg(not(all(target_family = "wasm", target_os = "unknown")))'.dependencies]
-chrono = { version = "0.4.38", features = ["serde"] }
+chrono = { version = "0.4.38", default-features = false, features = ["serde", "now"] }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
 chrono = { version = "0.4.38", default-features = false, features = [


alexhancock (approved):

alexhancock:
Thanks for the contribution and fixing this

PR #833, opened 2026-05-05
fix: reply -32700 on stdio parse errors instead of closing

### LINKED ISSUE #825: stdio transport: parse errors close stream instead of returning -32700 Parse Error
## Summary

The `rmcp` stdio transport closes the stream cleanly when it receives a malformed JSON line or a UTF-8 BOM-prefixed message, instead of replying with a JSON-RPC `-32700 Parse Error` response as required by [JSON-RPC 2.0 §5.1](https://www.jsonrpc.org/specification#error_object).

This was found while writing integration tests for an MCP server using `rmcp` 1.5.

## Reproduction

Send a malformed line followed by a valid `initialize` request to any rmcp stdio server:

```bash
echo 'not json
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"x","version":"0"}}}' | <your-rmcp-stdio-server>
```

The same behavior occurs when the very first byte sequence is the UTF-8 BOM (`EF BB BF`) before an otherwise-valid JSON-RPC message — the BOM trips the parser and the stream is torn down.

### Observed

Server logs:
```
Error reading from stream: serde error expected ident at line 1 column 2
connection closed: initialize request
```
Process exits with status `0`. The client never receives a `-32700` response and has no way to distinguish parse-side failure from a normal peer-initiated EOF.

### Expected (per JSON-RPC 2.0 §5.1)

The server MUST emit:

```json
{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error"},"id":null}
```

…and remain available to process subsequent valid messages on the same session. Spec quote:

> **-32700  Parse error** — Invalid JSON was received by the server. An error occurred on the server while parsing the JSON text.

The `id` field is `null` because the malformed input means no request id could be extracted.

## Why this matters

- **Spec conformance.** JSON-RPC 2.0 §5.1 requires the `-32700` response. The current behavior silently drops it.
- **Client recovery.** Clients that follow the spec expect to recover from a single bad line and continue the session (e.g. when a wrapper accidentally injects a stray log line, or when a tool emits a BOM). Today the entire connection is torn down with exit `0`, indistinguishable from a clean shutdown.
- **Diagnosability.** A client that sees the stream close cleanly has no signal that the cause was a parse error on its side, only a generic "connection closed" — making it hard to surface actionable errors to users.
- **Robustness against UTF-8 BOM.** Some upstream tooling (Windows-origin scripts, certain editors) prepends a BOM. The line-delimited JSON reader treats the BOM as a parse failure rather than stripping or skipping it. Either tolerating the BOM or replying `-32700` (and continuing) would both be acceptable; silently closing the stream is not.

## Suggested direction

In the stdio transport's read loop, when `serde_json` fails on an incoming line:

1. Construct a `JsonRpcError` with `code: -32700`, `message: "Parse error"`, `id: null`.
2. Send it through the existing sink to the peer.
3. Continue the read loop instead of returning `None` / breaking.

A separate consideration is whether to strip a leading UTF-8 BOM on the very first read, since per RFC 8259 §8.1, JSON parsers MAY ignore a BOM but MUST NOT add one — being lenient on input would prevent a subset of these reports.

## Environment

- `rmcp` 1.5
- Stdio transport
- Reproduces on macOS (Apple Silicon) and Linux

DaleSeo:
Fixes #825

## Motivation and Context

The RMCP stdio transport used to close the stream silently when it encountered a malformed line. This made it hard for clients to tell the difference between a parse failure and a normal end-of-file initiated by the peer. Just one bad byte could bring down the entire session.

Now, the stdio transport responds with a JSON-RPC `-32700 Parse error` when it receives a malformed line, instead of closing the stream silently. It also removes a leading UTF-8 BOM, so messages from Windows tools that have a BOM prefix can be parsed correctly.

## How Has This Been Tested?

Added new tests 

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

### DIFF
--- a/crates/rmcp/src/transport/async_rw.rs
+++ b/crates/rmcp/src/transport/async_rw.rs
@@ -4,7 +4,7 @@ use futures::SinkExt;
 use serde::{Serialize, de::DeserializeOwned};
 use thiserror::Error;
 use tokio::{
-    io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader},
+    io::{AsyncBufReadExt, AsyncRead, AsyncWrite, BufReader},
     sync::Mutex,
 };
 use tokio_util::{
@@ -13,7 +13,10 @@ use tokio_util::{
 };
 
 use super::{IntoTransport, Transport};
-use crate::service::{RxJsonRpcMessage, ServiceRole, TxJsonRpcMessage};
+use crate::{
+    model::ErrorData,
+    service::{RxJsonRpcMessage, ServiceRole, TxJsonRpcMessage},
+};
 
 #[non_exhaustive]
 pub enum TransportAdapterAsyncRW {}
@@ -143,10 +146,11 @@ where
                     tracing::debug!("Parse error on incoming message: {e}");
                     let mut write = self.write.lock().await;
                     let framed = write.as_mut()?;
-                    let inner = framed.get_mut();
-                    if inner.write_all(PARSE_ERROR_RESPONSE).await.is_err()
-                        || inner.flush().await.is_err()
-                    {
+                    let response = TxJsonRpcMessage::<Role>::error(
+                        ErrorData::parse_error("Parse error", None),
+                        None,
+                    );
+                    if framed.send(response).await.is_err() {
                         return None;
                     }
                 }
@@ -207,12 +211,6 @@ fn without_carriage_return(s: &[u8]) -> &[u8] {
 
 /// UTF-8 byte order mark. RFC 8259 §8.1 allows JSON parsers to ignore a leading BOM.
 const UTF8_BOM: &[u8; 3] = b"\xEF\xBB\xBF";
-
-// JSON-RPC 2.0 §5.1: https://www.jsonrpc.org/specification#error_object
-// Hardcoded bytes because `RequestId` has no `Null` variant — we can't
-// build an `id: null` JsonRpcError through the typed codec.
-const PARSE_ERROR_RESPONSE: &[u8] =
-    b"{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32700,\"message\":\"Parse error\"},\"id\":null}\n";
 
 /// Check if a method is a standard MCP method (request, response, or notification).
 /// This includes both requests and notifications defined in the MCP specification.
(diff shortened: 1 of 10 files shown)


crates/rmcp/src/transport/async_rw.rs
```diff
@@ -172,13 +202,18 @@ impl<T> JsonRpcMessageCodec<T> {
...
 }
 
+/// UTF-8 byte order mark. RFC 8259 §8.1 allows JSON parsers to ignore a leading BOM.
+const UTF8_BOM: &[u8; 3] = b"\xEF\xBB\xBF";
+
+// JSON-RPC 2.0 §5.1: https://www.jsonrpc.org/specification#error_object
+// Hardcoded bytes because `RequestId` has no `Null` variant — we can't
+// build an `id: null` JsonRpcError through the typed codec.
+const PARSE_ERROR_RESPONSE: &[u8] =
+    b"{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32700,\"message\":\"Parse error\"},\"id\":null}\n";
```
alexhancock: is there really no way to build something we can serialize to get this value instead of hardcoding it?

alexhancock (dismissed):

crates/rmcp/src/transport/async_rw.rs
DaleSeo: Great question, @alexhancock! This was a workaround because `JsonRpcError` requires its `id` field and `RequestId` has no null variant.                               

I looked into the JSON RPC spec and the MCP spec. I noticed that the MCP spec separated the Response schema into ResultResponse and ErrorResponse in the 2025-11-25 version to match the JSON RPC spec.

|2025-06-18|2025-11-25
-|-
2026-05-07 at 12 01 33|2026-05-07 at 12 00 56

As seen above, the id field is required for ResultResponse while is optional for ErrorResponse.  I checked other SDKs, and they all follow this, so I updated our `JsonRpcError::id` struct to be optional.

https://github.com/modelcontextprotocol/rust-sdk/blob/690df598e8de8df9692e6682612af0cb9b7cb018/crates/rmcp/src/model.rs#L462-L470

alexhancock (approved):

PR #843, opened 2026-05-11
fix(rmcp): flatten Resource variant of PromptMessageContent

### LINKED ISSUE #842: Bug: PromptMessageContent::Resource serializes with doubly-nested resource field (violates MCP spec)
### Summary

`PromptMessageContent::Resource` currently serializes with a doubly-nested `resource` object, producing a wire shape that does not conform to the MCP specification's Prompts → Embedded Resources definition.

Affected versions: at least `rmcp-v1.4.0`, `rmcp-v1.5.0`, `rmcp-v1.6.0`, and the current `main` branch.

### Actual (current) JSON

```json
{
  "type": "resource",
  "resource": {
    "resource": {
      "uri": "file:///example.md",
      "mimeType": "text/markdown",
      "text": "..."
    }
  }
}
```

### Expected (MCP spec) JSON

Per the MCP spec, Prompts → Embedded Resources:
<https://modelcontextprotocol.io/specification/2025-06-18/server/prompts>

```json
{
  "type": "resource",
  "resource": {
    "uri": "file:///example.md",
    "mimeType": "text/markdown",
    "text": "..."
  }
}
```

### Root cause

In `crates/rmcp/src/model/prompt.rs`, the `Resource` variant of `PromptMessageContent` is the only variant missing `#[serde(flatten)]`:

```rust
pub enum PromptMessageContent {
    Text { text: String },
    Image {
        #[serde(flatten)]      // ← present
        image: ImageContent,
    },
    Resource { resource: EmbeddedResource },   // ← #[serde(flatten)] missing
    ResourceLink {
        #[serde(flatten)]      // ← present
        link: super::resource::Resource,
    },
}
```

Because `EmbeddedResource = Annotated<RawEmbeddedResource>` and `Annotated<T>` itself uses `#[serde(flatten)]` on its inner `T`, the inner `RawEmbeddedResource` field `resource: ResourceContents` is exposed *inside* the outer `Resource { resource: ... }` field, producing the doubly-nested wire shape.

The other three variants (`Text`, `Image`, `ResourceLink`) serialize flat. `Resource` is the only outlier.

### Impact

Spec-conformant MCP clients (e.g. Claude Code, which uses Zod schema validation against the spec) reject `prompts/get` responses containing an embedded resource message:

```
ZodError: invalid_union at messages[N].content
  - expected resource.uri (string), received undefined
```

This breaks the embedded-resource path of `prompts/get` end-to-end with conformant clients. Other content types (`text`, `image`, `resource_link`) are unaffected.

### Minimal reproducer

```rust
use rmcp::model::{PromptMessage, PromptMessageRole};

fn main() {
    let m = PromptMessage::new_resource(
        PromptMessageRole::User,
        "file:///example.md".to_string(),
        Some("text/markdown".to_string()),
        Some("# Hello".to_string()),
        None,
        None,
        None,
    );
    println!("{}", serde_json::to_string_pretty(&m).unwrap());
}
```

Observed output shows `content.resource.resource.{uri,mimeType,text}` instead of `content.resource.{uri,mimeType,text}`.

### Notes on related history

- PR #386 (`feat(rmcp): add optional _meta to CallToolResult, EmbeddedResource, and ResourceContents`) — the PR description illustrates the expected (flat) JSON shape, but the implementation still produces the doubly-nested shape. The discrepancy appears to be a review oversight rather than an intentional change.
- PR #381 (`feat: add resource_link support to tools and prompts`) — correctly uses `#[serde(flatten)]` on the `ResourceLink` variant. The same pattern needs to apply to `Resource`.

### Proposed fix

Add `#[serde(flatten)]` to the `Resource` variant's inner field, matching the existing pattern used by `Image` and `ResourceLink` in the same enum. Single-line change. Public Rust API (variant name, field name, field type) is unchanged; only the serde representation is corrected.

I have a fix ready on a fork branch (commits include a regression test and a regenerated schema snapshot) and will open a PR linked to this issue.

ynishi:
Fixes #842

## Motivation and Context

`PromptMessageContent::Resource` currently serializes with a doubly-nested
`resource` object, producing:

```json
{
  "type": "resource",
  "resource": {
    "resource": { "uri": "file:///example.md", "text": "...", "mimeType": "text/markdown" }
  }
}
```

The MCP specification (Prompts — Embedded Resources) defines the wire shape
as a single, flat `resource` field:

```json
{
  "type": "resource",
  "resource": { "uri": "file:///example.md", "text": "...", "mimeType": "text/markdown" }
}
```

The other variants of `PromptMessageContent` (`Text`, `Image`, `ResourceLink`)
already use `#[serde(flatten)]` and serialize flat. `Resource` is the only
outlier: the `resource: EmbeddedResource` field was missing `#[serde(flatten)]`,
so the outer enum tag wrapper plus the inner struct's own `resource` field both
serialize, yielding the double nesting.

This breaks interop with any MCP client that follows the spec literally — the
inner `{uri, text, mimeType}` is unreachable without descending two levels.

## How Has This Been Tested?

- Added a regression test (`test_prompt_message_resource_serialization_is_flat`)
  that constructs a `PromptMessage::new_resource(...)`, serializes it, and
  asserts:
  - `content.type == "resource"`
  - `content.resource.uri`, `content.resource.mimeType`, `content.resource.text`
    are reachable directly (flat shape).
  - `content.resource.resource` is **absent** (regression guard against the
    doubly-nested shape).
- Regenerated the affected schema snapshot via `UPDATE_SCHEMA=1`.
- `cargo test --all-features` passes locally for the `rmcp` crate (lib + all
  integration tests).
- `cargo +nightly fmt --all -- --check` and
  `cargo clippy --all-targets --all-features -- -D warnings` pass.

## Breaking Changes

**Wire format**: yes, but in the spec-compliance direction. Any client that
relied on the previous (non-spec) doubly-nested shape would break, but such
clients are already non-conformant against the MCP spec and against the other
content variants in this same enum.

**Rust API**: none. The variant's field name (`resource`) and type
(`EmbeddedResource`) are unchanged; only the serde representation is corrected.
`cargo semver-checks` should be unaffected.

## Types of Changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Additional Context

Single-line behavioral fix in `crates/rmcp/src/model/prompt.rs`: add
`#[serde(flatten)]` to the `Resource` variant's inner field, matching the
pattern used by `Image` and `ResourceLink` in the same enum.

### DIFF
--- a/crates/rmcp/src/model/prompt.rs
+++ b/crates/rmcp/src/model/prompt.rs
@@ -158,7 +158,10 @@ pub enum PromptMessageContent {
         image: ImageContent,
     },
     /// Embedded server-side resource
-    Resource { resource: EmbeddedResource },
+    Resource {
+        #[serde(flatten)]
+        resource: EmbeddedResource,
+    },
     /// A link to a resource that can be fetched separately
     ResourceLink {
         #[serde(flatten)]
@@ -319,6 +322,57 @@ mod tests {
         assert!(json.contains("\"type\":\"resource_link\""));
         assert!(json.contains("\"uri\":\"file:///test.txt\""));
         assert!(json.contains("\"name\":\"test.txt\""));
+    }
+
+    #[test]
+    fn test_prompt_message_resource_serialization_is_flat() {
+        // Regression test: PromptMessageContent::Resource must serialize to
+        // the spec-compliant flat shape `{ "type": "resource", "resource": { "uri", "mimeType", "text" } }`
+        // and NOT the double-nested shape `{ "type": "resource", "resource": { "resource": {...} } }`.
+        // See: https://modelcontextprotocol.io/specification/2025-06-18/server/prompts
+        let message = PromptMessage::new_resource(
+            PromptMessageRole::User,
+            "alc://packages/sc/narrative".to_string(),
+            Some("text/markdown".to_string()),
+            Some("# Hello".to_string()),
+            None,
+            None,
+            None,
+        );
+
+        let value: serde_json::Value = serde_json::to_value(&message).unwrap();
+
+        // Drill into content
+        let content = value.get("content").expect("content present");
+        assert_eq!(
+            content.get("type").and_then(|v| v.as_str()),
+            Some("resource")
+        );
+
+        let resource = content
+            .get("resource")
+            .expect("resource field present at content level");
+
+        // Spec-compliant: resource.uri / resource.mimeType / resource.text MUST be flat
+        assert_eq!(
+            resource.get("uri").and_then(|v| v.as_str()),
+            Some("alc://packages/sc/narrative"),
+            "expected flat resource.uri, got: {resource:#?}"
+        );
+        assert_eq!(
+            resource.get("mimeType").and_then(|v| v.as_str()),
+            Some("text/markdown")
+        );
+        assert_eq!(
+            resource.get("text").and_then(|v| v.as_str()),
+            Some("# Hello")
+        );
+
+        // Regression guard: content.resource MUST NOT contain a nested `resource` key.
+        assert!(
+            resource.get("resource").is_none(),
+            "double-nested resource detected (regression): {resource:#?}"
+        );
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


alexhancock (approved):

alexhancock:
Thanks for catching and fixing

ynishi:
Thanks for check and merge!

PR #848, opened 2026-05-13
Fix/issue 817 idle timeout log level

### LINKED ISSUE #817: LocalSessionWorker idle keep-alive timeout emits tracing::error! for normal zombie-session reaping

lutz-grex:
Fix issue #817 — route the streamable-http idle keep-alive reap through a dedicated `WorkerQuitReason::IdleTimeout` variant and log it at `debug` instead of `error`. Also stop emitting an `ERROR` log from                            
`LocalSessionManager::close_session` when the worker has already exited on its own.                                                                                                                                                   
                                                                                                                                                                                                                                        
## Motivation and Context                                                                                                                                                                                                               
`LocalSessionWorker`'s keep-alive timeout is documented as a safety net for zombie-session cleanup (silently-dropped HTTP/2 connections, `RST_STREAM`, etc.) — it is normal lifecycle, not a transport failure. Today it is returned as 
`WorkerQuitReason::Fatal { KeepAliveTimeout, .. }`, which `WorkerTransport` logs at `tracing::error!`. The result is a red log line on every idle reap (every ~5 min by default), indistinguishable from a real fatal like a peer reset.
                                                                                                                                                                                                                                        
Operators who alert on `ERROR` get paged on something that is normal by design, and silencing it via target filters (`rmcp::transport::worker=warn`) also silences real fatals. No downstream filter preserves the original signal.     
                                                                                                                                                                                                                                      
There's a related issue in the cleanup path: when the worker exits on its own (e.g. via the idle reap), `spawn_session_worker` calls `close_session` to remove the entry from the map. That call hits `handle.close().await`, which     
returns `SessionError::SessionServiceTerminated` because the worker is already gone — surfacing as another `ERROR` log for what is again a no-op.                                                                                     
                                                                                                                                                                                                                                        
This PR implements Option A from the issue (the preferred direction).                                                                                                                                                                   
 
## How Has This Been Tested?                                                                                                                                                                                                            
- New integration test `crates/rmcp/tests/test_streamable_http_idle_timeout_log.rs`:                                                                                                                                                  
  - `test_keep_alive_timeout_does_not_emit_error_log` — spins up a real `StreamableHttpService` with a 200 ms `keep_alive`, drives a full `initialize` + `notifications/initialized` handshake, waits for the idle reap, and asserts via
 a `tracing_subscriber::Layer` collector that **zero** `ERROR` records are emitted and at least one `DEBUG` record mentions `IdleTimeout`.                                                                                              
  - `test_explicit_close_on_live_session_succeeds` — regression guard that `close_session` on a still-live worker continues to succeed and remove the session.                                                                          
- `cargo test -p rmcp --features "server,transport-streamable-http-server,transport-streamable-http-client-reqwest" --test test_streamable_http_idle_timeout_log` passes locally.                                                       
- Full `cargo test -p rmcp` continues to pass.                                                                                                                                                                                          
                                                                                                                                                                                                                                        
## Breaking Changes                                                                                                                                                                                                                     
None for typical users.                                                                                                                                                                                                               
                                                                                                                                                                                                                                        
- `WorkerQuitReason` is `#[non_exhaustive]`, so adding the `IdleTimeout(Duration)` variant is additive.                                                                                                                                 
- `LocalSessionWorkerError::KeepAliveTimeout` is removed. Any downstream code that pattern-matched on this specific variant would need to update — but the idle-timeout signal now flows through `WorkerQuitReason::IdleTimeout` at the 
transport layer, which is the layer that actually observes it.                                                                                                                                                                          
                                                                                                                                                                                                                                      
## Types of changes                                                                                                                                                                                                                     
- [ ] Documentation update                                                                                                                                                                                                              
                                                                                                                                                                                                                                      
## Checklist                                                                                                                                                                                                                            
 
## Additional context                                                                                                                                                                                                                   
Implementation notes:                                                                                                                                                                                                                 
- `LocalSessionWorker::run` now returns `WorkerQuitReason::IdleTimeout(keep_alive)` from the `keep_alive_timeout` arm instead of constructing a `Fatal { KeepAliveTimeout, .. }`.                                                       
- `WorkerTransport::spawn_with_ct`'s `inspect_err` arm groups `IdleTimeout(_)` with the other expected-exit reasons (`Cancelled`, `TransportClosed`, `HandlerTerminated`) and logs at `debug!`. Real fatals continue to log at `error!`.
- `LocalSessionManager::close_session` swallows `SessionError::SessionServiceTerminated` from `handle.close()` — the worker has already exited and there's nothing left to clean up. Any other error still propagates. The session map  
lock is also dropped before awaiting `handle.close()` to avoid holding the write lock across an `await`.                                                                                                                                
                                                                                                                                                                                                                                        
Closes #817.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -928,8 +928,6 @@ pub enum LocalSessionWorkerError {
     FailToSendInitializeRequest(SessionError),
     #[error("fail to handle message: {0}")]
     FailToHandleMessage(SessionError),
-    #[error("keep alive timeout after {}ms", _0.as_millis())]
-    KeepAliveTimeout(Duration),
     #[error("Transport closed")]
     TransportClosed,
     #[error("Tokio join error {0}")]
@@ -1008,7 +1006,7 @@ impl Worker for LocalSessionWorker {
                     return Err(WorkerQuitReason::Cancelled)
                 }
                 _ = keep_alive_timeout => {
-                    return Err(WorkerQuitReason::fatal(LocalSessionWorkerError::KeepAliveTimeout(keep_alive), "poll next session event"))
+                    return Err(WorkerQuitReason::IdleTimeout(keep_alive))
                 }
             };
             match event {
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -66,9 +66,16 @@ impl SessionManager for LocalSessionManager {
         Ok(response)
     }
     async fn close_session(&self, id: &SessionId) -> Result<(), Self::Error> {
-        let mut sessions = self.sessions.write().await;
-        if let Some(handle) = sessions.remove(id) {
-            handle.close().await?;
+        let handle = {
+            let mut sessions = self.sessions.write().await;
+            sessions.remove(id)
+        };
+        if let Some(handle) = handle {
+            match handle.close().await {
+                // Worker already exited — nothing left to clean up.
+                Ok(()) | Err(SessionError::SessionServiceTerminated) => {}
+                Err(e) => return Err(e.into()),
+            }
         }
         Ok(())
     }
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -935,6 +935,9 @@ pub enum LocalSessionWorkerError {
     FailToSendInitializeRequest(SessionError),
     #[error("fail to handle message: {0}")]
     FailToHandleMessage(SessionError),
+    #[deprecated(note = "idle timeout now surfaces as WorkerQuitReason::IdleTimeout")]
+    #[error("keep alive timeout after {}ms", _0.as_millis())]
+    KeepAliveTimeout(Duration),
     #[error("Transport closed")]
     TransportClosed,
     #[error("Tokio join error {0}")]
--- a/crates/rmcp/src/transport/worker.rs
+++ b/crates/rmcp/src/transport/worker.rs
@@ -1,4 +1,4 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, time::Duration};
 
 use tokio_util::sync::CancellationToken;
 use tracing::{Instrument, Level};
@@ -22,6 +22,8 @@ pub enum WorkerQuitReason<E> {
     TransportClosed,
     #[error("Handler terminated")]
     HandlerTerminated,
+    #[error("Worker idle timeout ({}ms)", _0.as_millis())]
+    IdleTimeout(Duration),
 }
 
 impl<E: std::error::Error + Send + 'static> WorkerQuitReason<E> {
@@ -122,7 +124,8 @@ impl<W: Worker> WorkerTransport<W> {
                 .inspect_err(|e| match e {
                     WorkerQuitReason::Cancelled
                     | WorkerQuitReason::TransportClosed
-                    | WorkerQuitReason::HandlerTerminated => {
+                    | WorkerQuitReason::HandlerTerminated
+                    | WorkerQuitReason::IdleTimeout(_) => {
                         tracing::debug!("worker quit with reason: {:?}", e);
                     }
                     WorkerQuitReason::Join(e) => {
--- a/crates/rmcp/src/transport/worker.rs
+++ b/crates/rmcp/src/transport/worker.rs
@@ -22,7 +22,7 @@ pub enum WorkerQuitReason<E> {
     TransportClosed,
     #[error("Handler terminated")]
     HandlerTerminated,
-    #[error("Worker idle timeout ({}ms)", _0.as_millis())]
+    #[error("Worker idle timeout after {}ms", _0.as_millis())]
     IdleTimeout(Duration),
 }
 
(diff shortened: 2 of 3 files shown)


alexhancock:
Seems like a dupe of https://github.com/modelcontextprotocol/rust-sdk/pull/824/

PR #849, opened 2026-05-14
fix: accept 200 with empty body notification responses in addition to 202

alexhancock:
## Motivation and Context

Some Streamable HTTP MCP servers return `200 OK` with an empty body for JSON-RPC notifications like `notifications/initialized`, instead of the spec-required `202 Accepted`. Previously this could fail with `UnexpectedContentType(None)`. This change treats empty successful responses to notifications/responses/errors as accepted.

## How Has This Been Tested?

Added a focused test for an empty `200 OK` response to a client notification.

Ran:

```bash
cargo test -p rmcp --test test_streamable_http_empty_2xx_notification --features client,transport-streamable-http-client,transport-streamable-http-client-reqwest
cargo fmt -- --check
git diff --check
```

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

- [x] I have read the [[MCP Documentation](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io)

## Additional context

This improves compatibility with non-compliant servers while preserving the existing behavior for normal JSON/SSE responses.

### DIFF
--- a/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
@@ -178,11 +178,25 @@ impl StreamableHttpClient for reqwest::Client {
             .headers()
             .get(reqwest::header::CONTENT_TYPE)
             .map(|ct| String::from_utf8_lossy(ct.as_bytes()).to_string());
+        let content_length = response.content_length();
         let session_id = response
             .headers()
             .get(HEADER_SESSION_ID)
             .and_then(|v| v.to_str().ok())
             .map(|s| s.to_string());
+        // Spec requires 202 Accepted for these, but some servers return an empty 200.
+        // Treat empty success responses as equivalent to Accepted.
+        if status.is_success()
+            && content_length == Some(0)
+            && matches!(
+                message,
+                ClientJsonRpcMessage::Notification(_)
+                    | ClientJsonRpcMessage::Response(_)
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
```diff
@@ -178,11 +178,23 @@ impl StreamableHttpClient for reqwest::Client {
             .headers()
             .get(reqwest::header::CONTENT_TYPE)
             .map(|ct| String::from_utf8_lossy(ct.as_bytes()).to_string());
+        let content_length = response.content_length();
         let session_id = response
             .headers()
             .get(HEADER_SESSION_ID)
             .and_then(|v| v.to_str().ok())
             .map(|s| s.to_string());
+        if status.is_success()
```
jamadeo, line 189: In the past, we've opted to be pretty strict about non-compliant clients. This does have the benefit of exposing issues that other tools might mask.

But other than that, I think it is probably harmless to accept empty 200s as equivalent to 202s.

I would suggest putting a comment note here though, just to state we're deliberately allowing something outside of the spec.

jamadeo (dismissed):

crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
alexhancock: added a comment

jamadeo (approved):

PR #855, opened 2026-05-18
fix: align protocol version negotiation

### LINKED ISSUE #852: Streamable HTTP server accepts mismatched header/body protocol versions and negotiates 2024-11-05 on initialize
**Describe the bug**

When the initial `initialize` request carries a `MCP-Protocol-Version` HTTP header that disagrees with `initialize.params.protocolVersion` in the JSON-RPC body, the server accepts the request without error. In these tests, the negotiated version does not follow either the body or the header, the server returns a server-determined version (2024-11-05) regardless of both input values.

The current MCP `2025-11-25` specification does not explicitly require the server to check body/header consistency on `initialize`, so this is filed as an implementation observation rather than a strict spec-violation claim.

- Environment
  - Reproduced with stable release `rmcp-v1.7.0` (`3529c367`)
  - Also reproduced with a `main` snapshot from 2026-05-15: `cc66e309`
  - Transport: Streamable HTTP server (stateful profile)

**To reproduce**

1. Start a Rust SDK Streamable HTTP server.
2. Send an `initialize` request where the body `protocolVersion` is `2025-11-25` but the `MCP-Protocol-Version` header is `2025-03-26` (or vice versa).
3. Observe that the server returns HTTP 200 with a normal `initialize` result.
4. Check the negotiated protocol version in the response.

**Expected behavior**

Option A: the server rejects the mismatch before negotiation, for example with HTTP 400 or a JSON-RPC Invalid Request error.

Option B: the spec clarifies which field is authoritative, and the SDK documents that behavior and covers it with a regression test.

**Logs**

Both mismatch directions were accepted:

```text
body=2025-11-25 header=2025-03-26 -> HTTP 200, negotiated version = 2024-11-05 (server-determined)
body=2025-03-26 header=2025-11-25 -> HTTP 200, negotiated version = 2024-11-05 (server-determined)
```

The server returned its own preferred protocol version (`2024-11-05`) regardless of both the body and header values. In the observed behavior, the negotiated version did not follow either mismatched input value. However, the negotiated version (2024-11-05) may still differ from what a client would expect from either signal.

With a Streamable HTTP server running, set `ENDPOINT` to the server endpoint and send an initial `initialize` request whose HTTP header and JSON-RPC body disagree:

```bash
ENDPOINT=http://127.0.0.1:8080/mcp

curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'MCP-Protocol-Version: 2025-03-26' \
  --data '{"jsonrpc":"2.0","id":"init-conflict-1","method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"version-conflict-repro","version":"0.1.0"}}}'
```

Then repeat with the values reversed: body `2025-03-26`, header `2025-11-25`.

**Additional context**

- Related: SEP-2575 introduces a related future-state requirement that, for HTTP requests, the `MCP-Protocol-Version` header match `_meta["io.modelcontextprotocol/protocolVersion"]`.

DaleSeo:
Fixes #852

## Motivation and Context

The Rust SDK always responds with `2024-11-05`, no matter what protocol version the client requests in the initialize body. The cross-SDK comparison in spec issue modelcontextprotocol/modelcontextprotocol#2721 shows that Rust is one of only two SDKs that ignore the body.

This PR fix the protocol negotiation logic so that the server echoes the client-requested version when it's known. Otherwise, it returns the server-configured `protocol_version` as a fallback (this is the value set by `with_protocol_version`, which defaults to `LATEST`). 

## How Has This Been Tested?

Add integration tests

## Breaking Changes

None. The dead `UnsupportedProtocolVersion` variant is marked `#[deprecated]`.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,7 @@ jobs:
           cargo semver-checks \
             --package rmcp \
             --baseline-rev ${{ github.event.pull_request.base.sha }} \
+            --release-type minor \
             --only-explicit-features \
             --features default
 
@@ -97,6 +98,7 @@ jobs:
           cargo semver-checks \
             --package rmcp \
             --baseline-rev ${{ github.event.pull_request.base.sha }} \
+            --release-type minor \
             --only-explicit-features \
             --features "$FEATURES"
 
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -69,6 +69,10 @@ pub enum ServerInitializeError {
     #[error("initialize failed: {0}")]
     InitializeFailed(ErrorData),
 
+    #[deprecated(
+        since = "1.8.0",
+        note = "Negotiation now falls back to the server-configured version. This variant is never constructed and will be removed in a future major release."
+    )]
     #[error("unsupported protocol version: {0}")]
     UnsupportedProtocolVersion(ProtocolVersion),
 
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           cargo semver-checks \
             --package rmcp \
             --baseline-rev ${{ github.event.pull_request.base.sha }} \
+            --release-type minor \
```
DaleSeo, line 88: Without this, the CI fails completely on any non-patch change. With it, `release-plz` still determines the actual version bump, but the CI gate now only catches major bumps with breaking regressions. Minor additions, like this deprecation, are allowed.

alexhancock (approved):

PR #858, opened 2026-05-21
Add progress-aware request timeout reset

### LINKED ISSUE #834: Clarify request-timeout semantics for progress notifications and long-running requests
Hi, I was reviewing the current timeout behavior in rmcp and wanted to clarify whether the existing design is intentional.

From reading the code, request timeout appears to be enforced as a fixed wall-clock timeout in `RequestHandle::await_response()`, and on expiry rmcp emits a cancellation notification and returns a timeout error. At the same time, progress notifications seem to be handled independently as a dispatch/observation mechanism, and I could not find logic that refreshes or extends the request timeout when progress is received.

That raised a few questions for me.

Is it intentional that progress notifications do not affect request timeout semantics in rmcp?

If yes, what is the recommended model for long-running non-task requests that legitimately emit progress for a long time? Should clients always set a larger explicit request timeout, and reserve tasks for anything that may exceed that bound?

If no, would you be open to supporting optional progress-aware timeout behavior at the rmcp layer, such as resetting the request timeout on progress or supporting a pair of semantics like inactivity timeout plus max total timeout?

The reason I am asking is that downstream transports may need to retain correlation state while a request is still considered alive by rmcp. Without clear timeout semantics around progress, it is easy for transport-layer cleanup policies to drift from rmcp request lifetime.

I am not proposing a specific change yet. I mainly want to understand whether the current fixed-timeout behavior is a deliberate architectural choice or an area that is still open to evolution.

ContextVM-org:
I added opt-in support for resetting a request timeout when a matching `notifications/progress` notification is received.

This is useful for long-running MCP operations where the peer is still alive and making progress, but the final response may legitimately take longer than the idle request timeout.

The change keeps existing behavior unchanged by default.

### What changed

- Added two request options to [`PeerRequestOptions`](rust-sdk/crates/rmcp/src/service.rs):
  - `reset_timeout_on_progress: bool`
  - `max_total_timeout: Option<Duration>`
- Added small convenience helpers:
  - `PeerRequestOptions::with_timeout(...)`
  - `PeerRequestOptions::reset_timeout_on_progress()`
  - `PeerRequestOptions::with_max_total_timeout(...)`
- Added internal progress timeout watchers keyed by progress token.
- Registered timeout watchers when a request is created, so early progress notifications can be observed.
- Updated `RequestHandle::await_response()` so that:
  - the existing fast path is preserved when progress-aware behavior is not enabled
  - the idle timeout is reset when matching progress arrives and `reset_timeout_on_progress` is enabled
  - `max_total_timeout` acts as a hard cap and is not reset by progress
  - timeout still sends `notifications/cancelled`, matching the existing behavior
- Routed inbound progress notifications through the core service loop before user notification handlers run.
- Preserved existing user-facing progress delivery; `on_progress` handlers still receive the same notifications.
- Added focused tests covering the new behavior.

### Motivation

The current timeout implementation uses a single fixed `tokio::time::timeout(...)` around the response receiver. That works for ordinary requests, but it cannot distinguish between a peer that is offline and a peer that is actively sending progress for a long-running operation.

With this patch, callers can opt into idle-timeout semantics:

```rust
let options = PeerRequestOptions::with_timeout(Duration::from_secs(30))
    .reset_timeout_on_progress()
    .with_max_total_timeout(Duration::from_secs(600));
```

This means:

- if no response or progress is received for 30 seconds, the request times out
- if matching progress keeps arriving, the idle timeout is refreshed
- regardless of progress, the request cannot exceed 10 minutes total

### Backwards compatibility

This should be backwards compatible:

- `reset_timeout_on_progress` defaults to `false`
- `max_total_timeout` defaults to `None`
- existing timeout behavior is preserved unless the new options are explicitly used
- existing progress notification handlers continue to run as before

### Tests

I added a dedicated test target for request timeout/progress behavior.

Covered cases:

- request timeout still expires without progress
- progress does not reset timeout by default
- matching progress resets timeout when enabled
- `max_total_timeout` wins over progress reset
- unrelated progress tokens do not reset the timeout

I also verified that the existing progress subscriber behavior still works.

Commands run:

```sh
cargo check -p rmcp --features server,client,macros
cargo test -p rmcp --test test_progress_subscriber --features server,client,macros
cargo test -p rmcp --test test_request_timeout_progress --features server,client,macros
```

All focused checks passed.
resolves: #834


ContextVM-org:
Is this not desired? The pr still open with not reviews since the last month

crates/rmcp/src/service.rs
```diff
@@ -314,40 +348,142 @@ pub struct RequestHandle<R: ServiceRole> {
...
+                    reset_timeout_on_progress,
+                )
+                .await
+            };
+
+        Self::cleanup_progress_timeout_watcher(
+            &progress_timeout_watchers,
+            &progress_token,
+            has_progress_reset_rx,
+        )
```
DaleSeo, line 396: `RequestHandle` already stores `peer: Peer<R>` which owns the same `progress_timeout_watchers` registry. So, we might not need `progress_timeout_watchers` as a separate field.

```suggestion
        Self::cleanup_progress_timeout_watcher(
            &self.peer.progress_timeout_watchers,
            &progress_token,
            has_progress_reset_rx,
        )
```

crates/rmcp/src/service.rs
```diff
@@ -314,40 +348,142 @@ pub struct RequestHandle<R: ServiceRole> {
...
+        let max_total_timeout = self.options.max_total_timeout;
+        let reset_timeout_on_progress = self.options.reset_timeout_on_progress;
+
+        let has_progress_reset_rx = self.progress_reset_rx.is_some();
+        let progress_timeout_watchers = self.progress_timeout_watchers.clone();
+        let progress_token = self.progress_token.clone();
+
+        let result =
+            if timeout.is_some() && !reset_timeout_on_progress && max_total_timeout.is_none() {
+                let timeout = timeout.expect("timeout is checked above");
```
DaleSeo: This branch checks `timeout.is_some()` and then immediately unwraps it with `expect`. Clippy flags this pattern. Can we match on the option shape to make the three timeout paths clear without needing an assertion message?

crates/rmcp/src/service.rs
```diff
@@ -314,40 +348,142 @@ pub struct RequestHandle<R: ServiceRole> {
...
-                            reason: Some(Self::REQUEST_TIMEOUT_REASON.to_owned()),
-                        },
-                        method: crate::model::CancelledNotificationMethod,
-                        extensions: Default::default(),
-                    };
-                    let _ = self.peer.send_notification(notification.into()).await;
-                    error
+    pub const REQUEST_MAX_TOTAL_TIMEOUT_REASON: &str = "maximum total timeout exceeded";
+
+    async fn send_timeout_cancel_notification(&self, reason: &str) {
```
DaleSeo: Can we place this helper below the public method, `await_response` just like `await_response_with_progress_timeout`?

crates/rmcp/src/service.rs
```diff
@@ -314,40 +348,142 @@ pub struct RequestHandle<R: ServiceRole> {
...
+                request_id: self.id.clone(),
+                reason: Some(reason.to_owned()),
+            },
+            method: crate::model::CancelledNotificationMethod,
+            extensions: Default::default(),
+        };
+        let _ = self.peer.send_notification(notification.into()).await;
+    }
+
+    async fn cleanup_progress_timeout_watcher(
```
DaleSeo: Can we place this helper below the public method, `cancel`?

crates/rmcp/tests/test_request_timeout_progress.rs
```diff
@@ -0,0 +1,211 @@
...
+        &self,
+        _params: ProgressNotificationParam,
+        _context: rmcp::service::NotificationContext<rmcp::RoleClient>,
+    ) {
+        self.progress_count.fetch_add(1, Ordering::SeqCst);
+    }
+}
+
+struct ProgressTimeoutServer {
+    tool_router: ToolRouter<Self>,
```
DaleSeo: Clippy flags this as unused. This test server only exposes tools and does not customize `ServerHandler`, so we can use the `#[tool_router(server_handler)]` pattern.

```rs
struct ProgressTimeoutServer;

#[tool_router(server_handler)]
impl ProgressTimeoutServer {
```

crates/rmcp/src/service.rs
DaleSeo: Shouldn't we remove the watcher on this error path before returning `TransportClosed`?

DaleSeo (commented):
@ContextVM-org Sorry this sat for so long. I think this feature is useful. I added a few minor comments.

ContextVM-org:
Thanks for the review, @DaleSeo, all suggestions applied in the latest push.

1. expect after is_some()
   Replaced the if/else if/else chain with a match (timeout, max_total_timeout, reset_timeout_on_progress). The three timeout paths are now explicit match arms, and the expect is gone entirely.

2. send_timeout_cancel_notification placement
   Moved it just below the public await_response, next to await_response_with_progress_timeout.

3. cleanup_progress_timeout_watcher placement
   Moved it just below the public cancel.

4. Unused tool_router field in the test
   Switched to the #[tool_router(server_handler)] pattern with a unit struct, dropping the field and the separate #[tool_handler] impl ServerHandler block. Cleaned up the now-unused ServerHandler / ToolRouter / tool_handler imports.

5. Watcher leak on the TransportClosed error path
   Good catch. The watcher entry is now removed before returning the error, so it can't be orphaned. Since progress tokens come from an atomic counter and are never reused, an orphaned entry would otherwise live for the Peer's lifetime.

Let me know if anything else needs adjusting.

DaleSeo (approved):

PR #865, opened 2026-05-31
fix(rmcp): add Audio variant to PromptMessageContent

### LINKED ISSUE #864: Bug: PromptMessageContent is missing the Audio variant (diverges from MCP spec ContentBlock)
### Summary

`PromptMessageContent` does not include an `Audio` variant, so it diverges from the MCP specification's `ContentBlock` union used by `PromptMessage.content`. As a result, rmcp **cannot deserialize audio content in `prompts/get` responses** (and cannot construct it server-side). This is the audio analogue of the embedded-resource bug fixed in #842 / #843.

Affected versions: `rmcp-v1.4.0` through `rmcp-v1.7.0`, and the current `main` branch.

### Actual (current) behavior

`PromptMessageContent` has four variants (`crates/rmcp/src/model/prompt.rs`):

```rust
pub enum PromptMessageContent {
    Text { text: String },
    Image { #[serde(flatten)] image: ImageContent },
    Resource { #[serde(flatten)] resource: EmbeddedResource },
    ResourceLink { #[serde(flatten)] link: super::resource::Resource },
}
```

There is no `Audio` variant, and the enum is `#[serde(tag = "type")]` with no catch-all, so an `{"type":"audio",...}` content block fails to deserialize:

```
Error("unknown variant `audio`, expected one of `text`, `image`, `resource`, `resource_link`")
```

### Expected (MCP spec)

Per the spec, `PromptMessage.content` is a `ContentBlock`, whose union includes `AudioContent`:

```ts
export type ContentBlock =
  | TextContent
  | ImageContent
  | AudioContent      // ← missing from rmcp's PromptMessageContent
  | ResourceLink
  | EmbeddedResource;
```

(MCP spec, Server → Prompts: https://modelcontextprotocol.io/specification/2025-06-18/server/prompts — `ContentBlock` is unchanged through 2025-11-25.)

### Root cause

The `Audio` variant was simply never added to `PromptMessageContent`. The supporting type already exists — `AudioContent` (`Annotated<RawAudioContent>`, `crates/rmcp/src/model/content.rs`) — and `Audio` is already a variant of the general `RawContent` enum (used by tool results and sampling). Only `PromptMessageContent` omits it.

### Impact

A spec-conformant MCP server that returns audio content in a `prompts/get` response breaks rmcp clients end-to-end: `Peer::get_prompt` fails with `ServiceError::UnexpectedResponse` because the `GetPromptResult` body cannot be deserialized. The other content types (`text`, `image`, `resource`, `resource_link`) are unaffected. This mirrors the conformance break in #842, where the same class of mismatch was treated as a bug and fixed.

Server-side, there is also no way to construct an audio prompt message (no `PromptMessage::new_audio`, unlike `new_image`).

### Minimal reproducer

```rust
use rmcp::model::PromptMessageContent;

fn main() {
    // A spec-valid audio content block from prompts/get:
    let json = r#"{"type":"audio","data":"<base64>","mimeType":"audio/wav"}"#;
    let result = serde_json::from_str::<PromptMessageContent>(json);
    println\!("{result:?}"); // Err: unknown variant `audio`
}
```

### Proposed fix

Add the variant, mirroring `Image` (which already flattens `ImageContent`):

```rust
// in the content import:
content::{AudioContent, EmbeddedResource, ImageContent},

// in PromptMessageContent:
    /// Audio content with base64-encoded data
    Audio { #[serde(flatten)] audio: AudioContent },
```

A round-trip test (in the style of #843's regression test) and a `PromptMessage::new_audio` constructor, for symmetry with `new_image`, would round out the change.

### Related

- #842 / #843 — the embedded-resource analogue (same `PromptMessageContent` enum, same class of spec divergence), fixed by flattening the `Resource` variant.
- #684 — Conformance Testing (covers tool-result audio, which already works; prompt-message audio is the gap here).

bricef:
## What

Adds the missing `Audio` variant to `PromptMessageContent`, bringing it in line with the MCP spec's prompt-message `ContentBlock` union (`text | image | audio | resource_link | resource`).

## Why

`PromptMessageContent` had no `Audio` variant, and since it is `#[serde(tag = "type")]` with no catch-all, a spec-conformant `{"type":"audio",...}` content block fails to deserialize:

```
unknown variant `audio`, expected one of `text`, `image`, `resource`, `resource_link`
```

This breaks `prompts/get` end-to-end for any server that returns audio prompt content. It is the audio analogue of #842 / #843 (the embedded-resource variant). The supporting `AudioContent` type already exists, and `Audio` is already a variant of the general `RawContent` enum (tool results, sampling) — only `PromptMessageContent` omitted it.

Fixes #864.

## Changes

- Add the flattened `Audio { #[serde(flatten)] audio: AudioContent }` variant (mirrors `Image`).
- Add `PromptMessage::new_audio` constructor (mirrors `new_image`).
- Add serialization (flat shape, camelCase `mimeType`) + deserialization regression tests, in the style of #843's `test_prompt_message_resource_serialization_is_flat`.
- Regenerate the `ServerJsonRpcMessage` schema snapshot (`UPDATE_SCHEMA=1`) to include the new audio content block.

## Verification

- `just check` (clippy, `--all-features --all-targets -- -D warnings`): clean.
- `just test` (`cargo test --all-features`): passing, including the regenerated schema snapshot and the two new prompt tests. (`tests/test_with_python.rs` requires an external Python MCP server unavailable in my environment; unrelated to this change.)
- `cargo +nightly fmt --check`: clean.

### DIFF
--- a/crates/rmcp/src/model/prompt.rs
+++ b/crates/rmcp/src/model/prompt.rs
@@ -2,7 +2,7 @@ use serde::{Deserialize, Serialize};
 
 use super::{
     AnnotateAble, Annotations, Icon, Meta, RawEmbeddedResource,
-    content::{EmbeddedResource, ImageContent},
+    content::{AudioContent, EmbeddedResource, ImageContent},
     resource::ResourceContents,
 };
 
@@ -157,6 +157,11 @@ pub enum PromptMessageContent {
         #[serde(flatten)]
         image: ImageContent,
     },
+    /// Audio content with base64-encoded data
+    Audio {
+        #[serde(flatten)]
+        audio: AudioContent,
+    },
     /// Embedded server-side resource
     Resource {
         #[serde(flatten)]
@@ -224,6 +229,29 @@ impl PromptMessage {
                     data: base64,
                     mime_type: mime_type.into(),
                     meta,
+                }
+                .optional_annotate(annotations),
+            },
+        }
+    }
+
+    /// Create a new audio message. `annotations` is optional.
+    #[cfg(feature = "base64")]
+    pub fn new_audio(
+        role: PromptMessageRole,
+        data: &[u8],
+        mime_type: &str,
+        annotations: Option<Annotations>,
+    ) -> Self {
+        use base64::{Engine, prelude::BASE64_STANDARD};
+
+        let base64 = BASE64_STANDARD.encode(data);
+        Self {
+            role,
+            content: PromptMessageContent::Audio {
+                audio: crate::model::RawAudioContent {
+                    data: base64,
+                    mime_type: mime_type.into(),
                 }
                 .optional_annotate(annotations),
             },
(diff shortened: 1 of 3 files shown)


DaleSeo (approved):
@bricef Sorry for the delay, and thanks for your contribution. This seems like a reasonable fix to me.

According to the semver check, `PromptMessageContent` is a public exhaustive enum. So, adding a new variant is a semver-breaking API change. I think this change alone might be too small to justify a major release though.

I'll talk to the other maintainers to decide whether to batch this with the other breaking changes for the upcoming spec support or to cut v2.0 before that.

PR #887, opened 2026-06-04
fix(auth): align OAuth metadata discovery ordering

### LINKED ISSUE #878: Implement SEP-2351: RFC 8414 well-known URI suffix handling
# SEP-2351: Explicitly specify RFC 8414 well-known URI suffix for MCP — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2351
**Needs code changes:** Yes (Small)

## Summary
States that MCP uses the default `oauth-authorization-server` well-known URI suffix from RFC 8414
§3.1 and does **not** define an application-specific suffix. Clarifies metadata discovery URL
construction.

The client **already** builds RFC 8414-style candidates in `auth.rs` (the discovery-candidate
builder around lines 1380–1395 pushes `/.well-known/oauth-authorization-server`, the path-insertion
variants `/.well-known/oauth-authorization-server/{path}` and
`/.well-known/openid-configuration/{path}`, the path-append variant
`/{path}/.well-known/openid-configuration`, and a canonical OAuth fallback). So this is mostly a
**verify-and-tighten** task, not a from-scratch build.

Conformance currently reports **`auth/metadata-var3` FAIL (0/4)** — a metadata-discovery edge case.
The most likely culprits given the current code: candidate **ordering/priority** vs. the spec, the
path-append-only-for-OIDC behavior, or how the canonical fallback interacts with issuers that have
path components.

## Proposed work
- [ ] Pin the exact candidate set + **ordering** required by SEP-2351 / RFC 8414 §3.1 and reconcile it with the existing `push_candidate(..)` list (lines ~1383–1394) — the bug is probably ordering or a missing/extra variant, not a missing suffix.
- [ ] Confirm only the default `oauth-authorization-server` suffix is used (no MCP-specific suffix) — current code already does this; assert it with a test.
- [ ] Trace the `auth/metadata-var3` scenario against the candidate list to find which URL it expects vs. what the client tries first; fix the discrepancy.
- [ ] Re-run `auth/metadata-var*` to green; add unit tests over the candidate builder for issuers with and without path components.

## Affected areas
`crates/rmcp/src/transport/auth.rs` (the well-known discovery candidate builder; `discover_metadata` / `try_discover_oauth_server`).

## Notes / risks
- Pairs with SEP-2352 (AS binding/migration) and the `auth/metadata-var3` conformance fix. Bundle with the broader auth cleanup.

jstar0:
Fixes #878.

## Summary

- align path-bearing authorization-server discovery with the MCP authorization spec / RFC 8414 ordering
- keep OAuth authorization-server metadata discovery ahead of OpenID Connect fallback candidates
- add regression coverage for both candidate ordering and a real client credentials discovery flow

## Changes

For a resource URL such as `https://example.com/mcp`, the direct authorization-server discovery list is:

- `https://example.com/.well-known/oauth-authorization-server/mcp`
- `https://example.com/.well-known/openid-configuration/mcp`
- `https://example.com/mcp/.well-known/openid-configuration`
- `https://example.com/.well-known/oauth-authorization-server`

The first three entries match the path-bearing issuer ordering from the MCP authorization spec. The existing canonical OAuth fallback remains last.

The client credentials regression now serves metadata at the RFC 8414 path-inserted OAuth endpoint, so the test exercises the same URL form as the candidate builder.

## Verification

```text
cargo test -p rmcp --features auth generate_discovery_urls
cargo test -p rmcp --features auth --test test_client_credentials test_client_credentials_discovers_path_inserted_oauth_metadata
cargo test -p rmcp --features auth
cargo test --all-features
cargo fmt --all -- --check
cargo clippy -p rmcp --features auth --lib -- -D warnings
cargo clippy -p rmcp --features auth --test test_client_credentials -- -D warnings
```

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1409,7 +1409,7 @@ impl AuthorizationManager {
             push_candidate("/.well-known/oauth-authorization-server".to_string());
             push_candidate("/.well-known/openid-configuration".to_string());
         } else {
-            // Path components present: follow spec priority order
+            // Path components present: prefer OAuth discovery before OpenID Connect fallbacks.
             // 1. OAuth 2.0 with path insertion
             push_candidate(format!("/.well-known/oauth-authorization-server/{trimmed}"));
             // 2. OpenID Connect with path insertion
--- a/crates/rmcp/tests/test_client_credentials.rs
+++ b/crates/rmcp/tests/test_client_credentials.rs
@@ -136,6 +136,25 @@ async fn start_mock_server() -> (String, SocketAddr) {
     (base_url, addr)
 }
 
+async fn start_path_insert_metadata_server() -> (String, SocketAddr) {
+    let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let addr = listener.local_addr().unwrap();
+    let base_url = format!("http://{}", addr);
+
+    let app = Router::new()
+        .route(
+            "/.well-known/oauth-authorization-server/mcp",
+            get(auth_server_metadata_handler),
+        )
+        .route("/token", post(token_handler));
+
+    tokio::spawn(async move {
+        axum::serve(listener, app).await.unwrap();
+    });
+
+    (base_url, addr)
+}
+
 #[tokio::test]
 async fn test_client_credentials_flow_client_secret() {
     let (base_url, _addr) = start_mock_server().await;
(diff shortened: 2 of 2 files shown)


michaelneale (dismissed):
thanks @jstar0 from a brief reading of the spec: https://www.rfc-editor.org/info/rfc8414/#section-3.1 - then the tenant should be at the end, not before .well-known: 

ie: 

https://auth.example.com/.well-known/oauth-authorization-server/tenant1

not so much: 

https://auth.example.com/tenant1/.well-known/oauth-authorization-server

I think, but otherwise looking good.

DaleSeo (approved):

PR #883, opened 2026-06-03
feat: specify OIDC application_type during dynamic client registration (SEP-837)

stefanoamorelli:
Closes #880

## Motivation and Context

[SEP-837](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/837) clarifies the client type requirements for OIDC Dynamic Client Registration: a client must send an `application_type`, because a server that receives none assumes `"web"`. That clashes with the loopback and `localhost` redirect URIs CLI and desktop clients use, and the server can reject the registration over it.

`register_client` currently sends no `application_type`, so the SDK hits exactly that. This PR makes it always send one: `"native"` by default (matching the loopback redirect), with an `OAuthClientConfig::with_application_type(..)` builder to opt into `"web"`. I also set `"native"` in `client-metadata.json` so the URL-based client metadata (SEP-991) and dynamic registration agree.

## How Has This Been Tested?

Added unit tests covering the serialized request (`application_type` present when set, omitted when `None`) and the config default and override. The full rmcp suite passes with the auth feature, and `cargo +nightly fmt --all` and clippy are clean.

## Breaking Changes

No API breaking changes: `OAuthClientConfig` is `#[non_exhaustive]` and built via `OAuthClientConfig::new`, and `ClientRegistrationRequest` is crate-private.

One behavior change: `register_client` now always includes `application_type` (`"native"` unless overridden) where it sent none before. Non-OIDC servers ignore it; OIDC servers treat the client as native instead of assuming `"web"`.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

No new error handling was needed: the change only sets a request field, and the registration failure SEP-837 mentions is already surfaced by `AuthError::RegistrationFailed`.

PR #888, opened 2026-06-04
feat: accumulate client-side scopes during step-up authorization (SEP-2350)

### LINKED ISSUE #877: Implement SEP-2350: Clarify client-side scope accumulation (step-up auth)
# SEP-2350: Clarify client-side scope accumulation in step-up authorization — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350
**Needs code changes:** Yes (Small)

## Summary
Clarifies step-up authorization scope handling:
- Servers report scopes needed for the **current operation** in 403 / insufficient-scope errors (RFC 6750 §3.1), **not** the union of previously granted scopes.
- Clients **accumulate** scopes: during re-authorization the client requests the **union** of previously requested scopes and the newly challenged scopes.
- "Authoritative" in Protected Resource Metadata means scopes required for the current operation, not the exclusive set the client should request.

The OAuth client's scope logic lives in `crates/rmcp/src/transport/auth.rs` on
`AuthorizationManager`:
- `select_scopes(www_authenticate_scope, default_scopes)` → calls `select_base_scopes(..)` then `add_offline_access_if_supported(..)`.
- `select_base_scopes(..)` currently **returns early with a single source** in priority order (WWW-Authenticate scope → stored `www_auth_scopes` → `resource_scopes` (RFC 9728) → AS `scopes_supported` → defaults). It **replaces** rather than **unions** — so on a step-up 403 it returns only the freshly challenged scope and drops previously granted ones.
- Supporting state: `current_scopes: RwLock<Vec<String>>`, `scope_upgrade_attempts`, `ScopeUpgradeConfig`, `can_attempt_scope_upgrade()`, `get_current_scopes()`, `WWWAuthenticateParams::is_insufficient_scope()`.

Conformance currently reports **`auth/scope-step-up` FAIL (13/14)** — consistent with the
replace-not-union behavior above.

## Proposed work
- [ ] In `select_base_scopes` (or a new wrapper), when a step-up challenge arrives, union the newly challenged scopes with `current_scopes` (the previously requested/granted set) rather than returning a single source early.
- [ ] Persist the accumulated union back into `current_scopes` after a successful re-authorization so subsequent rounds keep growing the set.
- [ ] Treat the server's reported scopes as "needed for the current operation," not authoritative/exclusive (per the SEP) — merge, don't overwrite.
- [ ] De-duplicate and keep ordering stable for testability.
- [ ] Re-run `auth/scope-step-up` to green; add unit tests covering multiple step-up rounds accumulating scopes.

## Affected areas
`crates/rmcp/src/transport/auth.rs` (`AuthorizationManager::select_scopes` / `select_base_scopes`, `current_scopes`).

## Notes / risks
- Directly fixes a known failing conformance scenario. Good candidate to bundle with SEP-2351/2352/2468 auth work.

stefanoamorelli:
Closes #877

## Motivation and Context

[SEP-2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350) makes scope accumulation a client responsibility. A `403` / `insufficient_scope` challenge carries only the scopes for the current operation (RFC 6750 §3.1), so on re-authorization the client must request the union of what it asked for before and what was just challenged.

`select_base_scopes()` returned a single source, so a challenge replaced the earlier scopes instead of widening them, and step-up dropped previously granted permissions (the `auth/scope-step-up` conformance test failed for this). This PR unions `current_scopes`, the WWW-Authenticate challenge, and the resource metadata scopes (RFC 9728); defaults only seed the first request.

On exchange, an explicit `scope` list stays authoritative so a server can still narrow the grant. When the server omits `scope` it has granted exactly what the client requested (RFC 6749 §5.1), so the grant falls back to the scopes requested in that round, not the previously granted set: otherwise a step-up the server confirms by omitting `scope` would silently drop the just-added permission and loop on the same 403. The requested set was not persisted anywhere `exchange_code_for_token` could read it, so it is now recorded on `StoredAuthorizationState` per authorization (defaulting empty for states stored before the field existed). Refreshing is unchanged: RFC 6749 §6 caps a refresh request at the originally granted scopes, so the existing fallback there is already correct.

## How Has This Been Tested?

Added unit tests for multi-round accumulation, union-not-replace, dedup of an already-requested challenge, and resource metadata accumulation. New tests cover grant resolution when the token response omits `scope` (falls back to the requested set), honoring an explicit server downgrade, the no-request fallback, and `requested_scopes` round-tripping through the state store (including the `serde(default)` for older states). The rmcp suite passes with the `auth` feature, and `cargo +nightly fmt --all` and clippy are clean.

## Breaking Changes

No API breaking changes: `StoredAuthorizationState` is `#[non_exhaustive]` and built via `StoredAuthorizationState::new`, so the added `requested_scopes` field and `with_requested_scopes` builder are additive (the same shape as #883), and the `--release-type minor` semver-checks job covers it. The behavior change is required by the SEP: re-authorization requests the union of prior and challenged scopes, and when the token response omits `scope` (RFC 6749 §5.1) the client retains the scopes it requested in that round rather than the prior grant. An explicit `scope` from the server is still authoritative and may narrow the grant.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This is the client half of SEP-2350; server-side per-operation scope reporting is out of scope for the SDK. No new error handling was needed since the change only widens the requested set.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -164,6 +164,10 @@ pub struct StoredAuthorizationState {
     #[serde(default)]
     pub require_issuer: bool,
     pub created_at: u64,
+    /// scopes requested in this round, used to resolve the grant when the token response omits
+    /// `scope` (RFC 6749 §5.1)
+    #[serde(default)]
+    pub requested_scopes: Vec<String>,
 }
 
 impl std::fmt::Debug for StoredAuthorizationState {
@@ -174,6 +178,7 @@ impl std::fmt::Debug for StoredAuthorizationState {
             .field("expected_issuer", &self.expected_issuer)
             .field("require_issuer", &self.require_issuer)
             .field("created_at", &self.created_at)
+            .field("requested_scopes", &self.requested_scopes)
             .finish()
     }
 }
@@ -225,7 +230,14 @@ impl StoredAuthorizationState {
                 .duration_since(std::time::UNIX_EPOCH)
                 .map(|d| d.as_secs())
                 .unwrap_or(0),
+            requested_scopes: Vec::new(),
         }
+    }
+
+    /// record the scopes requested in this authorization round (SEP-2350)
+    pub fn with_requested_scopes(mut self, scopes: Vec<String>) -> Self {
+        self.requested_scopes = scopes;
+        self
     }
 
     pub fn into_pkce_verifier(self) -> PkceCodeVerifier {
@@ -1054,7 +1066,7 @@ impl AuthorizationManager {
 
         let (auth_url, csrf_token) = auth_request.url();
 
-        // store pkce verifier and expected issuer for later use via state store
+        // store pkce verifier, expected issuer, and the requested scopes for later use via state store
         let expected_issuer = self
             .metadata
             .as_ref()
@@ -1074,7 +1086,8 @@ impl AuthorizationManager {
             &csrf_token,
             expected_issuer,
             require_issuer,
-        );
+        )
+        .with_requested_scopes(scopes.iter().map(|s| s.to_string()).collect());
         self.state_store
             .save(csrf_token.secret(), stored_state)
             .await?;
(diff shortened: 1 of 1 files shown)


michaelneale:
nice @stefanoamorelli - only thought so far `current_scopes` is remembered, but if scope is omitted when a token response is returned when user asked for an additional scope, it won't remember what it asked for (ie will only fall back to current_scopes, not the new ones). Not sure if that is intentional or not. I am a little confused as to what is meant to happen when a token comes back without a scopes list. 

Looks like that should be fixed: 

```
scope
         OPTIONAL, if identical to the scope requested by the client;
         otherwise, REQUIRED.  The scope of the access token as
         described by [Section 3.3](https://www.rfc-editor.org/info/rfc6749/#section-3.3).
```

https://www.rfc-editor.org/info/rfc6749/#section-5.1

so I think that is all, if that can be patched then this could be good to go.

michaelneale (dismissed):
Around the `exchange_code_for_token` change, roughly original/current PR hunk near line **1239**:

```rust
let granted_scopes: Vec<String> = match token_result.scopes() {
    Some(scopes) => scopes.iter().map(|s| s.to_string()).collect(),
    None => self.current_scopes.read().await.clone(),
};
```

as mentioned in other comment, I think that should be the scopes requested, not the current scopes?

stefanoamorelli:
> Around the `exchange_code_for_token` change, roughly original/current PR hunk near line **1239**:
> 
> ```rust
> let granted_scopes: Vec<String> = match token_result.scopes() {
>     Some(scopes) => scopes.iter().map(|s| s.to_string()).collect(),
>     None => self.current_scopes.read().await.clone(),
> };
> ```
> 
> as mentioned in other comment, I think that should be the scopes requested, not the current scopes?

@michaelneale thanks for this, I dig deeper in [RFC 6749 §5.1](https://www.rfc-editor.org/info/rfc6749/#section-5.1) and you're right (good catch). Basically an omitted `scope` means the grant equals the *request*, so falling back to `current_scopes` dropped the scope that step-up added and would loop on the same _403_.

I amended and force-pushed (`fdceb0f`) (+ rebased) so the grant now falls back to the scopes requested that round (recorded on `StoredAuthorizationState`), while an explicit server `scope` stays authoritative.

qq on `select_base_scopes`: it already used `try_read()` for the other scope sources, so I kept that and added `current_scopes` the same way. But IIUC `try_read()` skips a source if the lock is busy. should these be `read().await` instead so we never silently drop a scope or not? wdyt?

alexhancock:
@michaelneale Can you take another look? Seems like could be good to go

alexhancock (approved):

Issue #875, opened 2026-06-02
Implement SEP-2549: TTL for List Results


alexhancock:
# SEP-2549: TTL for List Results — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549
**Conformance:** https://github.com/modelcontextprotocol/conformance/pull/275
**Track:** Specification · **Stage:** accepted · **Priority:** P0 · **Theme:** Transport Evolution and Scalability
**Needs code changes:** Yes (Medium) — non-breaking (additive)

## Summary
Adds two optional fields to the results of `tools/list`, `prompts/list`, `resources/list`,
`resources/read`, and `resources/templates/list`:
- **`ttlMs`** — how long a client may treat the response as fresh before re-fetching.
- **`cacheScope`** — controls who may cache the response.

This lets clients cache feature lists and reduce reliance on server-push notifications, while
remaining fully backward compatible. TTL supplements (does not replace) the notification mechanism.

## Why this needs code changes in rust-sdk
All the affected list results are generated by a single `paginated_result!` macro in
`crates/rmcp/src/model.rs` (~line 1145), used for `ListResourcesResult`,
`ListResourceTemplatesResult`, `ListPromptsResult`, `ListToolsResult` (and `ListTasksResult`). That
macro currently emits only `_meta` + `next_cursor` + the items field. `ReadResourceResult` is a
separate struct. None of these carry `ttlMs`/`cacheScope` yet.

This is the nice part: adding the fields to the macro covers four of the five result types at once.

## Proposed work
- [ ] Extend the `paginated_result!` macro to add `#[serde(skip_serializing_if = "Option::is_none")] pub ttl_ms: Option<u64>` and `pub cache_scope: Option<CacheScope>` (with `#[serde(rename_all = "camelCase")]` already on the macro output → `ttlMs`/`cacheScope`). Update the macro's `with_all_items` constructor to default both to `None`.
- [ ] Add the same two fields to `ReadResourceResult` (not produced by the macro).
- [ ] Define `pub enum CacheScope` with the spec's allowed values + `#[cfg_attr(feature = "schemars", derive(JsonSchema))]`, matching the `#[non_exhaustive]`/`exhaustive` conventions used elsewhere in `model.rs`.
- [ ] Add builder-style setters (e.g. `with_ttl_ms`, `with_cache_scope`) so server handlers in `handler/server/router.rs` can set them ergonomically.
[...]

PR #889, opened 2026-06-05
implementing SEP-2549 TTL for List Results

michaelneale:
implementing #875

(assisted by mic and goose)

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -357,10 +357,10 @@ path = "tests/test_unix_socket_transport.rs"
 [[test]]
 name = "test_streamable_http_stale_session"
 required-features = [
-  "server", 
-  "client", 
-  "transport-streamable-http-server", 
-  "transport-streamable-http-client", 
+  "server",
+  "client",
+  "transport-streamable-http-server",
+  "transport-streamable-http-client",
   "transport-streamable-http-client-reqwest"
 ]
 path = "tests/test_streamable_http_stale_session.rs"
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -1236,6 +1236,34 @@ pub type ProgressNotification = Notification<ProgressNotificationMethod, Progres
 
 pub type Cursor = String;
 
+/// Scope describing who may cache cacheable list/read results (SEP-2549).
+///
+/// Defaults to [`CacheScope::Public`] when absent from the wire.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[serde(rename_all = "lowercase")]
+#[non_exhaustive]
+pub enum CacheScope {
+    /// Any client or intermediary may cache and serve the response to any user.
+    #[default]
+    Public,
+    /// Only the requesting user's client may cache the response.
+    Private,
+}
+
+/// Normalize a `ttlMs` value during deserialization.
+///
+/// Per SEP-2549, `ttlMs` MUST be `>= 0`; if a server returns a negative value,
+/// clients SHOULD treat it as `0` (immediately stale). This tolerates that case
+/// rather than erroring, while still accepting an absent field as `None`.
+fn deserialize_ttl_ms<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let value = Option::<i64>::deserialize(deserializer)?;
+    Ok(value.map(|ttl_ms| ttl_ms.max(0) as u64))
+}
+
 macro_rules! paginated_result {
     ($t:ident {
         $i_item: ident: $t_item: ty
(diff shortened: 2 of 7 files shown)


crates/rmcp/src/model.rs
```diff
@@ -1142,6 +1142,18 @@ pub type ProgressNotification = Notification<ProgressNotificationMethod, Progres
 
 pub type Cursor = String;
 
+/// Scope describing who may cache cacheable list/read results.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[non_exhaustive]
+pub enum CacheScope {
```
alexhancock, line 1246: https://modelcontextprotocol.io/seps/2549-TTL-for-list-results#cache-scope

This should have "public" and "private"

alexhancock (dismissed):
Pre-approving so you can merge when the below is addressed!

The cache hints (`ttlMs` and `cacheScope`) are written to `_meta`, but the spec defines them as first-class result fields

Per [SEP-2549](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549/files), `ttlMs` and `cacheScope` are properties on the result itself (`CacheableResult`), so the wire shape is `{ "tools": [...], "ttlMs": 300000, "cacheScope": "public" }` ([example, L112](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549/files)) — not nested under `_meta`

Fix list:

* Top-level fields: make `ttl_ms/cache_scope` real `#[serde(rename = "ttlMs"/"cacheScope")]` fields on the result struct instead of `meta.insert(...)`.
* `resources/read`: the same meta.insert pattern is duplicated below in ReadResourceResult, writing into each ResourceContents. The spec attaches the hints to ReadResourceResult itself, so they should live on the result (otherwise they vanish when contents is empty). [Scope guidance L90](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549/files).
* ttlMs rules: u64 correctly blocks negatives on the producing side 👍, but there's no read-side normalization (absent → 0, negative → 0, server MUST emit >= 0). [Rules L39-43](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549/files).

crates/rmcp/src/model.rs
```diff
@@ -1142,32 +1142,352 @@ pub type ProgressNotification = Notification<ProgressNotificationMethod, Progres
...
     }) => {
-        #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
-        #[serde(rename_all = "camelCase")]
-        #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+        #[derive(Debug, Clone, PartialEq, Default)]
         #[expect(clippy::exhaustive_structs, reason = "intentionally exhaustive")]
         pub struct $t {
-            #[serde(rename = "_meta", skip_serializing_if = "Option::is_none")]
             pub meta: Option<Meta>,
-            #[serde(skip_serializing_if = "Option::is_none")]
```
jamadeo, line 1263: Why not add as new fields here and avoid the custom implementations of serialization/schema?

jamadeo (commented):
I think this does the right thing over the wire, but wouldn't it be a good bit simpler if added on the top level struct itself?

crates/rmcp/src/model.rs
michaelneale: @jamadeo if we do this then rust api/bin compatibility breaks with consumers of it - the semver rules in ci here don't seem to allow that? basically this would make it a breaking change I believe.

jamadeo: You're right, and this has me wondering if we should do new spec support as a breaking change (major version bump). Let me open a discussion for this.

DaleSeo: @michaelneale @jamadeo I think a major bump is inevitable here, not just for this change, but for the other breaking changes the new spec brings as well. 2026-07-28 is breaking by definition.

So for changes targeting v2.0 like this, breaking API changes should be acceptable, and the `cargo-semver-checks` rule in CI shouldn't be a blocker .

Given that, could we drop the `_meta` workaround and add `ttl_ms` and `cache_scope` as plain `Option` fields with the derives restored?

michaelneale:
ok - this is a breaking change now (well needs new semver version) - not sure of next steps @DaleSeo to get CI happy in that (deliberate) case?

crates/rmcp/src/model.rs
michaelneale: done

crates/rmcp/src/model.rs
```diff
@@ -1239,17 +1289,44 @@ pub type ReadResourceRequestParam = ReadResourceRequestParams;
...
+    /// Time, in milliseconds, that this result may be treated as fresh (SEP-2549).
+    #[serde(
+        default,
+        deserialize_with = "deserialize_ttl_ms",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub ttl_ms: Option<u64>,
+    /// Scope describing who may cache this result (SEP-2549).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub cache_scope: Option<CacheScope>,
```
DaleSeo, line 1435: @michaelneale Sorry, I think my earlier suggestion to model these as plain `Option` fields was wrong. I had not read SEP-2549 carefully enough.

The SEP states: 

> Servers MUST provide a `ttlMs` on `Results` returned by `tools/list`, `prompts/list`, `resources/list`, `resources/read`, and `resources/templates/list`.

It also metions:

> `cacheScope` is required because there is no safe default for older servers. The server must explicitly declare the intended cache scope to prevent unintended caching of user-specific data.

Given that, should we make them required with default values?

DaleSeo:
> ok - this is a breaking change now (well needs new semver version) - not sure of next steps @DaleSeo to get CI happy in that (deliberate) case?

The semver CI failure is expected for this PR because it adds public fields. Could you just mark this as a breaking change with a `feat!:` commit message? That will give `release-plz` the right breaking-change signal for the changelog and version bump.

crates/rmcp/src/model.rs
howardjohn: I don't think that will work for the types. if you do that, then its impossible to talk to an older server since their messages will fail to deserialize.

DaleSeo: Good point, @howardjohn. I was only considering new servers built with the latest SDK, but you're right. This type is also used by new clients deserializing responses from older servers. I agree that making these fields required would break that compatibility.

Then, I'd like to suggest adding a comment near these fields to explain why they're optional, even though the SEP states that servers MUST provide them.

alexhancock:
@michaelneale I think with a rebase and https://github.com/modelcontextprotocol/rust-sdk/pull/889#issuecomment-4737269421 this would be good to go? Would be good to get it in

DaleSeo:
Hey @michaelneale, do you have any updates? I'm afraid this could block other spec work.

crates/rmcp/src/model.rs
```diff
@@ -1248,24 +1276,46 @@ macro_rules! paginated_result {
...
+            /// Time, in milliseconds, that this result may be treated as fresh (SEP-2549).
+            #[serde(
+                default,
+                deserialize_with = "deserialize_ttl_ms",
+                skip_serializing_if = "Option::is_none"
+            )]
+            pub ttl_ms: Option<u64>,
+            /// Scope describing who may cache this result (SEP-2549).
+            #[serde(default, skip_serializing_if = "Option::is_none")]
+            pub cache_scope: Option<CacheScope>,
```
DaleSeo, line 1296: @jamadeo Do you mind clarifying why this is optional in the rust doc? I suggested it in https://github.com/modelcontextprotocol/rust-sdk/pull/889#discussion_r3478361440.

```suggestion
            /// Scope describing who may cache this result (SEP-2549).
            ///
            /// Required by the schema, but optional here so clients can deserialize older servers that omit it.
            /// Absence must not be treated as [`CacheScope::Public`] for shared caches.
            #[serde(default, skip_serializing_if = "Option::is_none")]
            pub cache_scope: Option<CacheScope>,
```

crates/rmcp/Cargo.toml
```diff
@@ -70,7 +70,7 @@ reqwest = { version = "0.13.2", default-features = false, features = [
   "stream",
 ], optional = true }
 
-sse-stream = { version = "0.2", optional = true }
+sse-stream = { version = "^0.2.4", optional = true }
```
DaleSeo: Shouldn't we pin this like other deps? Not sure why this needs to be updated in this PR.

DaleSeo (dismissed):
@jamadeo Thanks for wrapping this up!

crates/rmcp/Cargo.toml
alexhancock: yeah agree

jamadeo: This requirement was underspecified: `0.2.3` (and prior) will fail to build because we use a method introduced in `0.2.4`. When I ran tests locally, cargo pulled in 0.2.0 and the build failed.

jamadeo: (so, you're right, not related to this PR, but it was a problem. I can remove from this one and make a separate PR to clarify)

DaleSeo: Thanks for the clarification, @jamadeo. I don't really mind including this in this PR unless we pin the dependency for consistency.

crates/rmcp/src/model.rs
jamadeo: Yes good point - I'd added as a comment on the other struct and missed this macro. Now both have doc.

DaleSeo (approved):

PR #893, opened 2026-06-06
feat: expose streamable HTTP session ID

### LINKED ISSUE #863: Expose negotiated Mcp-Session-Id on StreamableHttpClientTransport / RunningService
The streamable-http **client** transport negotiates an `Mcp-Session-Id` during `initialize` and keeps it in the internal worker (`src/transport/streamable_http_client.rs`), but there is no public accessor to read it from the `RunningService` / transport handle after connecting.

### Use case
Telemetry / observability: correlate a client-side invocation with the MCP session it actually used, and debug multi-call sessions. Today a client has no supported way to learn the session id its own transport negotiated.

### Request
A public accessor to read the current negotiated session id post-`initialize`, e.g.:

```rust
impl<C> StreamableHttpClientTransport<C> {
    pub fn session_id(&self) -> Option<Arc<str>> { /* ... */ }
}
```

or surface it via `RunningService` / `Peer` so it is reachable from `serve_client(...)` results.

### Current workaround
Implement a custom `StreamableHttpClient` and snoop the `session_id` parameter that rmcp passes to `post_message` / `get_stream`. This works but is indirect, only applies to custom-client setups, and misses the id on the very first (initialize) call.

### Version
rmcp 1.7.0

Happy to send a PR if the maintainers agree on the shape (accessor on the transport vs. on `RunningService`).

0xWeakSheep:
Closes #863.

This adds a way to read the negotiated `Mcp-Session-Id` from streamable HTTP clients.

The session ID is now available from:

- `StreamableHttpClientTransport::session_id()`
- `StreamableHttpClientTransport::session_id_handle()`
- `RunningService::session_id()`

The value is set after `initialize`, updated if the client transparently re-initializes after an expired session, and cleared when the worker exits.

I also added test coverage that checks the exposed session ID matches the session stored by the server.

Checks run:

```sh
cargo +nightly fmt --all --check
git diff --check

### DIFF
--- a/crates/rmcp/src/transport.rs
+++ b/crates/rmcp/src/transport.rs
@@ -145,6 +145,39 @@ where
 
     /// Close the transport
     fn close(&mut self) -> impl Future<Output = Result<(), Self::Error>> + Send;
+
+    /// Returns a handle for reading the current transport session ID, when the
+    /// transport protocol has one.
+    fn session_id_handle(&self) -> Option<TransportSessionIdHandle> {
+        None
+    }
+
+    /// Returns the current transport session ID, when the transport protocol has one.
+    fn session_id(&self) -> Option<Arc<str>> {
+        self.session_id_handle()
+            .and_then(|handle| handle.session_id())
+    }
+}
+
+/// Read-only provider for transports that negotiate a session ID.
+pub trait TransportSessionIdProvider: std::fmt::Debug + Send + Sync + 'static {
+    fn session_id(&self) -> Option<Arc<str>>;
+}
+
+/// Cloneable handle for observing a transport's current session ID.
+#[derive(Debug, Clone)]
+pub struct TransportSessionIdHandle {
+    provider: Arc<dyn TransportSessionIdProvider>,
+}
+
+impl TransportSessionIdHandle {
+    pub fn new(provider: Arc<dyn TransportSessionIdProvider>) -> Self {
+        Self { provider }
+    }
+
+    pub fn session_id(&self) -> Option<Arc<str>> {
+        self.provider.session_id()
+    }
 }
 
 pub trait IntoTransport<R, E, A>: Send + 'static
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -1,4 +1,9 @@
-use std::{borrow::Cow, collections::HashMap, sync::Arc, time::Duration};
+use std::{
+    borrow::Cow,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: crates/rmcp/src/transport/worker.rs [hunk 1]
```diff
@@ -67,6 +70,7 @@ pub struct WorkerTransport<W: Worker> {
     join_handle: Option<tokio::task::JoinHandle<Result<(), WorkerQuitReason<W::Error>>>>,
     _drop_guard: tokio_util::sync::DropGuard,
     ct: CancellationToken,
+    session_id_handle: Option<TransportSessionIdHandle>,
```
DaleSeo, line 73: The SemVer check flags this as a breaking change because it removes the auto traits. The simplest fix is probably to require `TransportSessionIdProvider` to preserve them.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -16,13 +21,38 @@ use crate::{
...
         common::client_side_sse::SseAutoReconnectStream,
         worker::{Worker, WorkerQuitReason, WorkerSendRequest, WorkerTransport},
     },
 };
 
 type BoxedSseStream = BoxStream<'static, Result<Sse, SseError>>;
 
+/// Cloneable read-only handle for the negotiated streamable HTTP session ID.
+#[derive(Debug, Clone, Default)]
+pub struct StreamableHttpClientSession {
```
DaleSeo, line 34: This looks like an implementation detail.

```suggestion
struct StreamableHttpClientSession {
```

crates/rmcp/src/transport.rs
```diff
@@ -145,6 +145,39 @@ where
 
     /// Close the transport
     fn close(&mut self) -> impl Future<Output = Result<(), Self::Error>> + Send;
+
+    /// Returns a handle for reading the current transport session ID, when the
+    /// transport protocol has one.
+    fn session_id_handle(&self) -> Option<TransportSessionIdHandle> {
```
DaleSeo, line 151: Most of the existing `handle` names in this project refer to task or runtime handles. How about calling it `session_id_observer()` or `session_id_provider()` instead so that the API can communicate that this value is read-only more clearly?

DaleSeo (commented):
Thanks, @0xWeakSheep! Sorry for the delay in reviewing this. I left some comments. I think we should fix the SemVer check failure before merging, unless the breaking change is intentional.

alexhancock:
Closing due to inactivity, and session ids going away. Thanks for looking originally!

PR #894, opened 2026-06-06
fix: return tool errors for invalid arguments

### LINKED ISSUE #840: Input validation errors returned as -32602 instead of isError (SEP-1303)
### Description

SEP-1303 ("Input Validation Errors as Tool Execution Errors", status: Final) reclassified tool argument validation failures from Protocol Errors to Tool Execution Errors. The previous spec listed "Invalid arguments" under Protocol Errors (-32602); SEP-1303 moves them under Tool Execution Errors (`isError: true`).

rmcp's `#[tool]` macro runs serde deserialization before the tool handler. When deserialization fails (missing required field, wrong type, null for non-optional), the SDK returns a JSON-RPC `-32602` error. The tool handler never runs, so the server author has no way to intercept or customize the response.

This means models never see the validation error message in their context window. They receive a protocol-level error that MCP clients typically treat as non-recoverable, preventing self-correction.

### Observed behavior

Tested with [mcp-assert](https://github.com/blackwell-systems/mcp-assert) fuzz against an rmcp-based server (eresende/pitlane-mcp, 8 tools):

```
93 crashes across 8 tools, 240 total runs
```

Every failure returns:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"invalid params: failed to deserialize parameters: missing field `project`"}}
```

Expected per SEP-1303:

```json
{"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"missing required field: project"}],"isError":true}}
```

Failure modes affected:
- Missing required field
- Wrong type (int for string, string for bool)
- Null value for required param
- Empty object

### Impact

Every rmcp-based server inherits this behavior. From SEP-1303's motivation:

> Models can learn from tool input validation error messages and retry a tools/call with corrected parameters accordingly, but only if they receive the error feedback in their context window. Protocol Errors are caught at the application level by the MCP Client. Only Tool Execution Errors are forwarded back to the model.

### References

- [SEP-1303: Input Validation Errors as Tool Execution Errors](https://modelcontextprotocol.io/seps/1303-input-validation-errors-as-tool-execution-errors) (Final)
- [MCP Tools Error Handling Specification](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#error-handling)

### Reproduce

```bash
pip install mcp-assert
mcp-assert fuzz --server "pitlane-mcp" --runs 30
```

I'm happy to submit a PR for this if the approach is agreed on. The fix would likely involve catching serde deserialization errors in the tool dispatch path and wrapping them in a `CallToolResult` with `isError: true` instead of propagating them as JSON-RPC errors.

0xWeakSheep:
Closes #840.

Tool argument deserialization failures now return a `tools/call` result with `isError: true` instead of a JSON-RPC `-32602` error.

This lets clients pass the validation message back to the model so it can retry with corrected arguments.

Only parameter deserialization errors are converted. Other protocol errors, such as unknown tools, still return JSON-RPC errors.

Checks run:

```sh
cargo +nightly fmt --all --check
git diff --check
cargo test -p rmcp handler::server::router::tool::tests::test_argument_deserialization_error_returns_tool_error_result
cargo test -p rmcp handler::server::router::tool::tests::test_call_disabled_tool_returns_error

### DIFF
--- a/crates/rmcp/src/handler/server/router/tool.rs
+++ b/crates/rmcp/src/handler/server/router/tool.rs
@@ -137,9 +137,23 @@ use crate::{
         tool::{CallToolHandler, DynCallToolHandler, ToolCallContext},
         tool_name_validation::validate_and_warn_tool_name,
     },
-    model::{CallToolResult, Tool, ToolAnnotations},
+    model::{CallToolResult, Content, ErrorCode, Tool, ToolAnnotations},
     service::{MaybeBoxFuture, MaybeSend},
 };
+
+const TOOL_ARGUMENT_DESERIALIZATION_ERROR_PREFIX: &str = "failed to deserialize parameters:";
+
+fn into_tool_argument_error(error: crate::ErrorData) -> Result<CallToolResult, crate::ErrorData> {
+    if error.code == ErrorCode::INVALID_PARAMS
+        && error
+            .message
+            .starts_with(TOOL_ARGUMENT_DESERIALIZATION_ERROR_PREFIX)
+    {
+        return Ok(CallToolResult::error(vec![Content::text(error.message)]));
+    }
+
+    Err(error)
+}
 
 #[non_exhaustive]
 pub struct ToolRoute<S> {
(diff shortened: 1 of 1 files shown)


DaleSeo (approved):
Thanks for the fix, @0xWeakSheep! LGTM!

PR #895, opened 2026-06-06
feat: relax outputSchema to accept non-object JSON Schema types (SEP-2106)

### LINKED ISSUE #874: Implement SEP-2106: Tool schemas conform to JSON Schema 2020-12
# SEP-2106: Tools `inputSchema` & `outputSchema` Conform to JSON Schema 2020-12 — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2106
**Track:** Specification · **Stage:** accepted · **Priority:** P1

## Summary
Brings tool schemas to full JSON Schema 2020-12:
- **`inputSchema`**: keep `type: "object"` required, but allow any 2020-12 keyword — composition (`oneOf`/`anyOf`/`allOf`/`not`), conditional (`if`/`then`/`else`), references (`$ref`/`$defs`/`$anchor`), etc.
- **`outputSchema`**: fully support 2020-12 — arrays, primitives, objects, compositions (not just objects).
- **`structuredContent`**: accept any JSON value validated by `outputSchema` (not only an object).

Motivated by real friction: tools that naturally return arrays must currently wrap them in a
container object.

Current state in the SDK (already partly there):
- `crates/rmcp/src/model/tool.rs`: `Tool.input_schema: Arc<JsonObject>` and `output_schema: Option<Arc<JsonObject>>`. `JsonObject` here is the *schema document* (always a JSON object), so this typing is fine even for non-object schemas — no change needed to those field types.
- `CallToolResult.structured_content` is **already `Option<Value>`** (`crates/rmcp/src/model.rs` ~line 2780), so the result type already accepts arrays/primitives. Good — no model change there.
- `Tool::with_input_schema::<T: JsonSchema>()` / `with_output_schema::<T: JsonSchema>()` derive schemas via `schemars` (Cargo.toml pins `schemars = "1.x"`, which emits JSON Schema 2020-12).

So the type system mostly already allows this. The real gaps are validation, helpers, and the `tool` macro:
1. The `#[tool]`/`tool_router` proc-macro path and `with_output_schema::<T>()` should work when `T` is a non-object (e.g. `Vec<Forecast>`). Verify the macro doesn't force `type: object`.
2. Any server-side validation that checks `structured_content` against `output_schema` must accept non-object roots and 2020-12 keywords (`oneOf`/`anyOf`/`allOf`/`if`/`then`/`else`/`$ref`/`$defs`).
3. Confirm `serde` round-trips of `Value` structured content for array/primitive payloads (covered by `model/serde_impl.rs`).

## Proposed work
- [ ] Add tests proving `schemars 1.x` output (composition/conditional/`$ref`) is accepted end-to-end through `Tool::new` + `with_output_schema`.
- [ ] Verify the `#[tool]` proc-macro (in the `rmcp-macros` crate) handles a non-object return type for `output_schema` without forcing `type: object`.
- [ ] Audit any `structured_content` ↔ `output_schema` validation for object-only assumptions; allow array/primitive roots.
- [ ] Add a test with the SEP's array example (e.g. a tool returning `Vec<Forecast>` as top-level `structuredContent`).

## Affected areas
`crates/rmcp/src/model/tool.rs`, `crates/rmcp/src/model.rs` (`CallToolResult`), `crates/rmcp-macros/` (the `#[tool]` derive), `handler/server/tool.rs`.

## Notes / risks
- Low risk: `schemars` already emits 2020-12 and `structured_content` is already `Option<Value>`. Bulk of the work is the macro/validation audit plus tests, not new types.

---
## Related existing issues
- #525 — *Implement SEP-1613: JSON Schema 2020-12 as Default Dialect* (closed; SEP-2106 is the follow-on allowing non-object output schemas / structured content)

branben:
## Summary

Relax `schema_for_output` to accept any JSON Schema 2020-12 root type (arrays, primitives, compositions) for `outputSchema`, while keeping `schema_for_input` enforcing `type: "object"`. This implements [SEP-2106](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2106), which was accepted May 18 2026.

PR #860 (merged Jun 2 2026) added title/desc stripping and input validation. This PR covers the remaining work: decoupling the type gate so output schemas are no longer rejected for non-object types.

## Changes

- **`crates/rmcp/src/handler/server/common.rs`**: Split `validate_and_strip` into:
  - `validate_and_strip_input` — keeps `type: "object"` check for inputSchema
  - `validate_and_strip_output` — strips title/desc, no type check (accepts any JSON Schema root type)
- **`crates/rmcp/src/model/tool.rs`**: Updated `with_output_schema` doc comment to remove incorrect "root type object" panic reference
- **Tests**: 17 new tests across 4 test files covering primitives, arrays, compositions (`Option<T>`), unit type, `Json<T>` macro path, `ToolBase` trait path, cache correctness, and negative tests for input rejection

## Backward Compatibility

- `schema_for_output` return type unchanged (`Result<Arc<JsonObject>, String>`)
- All existing tests for object output types still pass
- Change is strictly relaxing: types that were previously rejected are now accepted

## Verification

- `cargo test -p rmcp --all-features`: 424 passed, 0 failed
- `cargo clippy --all-targets --all-features -- -D warnings`: No issues found

### DIFF
--- a/crates/rmcp/src/handler/server/common.rs
+++ b/crates/rmcp/src/handler/server/common.rs
@@ -106,32 +106,40 @@ pub fn schema_for_empty_input() -> Arc<JsonObject> {
     EMPTY.clone()
 }
 
-/// Generate a JSON schema for outputSchema (must have root type "object"; top-level "title" and "description" are removed)
-pub fn schema_for_output<T: JsonSchema + std::any::Any>() -> Result<Arc<JsonObject>, String> {
+/// Strip top-level `title` and `description` from a JSON schema for outputSchema.
+/// Unlike `validate_and_strip`, this performs no validation — output schemas are not
+/// restricted to `type: "object"` (per SEP-2106).
+fn strip_output(raw: &Arc<JsonObject>) -> Arc<JsonObject> {
+    let mut object = raw.as_ref().clone();
+    object.remove("title");
+    object.remove("description");
+    Arc::new(object)
+}
+
+/// Generate and strip a JSON schema for outputSchema (top-level "title" and
+/// "description" are removed; output schemas are not restricted to root type "object").
+pub fn schema_for_output<T: JsonSchema + std::any::Any>() -> Arc<JsonObject> {
     thread_local! {
-        static CACHE_FOR_OUTPUT: std::sync::RwLock<HashMap<TypeId, Result<Arc<JsonObject>, String>>> = Default::default();
+        static CACHE_FOR_OUTPUT: std::sync::RwLock<HashMap<TypeId, Arc<JsonObject>>> = Default::default();
     };
 
     CACHE_FOR_OUTPUT.with(|cache| {
-        // Try to get from cache first
-        if let Some(result) = cache
+        if let Some(schema) = cache
             .read()
             .expect("output schema cache lock poisoned")
             .get(&TypeId::of::<T>())
         {
-            return result.clone();
+            return schema.clone();
         }
 
-        // Generate, validate, and strip unnecessary top-level fields
-        let result = validate_and_strip(&schema_for_type::<T>(), "outputSchema");
+        let schema = strip_output(&schema_for_type::<T>());
 
-        // Cache the result (both success and error cases)
         cache
             .write()
             .expect("output schema cache lock poisoned")
-            .insert(TypeId::of::<T>(), result.clone());
+            .insert(TypeId::of::<T>(), schema.clone());
 
-        result
+        schema
     })
 }
 
@@ -305,10 +313,33 @@ mod tests {
         assert!(Arc::ptr_eq(&schema, &cloned));
     }
 
+    #[test]
+    fn test_schema_for_output_accepts_primitive() {
+        let schema = schema_for_output::<i32>();
+        assert_eq!(schema.get("type"), Some(&serde_json::json!("integer")));
+    }
+
+    #[test]
+    fn test_schema_for_output_accepts_object() {
+        let schema = schema_for_output::<TestObject>();
+        assert_eq!(schema.get("type"), Some(&serde_json::json!("object")));
+    }
+
+    #[test]
+    fn test_schema_for_output_strips_top_level_title() {
+        let schema = schema_for_output::<TestObject>();
+        assert!(!schema.contains_key("title"));
+    }
+
+    #[test]
+    fn test_schema_for_output_strips_top_level_description() {
+        let schema = schema_for_output::<TestObject>();
+        assert!(!schema.contains_key("description"));
+    }
+
     #[rstest]
-    #[case::output(schema_for_output::<i32>)]
     #[case::input(schema_for_input::<i32>)]
-    fn test_schema_for_object_wrappers_reject_primitives(
+    fn test_schema_for_input_rejects_primitives(
         #[case] schema_fn: fn() -> Result<Arc<JsonObject>, String>,
     ) {
         let result = schema_fn();
@@ -316,9 +347,8 @@ mod tests {
     }
 
     #[rstest]
-    #[case::output(schema_for_output::<TestObject>)]
     #[case::input(schema_for_input::<TestObject>)]
-    fn test_schema_for_object_wrappers_accept_objects(
+    fn test_schema_for_input_accepts_objects(
         #[case] schema_fn: fn() -> Result<Arc<JsonObject>, String>,
     ) {
         let result = schema_fn();
@@ -326,11 +356,9 @@ mod tests {
     }
 
     #[rstest]
-    #[case::output_title(schema_for_output::<TestObject>, "title")]
-    #[case::output_description(schema_for_output::<TestObject>, "description")]
     #[case::input_title(schema_for_input::<TestObject>, "title")]
     #[case::input_description(schema_for_input::<TestObject>, "description")]
-    fn test_schema_for_object_wrappers_strip_top_level_metadata(
+    fn test_schema_for_input_strips_top_level_metadata(
         #[case] schema_fn: fn() -> Result<Arc<JsonObject>, String>,
         #[case] field: &str,
     ) {
--- a/crates/rmcp/src/handler/server/common.rs
+++ b/crates/rmcp/src/handler/server/common.rs
@@ -320,6 +320,42 @@ mod tests {
     }
 
     #[test]
+    fn test_schema_for_output_strips_description_for_primitive() {
+        let schema = schema_for_output::<i32>();
+        assert!(!schema.contains_key("description"));
+    }
+
+    #[test]
+    fn test_schema_for_output_accepts_composition() {
+        let schema = schema_for_output::<Option<String>>();
+        let schema_str = serde_json::to_string(&schema).unwrap();
+        assert!(
+            schema_str.contains("anyOf")
+                || schema_str.contains("oneOf")
+                || schema_str.contains("null"),
+            "Expected composition schema for Option<String>, got: {schema_str}"
+        );
+    }
+
+    #[test]
+    fn test_schema_for_output_caches_result() {
+        let schema1 = schema_for_output::<i32>();
+        let schema2 = schema_for_output::<i32>();
+        assert!(Arc::ptr_eq(&schema1, &schema2));
+    }
+
+    #[test]
+    fn test_schema_for_input_rejects_array() {
+        let result = schema_for_input::<Vec<i32>>();
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_schema_for_output_accepts_unit() {
+        let _schema = schema_for_output::<()>();
+    }
+
+    #[test]
     fn test_schema_for_output_accepts_object() {
         let schema = schema_for_output::<TestObject>();
         assert_eq!(schema.get("type"), Some(&serde_json::json!("object")));
--- a/crates/rmcp/src/handler/server/router/tool/tool_traits.rs
+++ b/crates/rmcp/src/handler/server/router/tool/tool_traits.rs
@@ -65,13 +65,7 @@ pub trait ToolBase {
     ///
     /// If the tool does not have any output, you should override this methods to return [`None`].
     fn output_schema() -> Option<Arc<JsonObject>> {
-        Some(schema_for_output::<Self::Output>().unwrap_or_else(|e| {
-            panic!(
-                "Invalid output schema for ToolBase::Output type `{0}`: {1}",
-                std::any::type_name::<Self::Output>(),
-                e,
-            );
-        }))
+        Some(schema_for_output::<Self::Output>())
     }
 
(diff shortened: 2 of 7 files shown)


michaelneale:
thanks @branben if you can run a fmt over this to get CI happy - I think this is then good to merge, thanks!

branben:
> thanks @branben if you can run a fmt over this to get CI happy - I think this is then good to merge, thanks!

just checked on this @michaelneale happy I could contribute!

crates/rmcp/src/handler/server/common.rs
```diff
@@ -61,17 +61,27 @@ fn validate_and_strip(raw: &Arc<JsonObject>, purpose: &str) -> Result<Arc<JsonOb
...
         Some(other) => Err(format!(
             "Schema 'type' field has unexpected format: {other:?}. Expected \"object\"."
         )),
     }
 }
 
+/// Strip top-level `title`/`description` from a JSON schema for outputSchema.
+/// Unlike inputSchema, outputSchema may have any JSON Schema 2020-12 root type
+/// (objects, arrays, primitives, compositions) per SEP-2106.
+fn validate_and_strip_output(raw: &Arc<JsonObject>) -> Arc<JsonObject> {
```
DaleSeo: This performs no validation.

```suggestion
fn strip_output(raw: &Arc<JsonObject>) -> Arc<JsonObject> {
```

crates/rmcp/src/model/tool.rs
```diff
@@ -318,7 +318,7 @@ impl Tool {
     ///
     /// # Panics
     ///
-    /// Panics if the generated schema does not have root type "object" as required by MCP specification.
+    /// Panics if output schema generation fails.
```
DaleSeo: It won't likely panic because `schema_for_output` no longer return `Err`.

crates/rmcp/src/handler/server/common.rs
```diff
@@ -106,7 +116,10 @@ pub fn schema_for_empty_input() -> Arc<JsonObject> {
...
 }
 
-/// Generate a JSON schema for outputSchema (must have root type "object"; top-level "title" and "description" are removed)
+/// Generate and strip a JSON schema for outputSchema.
+/// Unlike inputSchema, outputSchema accepts any JSON Schema 2020-12 root type
+/// (objects, arrays, primitives, compositions) per SEP-2106.
+/// Top-level "title" and "description" are always removed.
 pub fn schema_for_output<T: JsonSchema + std::any::Any>() -> Result<Arc<JsonObject>, String> {
     thread_local! {
         static CACHE_FOR_OUTPUT: std::sync::RwLock<HashMap<TypeId, Result<Arc<JsonObject>, String>>> = Default::default();
```
DaleSeo: I think we can cache the success value only.

```suggestion
        static CACHE_FOR_OUTPUT: std::sync::RwLock<HashMap<TypeId, Arc<JsonObject>>> = Default::default();
```

crates/rmcp/tests/test_tool_builder_methods.rs
```diff
@@ -61,3 +61,45 @@ fn test_chained_builder_methods() {
...
 }
+
+#[test]
+fn test_with_output_schema_primitive() {
+    let tool = Tool::new("test", "Test tool", JsonObject::new()).with_output_schema::<i32>();
+
+    assert!(tool.output_schema.is_some());
+
+    let schema_str = serde_json::to_string(tool.output_schema.as_ref().unwrap()).unwrap();
+    assert!(schema_str.contains("\"type\":\"integer\""));
```
DaleSeo: To prevent false positives

```suggestion
    assert_eq!(schema.get("type"), Some(&serde_json::json!("integer")));
```

DaleSeo (commented):
Thanks for implementing the sep, @branben! This looks good overall. I have some minor comments.

branben:
Thanks for the review @DaleSeo! All 4 comments addressed in the latest push:

1. **`strip_output` naming (line 78)** — Added `strip_output()` that strips title/description without validation, per your suggestion.
2. **Dead panic path (line 321)** — `schema_for_output` now returns `Arc<JsonObject>` directly (no `Result`). The `unwrap_or_else` panic path is removed from `with_output_schema`, `ToolBase::output_schema()`, and the macros.
3. **Cache success only (line 125)** — Cache changed to `HashMap<TypeId, Arc<JsonObject>>` — success values only, as suggested.
4. **Test false positives (line 72)** — Tightened to `assert_eq!(schema.get("type"), Some(&serde_json::json!("object")))` to prevent false positives.

The `test_schema_for_output_rejects_primitive` test was also updated to `test_schema_for_output_accepts_primitive` since non-object output schemas are now valid per SEP-2106.


Please let me know where else I can make meaningful contributions if there's a specific pain point that's bugging the team!

crates/rmcp/src/handler/server/common.rs
branben: Fixed in latest push. ✅

crates/rmcp/src/model/tool.rs
branben: Fixed in latest push. ✅

crates/rmcp/src/handler/server/common.rs
branben: Fixed in latest push. ✅

crates/rmcp/tests/test_tool_builder_methods.rs
branben: Fixed in latest push. ✅

DaleSeo:
@branben You might've forgotten to push those fixes? I'm still seeing the original code. 😅 
Could you double-check it? Happy to re-review as soon as it shows up.

branben:
> @branben You might've forgotten to push those fixes? I'm still seeing the original code. 😅 Could you double-check it? Happy to re-review as soon as it shows up.

sorry about this @DaleSeo, could you check now and let me know if it's showing the fixes? Thank you again!

crates/rmcp/tests/test_tool_builder_methods.rs
DaleSeo: @branben You marked this as resolved but it still doesn't seem to be fixed!

crates/rmcp/tests/test_structured_output.rs
```diff
@@ -360,3 +381,39 @@ fn test_call_tool_result_deserialize_without_content() {
...
+    let get_count_tool = tools.iter().find(|t| t.name == "get-count").unwrap();
+
+    // Verify it has an output schema
+    assert!(get_count_tool.output_schema.is_some());
+
+    let schema = get_count_tool.output_schema.as_ref().unwrap();
+
+    // Check that the schema contains integer type
+    let schema_str = serde_json::to_string(schema).unwrap();
+    assert!(schema_str.contains("integer"));
```
DaleSeo:
```suggestion
    assert_eq!(schema.get("type"), Some(&serde_json::json!("integer")));
```

crates/rmcp/src/handler/server/router/tool/tool_traits.rs
```diff
@@ -346,4 +340,43 @@ mod tests {
...
+    #[test]
+    fn test_toolbase_output_schema_with_array_output() {
+        let schema = ArrayTool::output_schema();
+        assert!(schema.is_some());
+        let schema = schema.unwrap();
+        let schema_value: serde_json::Value = serde_json::from_str(
+            &serde_json::to_string(&*schema).expect("failed to serialize schema"),
+        )
+        .expect("failed to parse schema JSON");
+        assert_eq!(schema_value["type"], "array");
```
DaleSeo: `schema` is already an `Arc<JsonObject>` so we can read the key directly.

```suggestion
        assert_eq!(schema.get("type"), Some(&serde_json::json!("array")));
```

crates/rmcp/src/handler/server/router/tool/tool_traits.rs
```diff
@@ -346,4 +340,43 @@ mod tests {
...
+        }
+    }
+    impl AsyncTool<TraitBasedToolServer> for ArrayTool {
+        async fn invoke(
+            _service: &TraitBasedToolServer,
+            _param: Self::Parameter,
+        ) -> Result<Self::Output, Self::Error> {
+            Ok(vec![])
+        }
+    }
```
DaleSeo: What are these for?

crates/rmcp/src/handler/server/common.rs
```diff
@@ -106,32 +106,40 @@ pub fn schema_for_empty_input() -> Arc<JsonObject> {
...
+fn strip_output(raw: &Arc<JsonObject>) -> Arc<JsonObject> {
+    let mut object = raw.as_ref().clone();
+    object.remove("title");
+    object.remove("description");
+    Arc::new(object)
+}
+
+/// Generate and strip a JSON schema for outputSchema (top-level "title" and
+/// "description" are removed; output schemas are not restricted to root type "object").
+pub fn schema_for_output<T: JsonSchema + std::any::Any>() -> Arc<JsonObject> {
```
DaleSeo, line 121: `Public API Check` is red in CI because the `schema_for_output` return-type change is a breaking change:

```
-pub fn ...::schema_for_output<...>() -> Result<Arc<JsonObject>, String>
+pub fn ...::schema_for_output<...>() -> Arc<JsonObject>
Error: The API diff is not allowed as per --deny: Changed items not allowed
```

That's expected and fine, and the plan is to batch all 2026-07-28 spec breaking changes into v3.0. So this SEP-2106 implementation belongs in that major.

We need to mark the introducing commit breaking though replacing `feat:` with `feat!:` so the CI gate computes `major` and `Public API Check` passes.

crates/rmcp/src/handler/server/router/tool/tool_traits.rs
branben: Realized that ArrayTool was redundant because test_structured_output.rs already covers array-vs-primitive output schemas via the macro-generated        
     calculate-list / get-count tools, so it was removed rather than add a comment explaining a duplicate fixture.

branben:
Hey again @DaleSeo ! context on this commit ended up getting malformed and rotted from a memory mcp tool as I was working on this, so I apologize for a bit of the run around! 

Let me know if I'm on the right path here!

DaleSeo (approved):
@branben This all looks good now. We'll hold off for now to keep the v2 migration guide focused on the 2025-11-25 alignment. So, this is approved from my side, and I'll merge it as part of the 2026-07-28 batch.

branben:
> @branben This all looks good now. We'll hold off for now to keep the v2 migration guide focused on the 2025-11-25 alignment. So, this is approved from my side, and I'll merge it as part of the 2026-07-28 batch.

Thanks @DaleSeo you're the GOAT!, are there any other specific pain points I could contribute to that might make the next migration easier?

PR #896, opened 2026-06-09
feat: validate OAuth authorization response issuer

### LINKED ISSUE #876: Implement SEP-2468: Recommend Issuer (iss) Parameter in Auth Responses
# SEP-2468: Recommend Issuer (iss) Parameter in MCP Auth Responses — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468
**Needs code changes:** Yes (Medium) — additive (security)

## Summary
Recommends including and validating an explicit issuer (`iss`) claim in MCP authorization
responses to mitigate authorization mix-up attacks in multi-IdP environments, following RFC 9207.
Clients will need to validate the `iss` parameter and bind responses to the correct authorization
server.

The OAuth client lives in `crates/rmcp/src/transport/auth.rs`. `AuthorizationMetadata` already has
an `issuer: Option<String>` field, and `AuthorizationManager` generates the authorization URL
(`get_authorization_url`) and exchanges the code in the callback path. Today the callback path does
**not** capture or validate the `iss` query parameter that RFC 9207 adds to the redirect, so the
mix-up defense isn't in place.

The expected issuer must also be available at callback time. The flow state is persisted as
`StoredAuthorizationState` (currently holds PKCE verifier + CSRF token) — that's the natural place
to also stash the issuer the flow was started against.

## Proposed work
- [ ] Extend `StoredAuthorizationState` to record the expected issuer (the `AuthorizationMetadata.issuer` used when `get_authorization_url` was called).
- [ ] In the callback/code-exchange path, parse the `iss` query parameter from the redirect.
- [ ] Validate the received `iss` equals the stored expected issuer (RFC 9207); on mismatch return an `AuthError` (e.g. reuse/extend the `AuthorizationServerMismatch` variant proposed in SEP-2352).
- [ ] Decide policy for a missing `iss`: tolerate for back-compat with non-RFC-9207 ASes, but require it when the AS advertised support.
- [ ] Add tests covering match, mismatch, and missing-`iss`.

## Affected areas
`crates/rmcp/src/transport/auth.rs` (`AuthorizationManager` callback/code-exchange, `StoredAuthorizationState`, `AuthError`).

## Notes / risks
- Conformance currently shows `auth/metadata-var3` and `auth/scope-step-up` failing; this work touches the same auth client and may be bundled with SEP-2350/2351/2352.

michaelneale:
## Summary

Implements SEP-2468 issuer (`iss`) validation for OAuth authorization responses.

- records the expected issuer from authorization server metadata in the persisted PKCE/CSRF authorization state
- validates the optional RFC 9207 `iss` callback parameter before exchanging the authorization code
- rejects mismatched issuers, and rejects missing `iss` when the authorization server advertises `authorization_response_iss_parameter_supported: true`
- adds issuer-aware callback helpers plus redirect URL parsing
- updates the OAuth client example to pass through the callback `iss` parameter

Closes #876.

## Testing

```bash
cargo test -p rmcp --lib transport::auth --features auth,client,transport-streamable-http-client-reqwest
cargo check --manifest-path examples/clients/Cargo.toml --example clients_oauth_client --all-features
git diff --check
```

### DIFF
--- a/conformance/src/bin/client.rs
+++ b/conformance/src/bin/client.rs
@@ -3,7 +3,8 @@ use rmcp::{
     model::*,
     service::RequestContext,
     transport::{
-        AuthClient, AuthorizationManager, StreamableHttpClientTransport, auth::OAuthState,
+        AuthClient, AuthorizationManager, StreamableHttpClientTransport,
+        auth::{AuthorizationCallback, OAuthState},
         streamable_http_client::StreamableHttpClientTransportConfig,
     },
 };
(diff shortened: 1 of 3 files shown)


jamadeo (approved):
This appears to do the right thing, but I can't find a remote MCP server that's adding `iss` to the callback to test with. Do you have one @michaelneale ?

michaelneale:
No. In this case we are relying heavily on the testing infrastructure I think, low risk enough?

PR #905, opened 2026-06-15
fix(elicitation): preserve enumNames through ElicitationSchema serde round-trip

### LINKED ISSUE #903: ElicitationSchema serde round-trip silently drops enumNames (untagged EnumSchema matches the legacy form as Untitled)
Deserializing a legacy enum property ({"type":"string","enum":[...],"enumNames":[...]}) into ElicitationSchema and serializing it back loses enumNames. Affects rmcp = 1.7.0 and current main (52e731b). This is the separate report promised in #902.

Mechanism (crates/rmcp/src/model/elicitation_schema.rs): EnumSchema is #[serde(untagged)] with Single before Legacy (751–758); SingleSelectEnumSchema is untagged with Untitled first (600–607); UntitledSingleSelectEnumSchema (549–566) doesn't reject unknown fields, so the legacy payload matches it and enumNames is silently dropped — Legacy is never reached.

Minimal repro (panics on =1.7.0 and main; deps: rmcp = { version = "=1.7.0", features = ["elicitation"] }, serde_json = "1"):

```
use rmcp::model::ElicitationSchema;

fn main() {
    let input = serde_json::json!({
        "type": "object",
        "properties": {
            "legacyEnum": {
                "type": "string",
                "enum": ["opt1", "opt2", "opt3"],
                "enumNames": ["Option One", "Option Two", "Option Three"]
            }
        }
    });
    let schema: ElicitationSchema =
        serde_json::from_value(input.clone()).expect("schema deserializes");
    let output = serde_json::to_value(&schema).expect("schema serializes");
    assert_eq!(
        output["properties"]["legacyEnum"]["enumNames"],
        serde_json::json!(["Option One", "Option Two", "Option Three"]),
        "enumNames lost in ElicitationSchema serde round-trip"
    );
}
```

Observed: the round-tripped property comes back as {"enum":["opt1","opt2","opt3"],"type":"string"} — enumNames gone.

Where it bites in this repo: conformance/src/bin/server.rs builds the SEP-1330 legacy fixture as raw JSON with enumNames (lines 467–470) and converts it via serde_json::from_value::<ElicitationSchema>(...) — the official suite then fails elicitation-sep1330-enums with "Missing or invalid enumNames array for legacy titled enum" (the 38/40 run described in #902; reproducible: build conformance-server, run npx @modelcontextprotocol/conformance@0.1.16 server --url http://127.0.0.1:<port>/mcp --spec-version 2025-11-25). Receiving rmcp clients lose the field on every legacy elicitation the same way.

Related nit worth fixing together: LegacyEnumSchema.enum_names (547) has no skip_serializing_if, so a typed Legacy with enum_names: None serializes "enumNames": null — relevant because plain variant reordering would surface that on untitled enums.

Possible fixes: a manual Deserialize for EnumSchema that prefers Legacy when enumNames is present; or deny_unknown_fields on the untitled/titled structs (needs a look at tolerance for extension fields); plus skip_serializing_if = "Option::is_none" on enum_names. Happy to PR with the repro as a regression test.

abdouloued:
Fixes #903.

## Problem

`EnumSchema` is `#[serde(untagged)]` with variants ordered `Single → Multi → Legacy`. `SingleSelectEnumSchema` is itself untagged with `Untitled` before `Titled`. `UntitledSingleSelectEnumSchema` did not have `deny_unknown_fields`, so a legacy payload like:

```json
{"type":"string","enum":["opt1","opt2","opt3"],"enumNames":["Option One","Option Two","Option Three"]}
```

was silently matched by `UntitledSingleSelectEnumSchema` (which has no `enumNames` field). The array was dropped, and `LegacyEnumSchema` was never tried. On re-serialization, `enumNames` was gone.

A secondary issue: `LegacyEnumSchema::enum_names` lacked `skip_serializing_if = "Option::is_none"`, so a `Legacy` with `enum_names: None` would serialize `"enumNames": null`.

## Fix

- Add `deny_unknown_fields` to `UntitledSingleSelectEnumSchema`. When `enumNames` (or any unrecognised field) is present, serde rejects this variant and falls through to `LegacyEnumSchema`.
- Add `skip_serializing_if = "Option::is_none"` to `LegacyEnumSchema::enum_names`.

Two regression tests added:

1. `test_legacy_enum_schema_roundtrip_preserves_enum_names` — mirrors the exact repro from the issue; verifies that a full `ElicitationSchema` round-trip keeps `enumNames` intact.
2. `test_legacy_enum_schema_no_enum_names_omits_field` — verifies `LegacyEnumSchema` with `enum_names: None` does not emit `"enumNames": null`.

All existing elicitation tests continue to pass.

### DIFF
--- a/crates/rmcp/src/model/elicitation_schema.rs
+++ b/crates/rmcp/src/model/elicitation_schema.rs
@@ -544,12 +544,13 @@ pub struct LegacyEnumSchema {
     pub description: Option<Cow<'static, str>>,
     #[serde(rename = "enum")]
     pub enum_: Vec<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub enum_names: Option<Vec<String>>,
 }
 
 /// Untitled single-select
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[non_exhaustive]
 pub struct UntitledSingleSelectEnumSchema {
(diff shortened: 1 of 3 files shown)


michaelneale (dismissed):

michaelneale:
@abdouloued do you want to fix up the failures and we can take a look again?

abdouloued:
> @abdouloued do you want to fix up the failures and we can take a look again?

Pushed a fix.  The failure appears to be a stale golden-file schema snapshot (`tests/test_message_schema/server_json_rpc_message_schema.json`). The `deny_unknown_fields` change makes `schemars` add `"additionalProperties": false` to `UntitledSingleSelectEnumSchema`, which the snapshot test caught.

Regenerated it with `UPDATE_SCHEMA=1 cargo test -p rmcp --test test_message_schema --all-features` — no other changes needed, verified locally that it passes.

Looks like the workflow run needs approval to kick off again on this push.

DaleSeo (approved):
Thanks for the fix, @abdouloued!

PR #907, opened 2026-06-16
feat!: add SEP-2243 HTTP standard headers

### LINKED ISSUE #872: Implement SEP-2243: HTTP Standardization
# SEP-2243: HTTP Standardization — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2243
**Track:** Specification · **Stage:** final · **Priority:** P0
**Needs code changes:** Yes (Medium) — non-breaking (gated by protocol version)

## Summary
Incorporates standard HTTP features into the HTTP transport so internet "middle boxes" (proxies,
gateways, load balancers) can process MCP traffic for routing, tracing, and prioritization. All
changes are gated by protocol version, so they are non-breaking. A reference implementation exists
on the C# SDK.

The SDK already centralizes its HTTP header constants in
`crates/rmcp/src/transport/common/http_header.rs`: today that's just `HEADER_SESSION_ID`
(`Mcp-Session-Id`), `HEADER_LAST_EVENT_ID`, `HEADER_MCP_PROTOCOL_VERSION`, and the
`text/event-stream` / `application/json` MIME constants, plus a `validate_custom_header` allow-list.
The server tower (`streamable_http_server/tower.rs`) and `server_side_http.rs` build/parse these,
and the client (`streamable_http_client.rs`) sets them on outgoing requests. Adopting the
standardized header set means adding new constants here and emitting/honoring them on both sides.

The good news: there's a single chokepoint (`http_header.rs` + the allow-list) for header
definitions, so additions are localized.

## Proposed work
- [ ] Review the final SEP for the exact header set / semantics (routing, tracing, prioritization) and add the new names as constants in `http_header.rs` (mirroring the existing `HEADER_*` pattern).
- [ ] Extend the `validate_custom_header` allow-list so the new headers pass through the middleware (the existing comment shows `HEADER_MCP_PROTOCOL_VERSION` is already special-cased there).
- [ ] Emit the standardized headers on outgoing requests/responses: client in `streamable_http_client.rs`, server in `streamable_http_server/tower.rs` / `server_side_http.rs`.
- [ ] Parse and honor them where applicable (e.g. propagate tracing/priority context).
- [ ] Gate behind `ProtocolVersion::V_2026_07_28` so older peers are unaffected (the protocol version is already available per-request in `tower.rs`).
- [ ] Add tests; cross-check behavior against the C# reference implementation.

## Affected areas
`transport/common/http_header.rs` (new constants + allow-list), `transport/common/server_side_http.rs`,
`transport/streamable_http_client.rs`, `transport/streamable_http_server/tower.rs`.

## Notes / risks
- Coordinate with SEP-414 (OpenTelemetry trace context) — tracing headers may overlap.

DaleSeo:
Closes #872

## Motivation and Context

MCP Streamable HTTP keeps routing-related data within the JSON-RPC body. This means that middle boxes like proxies, gateways, and load balancers can't route or trace it without parsing. [SEP-2243](https://modelcontextprotocol.io/community/seps/2243-http-standardization) addresses this by introducing standard headers. This PR makes the client send `Mcp-Method`, `Mcp-Name`, and `Mcp-Param-*`. The last one is promoted from tool-call arguments that are marked with `x-mcp-header` and Base64-wrapped if they aren't header-safe. The server will validate these against the body and reject any mismatches with a `400 Bad Request` and JSON-RPC ~`-32001`~ `-32020`.
Everything is controlled by a new `ProtocolVersion::V_2026_07_28`, which matches the draft spec, so older peers won't be affected.

## How Has This Been Tested?

Added 32 unit tests and 8 integration tests

## Breaking Changes

`StreamableHttpService` now requires its service to implement `ServerHandler` instead of `Service<RoleServer>` so the server can read tool schemas for `Mcp-Param-*` validation.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -168,6 +168,9 @@ impl ProtocolVersion {
     pub const V_2024_11_05: Self = Self(Cow::Borrowed("2024-11-05"));
     pub const LATEST: Self = Self::V_2025_11_25;
 
+    /// First protocol version that requires SEP-2243 standard HTTP headers.
+    pub const STANDARD_HEADERS: Self = Self::V_2026_07_28;
+
     /// All protocol versions known to this SDK.
     pub const KNOWN_VERSIONS: &[Self] = &[
         Self::V_2024_11_05,
@@ -512,6 +515,7 @@ pub struct JsonRpcNotification<N = Notification> {
 pub struct ErrorCode(pub i32);
 
 impl ErrorCode {
+    pub const HEADER_MISMATCH: Self = Self(-32001);
     pub const RESOURCE_NOT_FOUND: Self = Self(-32002);
     pub const INVALID_REQUEST: Self = Self(-32600);
     pub const METHOD_NOT_FOUND: Self = Self(-32601);
@@ -557,7 +561,9 @@ impl ErrorData {
     pub fn resource_not_found(message: impl Into<Cow<'static, str>>, data: Option<Value>) -> Self {
         Self::new(ErrorCode::RESOURCE_NOT_FOUND, message, data)
     }
-
+    pub fn header_mismatch(message: impl Into<Cow<'static, str>>, data: Option<Value>) -> Self {
+        Self::new(ErrorCode::HEADER_MISMATCH, message, data)
+    }
     pub fn parse_error(message: impl Into<Cow<'static, str>>, data: Option<Value>) -> Self {
         Self::new(ErrorCode::PARSE_ERROR, message, data)
     }
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -515,7 +515,7 @@ pub struct JsonRpcNotification<N = Notification> {
 pub struct ErrorCode(pub i32);
 
 impl ErrorCode {
-    pub const HEADER_MISMATCH: Self = Self(-32001);
+    pub const HEADER_MISMATCH: Self = Self(-32020);
     pub const RESOURCE_NOT_FOUND: Self = Self(-32002);
     pub const INVALID_REQUEST: Self = Self(-32600);
     pub const METHOD_NOT_FOUND: Self = Self(-32601);
--- a/crates/rmcp/src/transport/common/mcp_headers.rs
+++ b/crates/rmcp/src/transport/common/mcp_headers.rs
@@ -252,7 +252,7 @@ pub(crate) fn standard_request_headers(
 /// Validates incoming SEP-2243 headers against the request body.
 ///
 /// Returns `Err(reason)` when a required header is missing or its value does not
-/// match the body; the caller maps this to a JSON-RPC `-32001` error (HTTP 400).
+/// match the body; the caller maps this to a JSON-RPC `-32020` error (HTTP 400).
 #[cfg(feature = "server-side-http")]
 pub(crate) fn validate_request_headers(
     headers: &http::HeaderMap,
--- a/crates/rmcp/src/transport/common/mcp_headers.rs
+++ b/crates/rmcp/src/transport/common/mcp_headers.rs
@@ -1,0 +1,654 @@
+//! SEP-2243 HTTP header standardization.
+//!
+//! Builds and validates the `Mcp-Method`, `Mcp-Name`, and `Mcp-Param-*` headers
+//! so middle boxes can route Streamable HTTP traffic without parsing the body.
+//! All emission/validation is gated by the negotiated protocol version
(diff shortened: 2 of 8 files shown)


crates/rmcp/src/transport/common/mcp_headers.rs
```diff
@@ -0,0 +1,654 @@
...
+        value.to_owned()
+    }
+}
+
+/// Reverses [`encode_header_value`]. Returns `None` if the sentinel wraps invalid Base64/UTF-8.
+#[cfg(feature = "server-side-http")]
+fn decode_header_value(value: &str) -> Option<String> {
+    use base64::{Engine, prelude::BASE64_STANDARD};
+    match value
+        .strip_prefix(BASE64_HEADER_PREFIX)
```
jamadeo, line 191: There's an "edge case" mentioned in the SEP stating the server should accept (case-insensitive match) but I'm wondering if that's a mistake, as earlier it says these markers must be lower-case.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -649,14 +712,20 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
...
+                    );
                     let response = self
                         .client
                         .post_message(
                             config.uri.clone(),
                             message.clone(),
                             session_id.clone(),
                             config.auth_header.clone(),
-                            protocol_headers.clone(),
+                            request_headers,
```
jamadeo, line 863: should this also be added to the POST in `perform_reinitialization()`?

jamadeo (dismissed):

crates/rmcp/src/transport/streamable_http_client.rs
DaleSeo: Good catch, @jamadeo! The `initialize` POST should stay the same since we won't know the version until we get the response, just like the first init in `run()`. But you're right about the `notifications/initialized` POST. I'll wire it through so both match. Thanks!

crates/rmcp/src/transport/common/mcp_headers.rs
DaleSeo: @jamadeo This seems like a real contradiction in the SEP. I've reported the spec issue: modelcontextprotocol/modelcontextprotocol#2936. For now, I'll keep the implementation as it is since it matches the prose.

jamadeo (dismissed):

DaleSeo:
This implements SEP-2243, which is part of the 2026-07-28 spec. Per our release sequencing, main is currently being prepared for v2.0 (alignment to the current 2025-11-25 schema), and all 2026-07-28 work is deferred to the v3.0 cycle. This PR stays parked until v2.0 is cut and we start landing 2026-07-28 changes on main.

crates/rmcp/src/model.rs
```diff
@@ -503,6 +506,7 @@ pub struct JsonRpcNotification<N = Notification> {
 pub struct ErrorCode(pub i32);
 
 impl ErrorCode {
+    pub const HEADER_MISMATCH: Self = Self(-32001);
```
howardjohn: This should be -32020 per the spec I think

DaleSeo: @howardjohn Good catch, thanks! I noticed the SEP text still had `-32001`, but the current draft spec renumbered `HeaderMismatch` to `-32020`.  I'll report this.

SEP-2243|Draft Spec
-|-
2026-06-30 at 09 32 28|2026-06-30 at 09 30 35

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -12,17 +12,89 @@ use super::common::client_side_sse::{ExponentialBackoff, SseRetryPolicy, SseStre
...
+                    tracing::warn!(tool = %tool.name, "ignoring x-mcp-header annotations: {reason}");
+                    continue;
+                }
+                cache.insert(tool.name.to_string(), tool.input_schema.clone());
+            }
+        }
+    }
+}
+
+/// Derives the negotiated protocol version and base request headers from an
```
alexhancock: I might drop a lot of these comments where the code already explains itself well, as they have a tendency to drift/become incorrect

alexhancock (dismissed):

crates/rmcp/src/transport/streamable_http_client.rs
DaleSeo: @alexhancock Cleaned that up and now I need re-approval 🙃

alexhancock (approved):

jamadeo (approved):

PR #908, opened 2026-06-16
Allow custom HTTP clients for OAuth

jif-oai:
## Why

RMCP currently performs OAuth network operations through internally created reqwest clients. Callers therefore cannot route OAuth through another network environment, even when their MCP HTTP transport supports it.

## What changed

- Add one injectable OAuth HTTP client boundary.
- Route protected-resource discovery, OAuth/OIDC discovery, dynamic registration, code exchange, refresh, and client credentials through it.
- Keep reqwest as the default for existing callers.
- Preserve the injected client across OAuth state transitions.
- Expose timeout and redirect requirements with each request so token operations can disable redirects.
- Add focused fake-client coverage for discovery, registration, exchange, and refresh.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -857,7 +857,7 @@ impl AuthorizationManager {
         Self::new_inner(
             base_url,
             Arc::new(ReqwestOAuthHttpClient::new(http_client)?),
-            OAuthHttpRedirectPolicy::Follow,
+            OAuthHttpRedirectPolicy::Stop,
         )
         .await
     }
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use async_trait::async_trait;
+use futures::StreamExt;
 use oauth2::{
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -23,39 +26,152 @@ use tracing::{debug, warn};
...
+/// Future returned by [`OAuthHttpClient::execute`].
+pub type OAuthHttpClientFuture<'a> =
+    Pin<Box<dyn Future<Output = Result<HttpResponse, OAuthHttpClientError>> + Send + 'a>>;
+
+/// Executes every outbound HTTP request made by the OAuth state machine.
+///
+/// Implementations may route requests through a remote execution environment.
+/// They must honor the request's redirect policy and return the raw response
+/// status, headers, and body.
+pub trait OAuthHttpClient: Send + Sync {
```
alexhancock, line 90: Worth adding these to this list of re-exports I think, right?

https://github.com/modelcontextprotocol/rust-sdk/blob/3c5ce2b0d78a5dee13457076c72971213829e670/crates/rmcp/src/transport.rs#L101

alexhancock (dismissed):

alexhancock (approved):

PR #915, opened 2026-06-20
feat!: add MRTR model types (SEP-2322)

### LINKED ISSUE #871: Implement SEP-2322: Multi Round-Trip Requests
# SEP-2322: Multi Round-Trip Requests (MRTR) — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322
**Conformance:** https://github.com/modelcontextprotocol/conformance/pull/188
**Track:** Specification · **Stage:** accepted · **Priority:** P0 · **Theme:** Transport Evolution and Scalability
**Needs code changes:** Yes (Large) — **Breaking**

## Summary
A Transports Working Group proposal for Multi Round-Trip Requests: a transport-level mechanism so a
single logical request can involve multiple message round-trips (e.g. for server→client requests
that occur within the lifetime of a client request) without depending on long-lived sessions. The
SEP carries schema changes and is explicitly flagged as a **breaking change**.

The current transports assume a session/connection over which server→client requests can occur at
any time. The request/response correlation today lives in the generic `service.rs` peer machinery
(request id → pending response map) and the streamable HTTP server `tower.rs`, which currently maps
each POST to an SSE stream for that request. MRTR reframes round trips around the lifetime of a
client request, which changes both that correlation and how `tower.rs` frames the response stream
when the server needs to call back into the client mid-request.

## Proposed work
- [ ] Add the MRTR schema fields/types to `model.rs` per the SEP.
- [ ] Add the MRTR framing to the streamable HTTP server (`streamable_http_server/tower.rs`): allow a single client POST's SSE response stream to carry interleaved server→client requests and their responses within that request's lifetime.
- [ ] Extend the client (`streamable_http_client.rs`) to answer server→client requests that arrive on the response stream of its own outbound request.
- [ ] Adjust request correlation in `service.rs` so server-initiated requests are tagged with the originating client request id (shared mechanism with SEP-2260).
- [ ] Gate behind `ProtocolVersion::V_2026_07_28` (breaking vs. older always-on-connection behavior).
- [ ] Wire up the conformance scenario from conformance#188.

## Affected areas
`crates/rmcp/src/service.rs` (peer request correlation), `service/client.rs`, `service/server.rs`,
`model.rs` (new framing/schema fields).

## Notes / risks
- Tightly coupled with SEP-2260 (server requests must associate with a client request) and SEP-2663 (tasks). Sequence MRTR first, then layer 2260 and tasks on top.
- Breaking vs. older always-on-connection behavior; gate behind `ProtocolVersion::V_2026_07_28` so older peers are unaffected.

DaleSeo:
## Motivation and Context

[SEP-2322](https://modelcontextprotocol.io/seps/2322-MRTR) introduces a stateless pattern where a server responds to `tools/call`, `prompts/get`, or `resources/read` with an `InputRequiredResult` (carrying `inputRequests` and/or `requestState`), and the client retries with `inputResponses`. This replaces the previous `URLElicitationRequiredError` (-32042) approach. Relates to #871.

The draft spec also adds a `resultType` discriminator to the base `Result` type (`required` in the JSON schema). This PR adds `result_type: ResultType` to all server result types that require it per the schema: `CallToolResult`, `GetPromptResult`, `ReadResourceResult`, `CompleteResult`, and the four paginated results (`ListToolsResult`, `ListPromptsResult`, `ListResourcesResult`, `ListResourceTemplatesResult`). `resultType` is always serialized (matching the Python SDK's approach per the spec's MUST). Types that don't have `resultType` in the schema (`InitializeResult`, task results, client results) are unchanged.

This PR adds only the model/serde layer. Behavior wiring (server emit, client retry, version gating) is planned for a follow-up.

## How Has This Been Tested?

13 unit tests in `model/mrtr.rs` covering serde round-trips against spec example JSON, rejection of missing/wrong `resultType`, unknown extension value preservation, and constructor behavior. Full suite passes with 0 failures.

```
cargo test -p rmcp --lib mrtr::tests
cargo test -p rmcp
```

## Breaking Changes

All server result types gain a `result_type: ResultType` field. The field defaults to `"complete"` and is always serialized, so responses now include `"resultType": "complete"`. Struct literals that construct result types directly (not through constructors or `..Default::default()`) will need the new field added. The `#[tool_handler]` and `#[prompt_handler]` proc macros have been updated. Adding `InputRequiredResult` to the exhaustive `ServerResult` union is an API-breaking change.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -12,6 +12,7 @@ mod content;
 mod elicitation_schema;
 mod extension;
 mod meta;
+mod mrtr;
 mod prompt;
 mod resource;
 mod serde_impl;
@@ -23,6 +24,7 @@ pub use content::*;
 pub use elicitation_schema::*;
 pub use extension::*;
 pub use meta::*;
+pub use mrtr::*;
 pub use prompt::*;
 pub use resource::*;
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
@@ -516,6 +518,10 @@ impl ErrorCode {
     pub const INVALID_PARAMS: Self = Self(-32602);
     pub const INTERNAL_ERROR: Self = Self(-32603);
     pub const PARSE_ERROR: Self = Self(-32700);
+    #[deprecated(
+        since = "2.0.0",
+        note = "URLElicitationRequiredError is removed by SEP-2322 (Multi Round-Trip Requests). Use InputRequiredResult instead."
+    )]
     pub const URL_ELICITATION_REQUIRED: Self = Self(-32042);
 }
 
@@ -572,6 +578,11 @@ impl ErrorData {
     pub fn internal_error(message: impl Into<Cow<'static, str>>, data: Option<Value>) -> Self {
         Self::new(ErrorCode::INTERNAL_ERROR, message, data)
     }
+    #[deprecated(
+        since = "2.0.0",
+        note = "URLElicitationRequiredError is removed by SEP-2322 (Multi Round-Trip Requests). Use InputRequiredResult instead."
+    )]
+    #[allow(deprecated)]
     pub fn url_elicitation_required(
         message: impl Into<Cow<'static, str>>,
         data: Option<Value>,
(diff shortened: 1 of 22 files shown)


crates/rmcp/src/model.rs
```diff
@@ -1185,6 +1261,9 @@ macro_rules! paginated_result {
         #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
         #[expect(clippy::exhaustive_structs, reason = "intentionally exhaustive")]
         pub struct $t {
+            /// Result type discriminator. Always serialized; older peers ignore it.
+            #[serde(default)]
+            pub result_type: ResultType,
```
howardjohn: I found this to be not actually happening with all peers. Notably, the MCP inspector explicitly rejected

howardjohn: I think it might have only been on the EmptyResult type though so perhaps here specifically is ok. This PR doesn't actually change EmptyResult to have `ResultType`.

In my work on this I ended up using `Option<ResultType>` to make it so we could avoid serializing this for older peers

DaleSeo: Good point, @howardjohn. I removed the wording that implied all older peers tolerate the extra field. This PR is focused on model changes, so I'll tackle your concern in a follow-up PR.

alexhancock (approved):

PR #919, opened 2026-06-23
feat!: relax tool result structuredContent type (SEP-2106)

### LINKED ISSUE #874: Implement SEP-2106: Tool schemas conform to JSON Schema 2020-12

DaleSeo:
## Motivation and Context

This was discovered while reviewing PR #895. Besides `CallToolResult`, `ToolResultContent` should also be typed with its structured content as `Option<Value>` to accommodate any JSON value.

2025-11-25|2026-07-28
-|-
2026-06-23 at 15 42 59|2026-06-23 at 15 41 47

## How Has This Been Tested?

Added tests

## Breaking Changes

Yes. The public field `ToolResultContent.structured_content` changes type from `Option<JsonObject>` to `Option<Value>`. 

The green SemVer Check is a false negative due to known upstream gap: obi1kenobi/cargo-semver-checks#148

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/tests/test_sampling.rs
+++ b/crates/rmcp/tests/test_sampling.rs
@@ -371,6 +371,38 @@ fn test_tool_result_content_requires_content() {
 }
 
 #[tokio::test]
+async fn test_tool_result_content_with_array_structured_content() -> Result<()> {
+    let structured =
+        serde_json::json!([{ "city": "SF", "temp": 72 }, { "city": "NY", "temp": 65 }]);
+    let mut tool_result = ToolResultContent::new("call_123", vec![ContentBlock::text("forecast")]);
+    tool_result.structured_content = Some(structured);
+
+    let json = serde_json::to_string(&tool_result)?;
+    let deserialized: ToolResultContent = serde_json::from_str(&json)?;
+    assert_eq!(tool_result, deserialized);
+    assert!(deserialized.structured_content.unwrap().is_array());
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_tool_result_content_with_primitive_structured_content() -> Result<()> {
+    let structured = serde_json::json!(42);
+    let mut tool_result = ToolResultContent::new("call_123", vec![ContentBlock::text("count")]);
+    tool_result.structured_content = Some(structured);
+
+    let json = serde_json::to_string(&tool_result)?;
+    let deserialized: ToolResultContent = serde_json::from_str(&json)?;
+    assert_eq!(tool_result, deserialized);
+    assert!(matches!(
+        deserialized.structured_content,
+        Some(serde_json::Value::Number(_))
+    ));
+
+    Ok(())
+}
+
+#[tokio::test]
 async fn test_sampling_message_with_tool_use() -> Result<()> {
     let message = SamplingMessage::assistant_tool_use(
         "call_123",
--- a/crates/rmcp/src/model/content.rs
+++ b/crates/rmcp/src/model/content.rs
@@ -8,7 +8,7 @@
 //! variants for sampling messages (SEP-1577).
 
 use serde::{Deserialize, Serialize};
-use serde_json::json;
+use serde_json::{Value, json};
 
 use super::{Annotations, Meta, resource::ResourceContents};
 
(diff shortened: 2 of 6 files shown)


crates/rmcp/tests/test_sampling.rs
```diff
@@ -370,6 +370,38 @@ fn test_tool_result_content_requires_content() {
...
+    let deserialized: ToolResultContent = serde_json::from_str(&json)?;
+    assert_eq!(tool_result, deserialized);
+    assert!(deserialized.structured_content.unwrap().is_array());
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_tool_result_content_with_primitive_structured_content() -> Result<()> {
+    let structured = serde_json::json!(42);
```
alexhancock, line 390: Worth one that loops through all the possible primitive types (string, number, boolean, null) and does the same test? This way it would catch if anyone changes it in some way that drops support for something.

alexhancock (approved):

crates/rmcp/tests/test_sampling.rs
DaleSeo: Good call, @alexhancock. I've opened #931 to bring in `rstest` for these repeated rmcp unit-test clusters (it also clears the existing `-D warnings` clippy failures on the test targets). Once #931 lands I'll come back and parametrize this with `rstest` so each primitive shows up as its own reported case.

PR #924, opened 2026-06-23
ci: add cargo-public-api check for breaking API changes

DaleSeo:
Adds a CI check using `cargo-public-api` to catch any source-breaking public API changes before the release. This is the safeguard suggested in https://github.com/modelcontextprotocol/rust-sdk/pull/850#issuecomment-4780318674.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,77 @@ jobs:
             --only-explicit-features \
             --features "$FEATURES"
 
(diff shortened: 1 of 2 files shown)


joshuajbouw (approved):
Looks good to me! Nice addition. I'm surprised that cargo-semver-checks can't catch an issue like this yet.

jamadeo (approved):

Issue #813, opened 2026-04-17
Follow semver, or document that rmcp doesn't follow it


jh-block:
There have been multiple breaking changes in 1.x releases (fields added to pub structs, structs marked non-exhaustive, etc), which means projects depending on rmcp can introduce build failures just by running `cargo update`. Based on the [typical semver approach](https://doc.rust-lang.org/cargo/reference/semver.html) taken by Rust crates, each of these should have been a new major version.

If rmcp has a different policy than following semver, it would be good to document that clearly so that people can be aware of the risk and consider pinning the version more precisely than the default caret requirement.

alexhancock:
Yeah let's discuss!

Our plan after https://github.com/modelcontextprotocol/rust-sdk/pull/715 and associated followups and the move to 1.0 was to use `release-plz` (which uses https://crates.io/crates/cargo-semver-checks) to follow semver in the standard way, but if we've shipped things that should have been majors and weren't it seems like a miss to me.

Adding fields to pub structs as long as they are optional should in theory be OK if the struct has appropriate constructors etc. This was the plan with `non_exhaustive` and `new` `new_with_X` constructors. But let's find where we've been missing.

@DaleSeo @jamadeo Do you have opinions/views?

jh-block:
Really good to know that following semver is a goal. To ground it a bit, the main breaking change I noticed (these caused a build failure after cargo update in goose, and also a packaging failure for Fedora) was that v1.3.0 marked a lot of pub structs as non_exhaustive (this is a breaking change, since any existing struct literals now fail to compile until they are modified to deal with possible future extra fields), e.g. `PromptContext`: https://github.com/modelcontextprotocol/rust-sdk/compare/rmcp-v1.2.0...rmcp-v1.3.0#diff-4a77e2940c44e190cc85d449d090c1ef6f46e61f161af436d6447956448dbe97R23 (there are a lot more in that same diff, any addition of `#[non_exhaustive]` to a pub struct where all fields are pub is a breaking change).

There was also breaking change with the macros; there was some code in a goose test that (possibly erroneously) used the tool_handler attribute macro without the router argument, and it built correctly on 1.2.0 but failed to build on 1.5.0.

The PR where I updated rmcp in goose is [here](https://github.com/aaif-goose/goose/pull/8618), in case the build logs and commits adapting to the breaking changes are helpful

DaleSeo:
[...]

PR #927, opened 2026-06-24
feat!: align model types with MCP 2025-11-25 spec

DaleSeo:
## Motivation and Context

Over time the `rmcp` model drifted from the MCP Schema Reference. The same content union existed in three slightly different forms (`Content`/`RawContent`, `PromptMessageContent`, `SamplingMessageContent`), many types sat behind an `Annotated<T>` wrapper, several names didn't match the spec (`ResourceReference`, `CreateElicitation*`, `GetTaskInfo*`, `PrimitiveSchema`, and others), and some spec fields were missing (`_meta` in several places, `audio` in prompt content). Finding the type that corresponds to a given spec construct, and knowing whether it was complete and correctly named, took real effort for SDK users and maintainers alike.

We had deferred this because doing it properly is breaking, and it wasn't worth a major bump on its own. The upcoming 2026-07-28 revision changes that: it requires breaking changes regardless, so realigning the existing model to the current 2025-11-25 schema in the same major (consistent with the versioning approach in #813) avoids an extra bump and a second round of migration. It also gives the 2026-07-28 work a much cleaner baseline to build on. A full migration guide lives in discussion #926, with each type mapped to the [2025-11-25 Schema Reference](https://modelcontextprotocol.io/specification/2025-11-25/schema).

## How Has This Been Tested?

- The `test-no-local` configuration: compute the feature set excluding `local` and run `cargo test -p rmcp --features "$FEATURES"` (mirrors the CI job that the default `--all-features` run does not cover)
- Regenerated `crates/rmcp/tests/test_message_schema/*.json` snapshots and cross-checked them against `schema/2025-11-25/schema.ts`
- Audited every model type field-by-field against `schema/2025-11-25/schema.ts` (names, required/optional, `_meta`, enum literals, method strings); the full `cargo test -p rmcp --all-features` suite passes

## Breaking Changes

Yes. This is a major (2.0) change. The JSON wire format is unchanged (only additive `_meta` / optional fields); the breaks are at the Rust API level: the unified `ContentBlock` (flattened, no more `Annotated<T>` / `Raw*`), prompts using `Role` + `ContentBlock`, spec-name renames (`ResourceReference` to `ResourceTemplateReference`, `CreateElicitation*` to `Elicit*`, `PrimitiveSchema` to `PrimitiveSchemaDefinition`, `GetTaskInfo*` / `GetTaskResult*` to `GetTask*` / `GetTaskPayload*`), most wire structs becoming `#[non_exhaustive]`, `CancelledNotificationParam.request_id` becoming optional, and the removal of the unused, non-spec `TaskList` (use `ListTasksResult`). `_meta` coverage is now complete across every `Result` type — `CreateTaskResult` and `ListRootsResult` were the last gaps. Old names are kept as `#[deprecated]` aliases where practical. `TaskStatusNotificationParam` now models `NotificationParams & Task` instead of being a `Task` alias; construct it with `TaskStatusNotificationParam::new(task)` or `task.into()`, and use `.task` where exact `Task` identity matters. `ToolResultContent.content` is now required for sampling tool results, so send `content: []` when there are no content blocks; `CallToolResult` still keeps its existing missing-content compatibility behavior.

## Types of changes

- [x] Documentation update

## Checklist


alexhancock (dismissed):

DaleSeo:
@alexhancock I pushed one follow-up commit after your approval. I removed the unused, no-spec `TaskList`.

alexhancock (approved):

PR #929, opened 2026-06-26
feat!: add MRTR behavior support (SEP-2322)

### LINKED ISSUE #871: Implement SEP-2322: Multi Round-Trip Requests

DaleSeo:
## Motivation and Context

On top of PR #915 that adds model types for [SEP-2322](https://modelcontextprotocol.io/seps/2322-MRTR), this PR wires up the behavior so servers can request input and clients can satisfy it without changing their call sites. Server tool/prompt/resource handlers can now return an `InputRequiredResult`, and the high-level client automatically fulfills the `input_required` rounds through the already-registered sampling, elicitation, and roots handlers before retrying the original call. The `*_once` methods remain as a manual escape hatch for callers that want to drive the rounds themselves. Native `InputRequiredResult` emission is gated on a negotiated protocol version of `2026-07-28` or newer so older peers never receive a result shape they cannot understand.

It also adds an opt-in `request-state` feature (`RequestStateCodec`, HMAC-SHA256) to safely round-trip the untrusted `requestState`, plus a `servers_mrtr` example and MRTR usage docs.
## How Has This Been Tested?

Add tests

## Breaking Changes

Server handler return types for `tools/call`, `prompts/get`, and `resources/read` are widened to MRTR-aware response enums (`CallToolResponse` / `GetPromptResponse` / `ReadResourceResponse`) so a handler can return either a final result or an `InputRequiredResult`. The `#[tool_handler]` / `#[prompt_handler]` proc macros, routers, conformance server, and existing tests are updated accordingly.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/handler/server.rs
+++ b/crates/rmcp/src/handler/server.rs
@@ -27,6 +27,9 @@ impl<H: ServerHandler> Service<RoleServer> for H {
     ) -> Result<<RoleServer as ServiceRole>::Resp, McpError> {
         // `context` is moved into the dispatch below, so read the negotiated version first.
         let protocol_version = context.protocol_version();
+        let mrtr_supported = protocol_version
+            .as_ref()
+            .is_some_and(|v| v.as_str() >= ProtocolVersion::V_2026_07_28.as_str());
         let result = match request {
             ClientRequest::InitializeRequest(request) => self
                 .initialize(request.params, context)
@@ -46,7 +49,7 @@ impl<H: ServerHandler> Service<RoleServer> for H {
             ClientRequest::GetPromptRequest(request) => self
                 .get_prompt(request.params, context)
                 .await
-                .map(ServerResult::GetPromptResult),
+                .map(ServerResult::from),
             ClientRequest::ListPromptsRequest(request) => self
                 .list_prompts(request.params, context)
                 .await
@@ -62,7 +65,7 @@ impl<H: ServerHandler> Service<RoleServer> for H {
             ClientRequest::ReadResourceRequest(request) => self
                 .read_resource(request.params, context)
                 .await
-                .map(ServerResult::ReadResourceResult),
+                .map(ServerResult::from),
             ClientRequest::SubscribeRequest(request) => self
                 .subscribe(request.params, context)
                 .await
@@ -105,7 +108,7 @@ impl<H: ServerHandler> Service<RoleServer> for H {
                 } else {
                     self.call_tool(request.params, context)
                         .await
-                        .map(ServerResult::CallToolResult)
+                        .map(ServerResult::from)
                 }
             }
             ClientRequest::ListToolsRequest(request) => self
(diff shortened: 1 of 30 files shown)


crates/rmcp/src/handler/server.rs
```diff
@@ -542,7 +556,7 @@ macro_rules! impl_server_handler_for_wrapper {
                 &self,
                 request: CallToolRequestParams,
                 context: RequestContext<RoleServer>,
-            ) -> impl Future<Output = Result<CallToolResult, McpError>> + MaybeSendFuture + '_ {
+            ) -> impl Future<Output = Result<CallToolResponse, McpError>> + MaybeSendFuture + '_ {
```
alexhancock, line 559: Thoughts on all these renames? Just wondering why the move from Result to Response

alexhancock (approved):

crates/rmcp/src/handler/server.rs
DaleSeo: @alexhancock Not a rename. `CallToolResult`, `GetPromptResult`, `ReadResourceResult` are unchanged. `CallToolResponse` is a new outcome enum mirroring what the spec's draft schema reference already models. Each method's response result is a union `InputRequiredResult | CallToolResult`. Rust has no anonymous unions, so I introduced one named enum per method to express that union at the handler return type.

Existing handlers don't need to change because the `IntoCallToolResult` conversions still accept a plain `CallToolResult`, `Json<T>`, `Content`, `Result<T, E>`, etc., and there are `From<CallToolResult>` / `From<InputRequiredResult>` impls, so `.into()` covers both arms. Only the trait-level return type widened.

alexhancock (dismissed):

jamadeo (approved):

PR #939, opened 2026-06-28
fix: ignore unparseable async rw messages

### LINKED ISSUE #938: rmcp shouldn't respond to unparseable messages
**Describe the bug**
When rmcp receives invalid json, it sends back `{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}}`. As far as I can tell, it's the only one of the official MCP SDKs that responds in this manner to invalid data. Most of the others just ignore invalid data, and none of the others respond. The problem with responding is it can trigger a storm of communication, where the client sending this back triggers the server to respond with more invalid data, which triggers another "parse error" response, which triggers more invalid data, and so on. I ported a component from TypeScript to Rust and thus switched from the TypeScript SDK to the Rust SDK, and it immediately caused problems when talking to a server that (erroneously) was putting some logging onto stdout.

**To Reproduce**
Try to connect to a stdio server that puts invalid data onto stdout, e.g. a logging message every time data is received.

**Expected behavior**
The invalid data will be ignored (this happens) and no response will be sent back (this doesn't happen).

dpersek:
## Summary
- Stop `AsyncRwTransport::receive()` from replying with a JSON-RPC parse error when an incoming line is not parseable JSON.
- Keep logging the parse failure and continue reading so a following valid message can still be received.
- Update the async-rw regression test to assert invalid JSON produces no peer response and the next valid notification is delivered.

Fixes #938

## Verification
- `git diff --check` -> passed
- `cargo fmt --check` -> passed, with existing stable-rustfmt warnings about unstable rustfmt options
- `cargo test -p rmcp --features server receive_recovers_from_parse_error -- --nocapture` -> passed; 1 passed, 0 failed
- `cargo test -p rmcp --features server transport::async_rw -- --nocapture` -> passed; 6 passed, 0 failed
- `cargo clippy -p rmcp --features server --lib -- -D warnings` -> passed
- `cargo test -p rmcp --features server --lib` -> passed; 124 passed, 0 failed

## Caveats
- `cargo clippy -p rmcp --features server --all-targets -- -D warnings` fails locally on pre-existing unrelated dead-code warnings in `tests/test_tool_routers.rs` and `tests/test_prompt_routers.rs`.
- The CI-style `cargo +nightly fmt --all -- --check` could not be run locally because this environment's `cargo` is not the `rustup` shim and `rustup` is unavailable; `cargo fmt --check` passed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,12 +4,12 @@ default-members = ["crates/rmcp", "crates/rmcp-macros"]
 resolver = "2"
 
 [workspace.dependencies]
-rmcp = { version = "1.8.0", path = "./crates/rmcp" }
-rmcp-macros = { version = "1.8.0", path = "./crates/rmcp-macros" }
+rmcp = { version = "2.1.0", path = "./crates/rmcp" }
+rmcp-macros = { version = "2.1.0", path = "./crates/rmcp-macros" }
 
 [workspace.package]
 edition = "2024"
-version = "1.8.0"
+version = "2.1.0"
 authors = ["4t145 <u4t145@163.com>"]
 license = "Apache-2.0"
 repository = "https://github.com/modelcontextprotocol/rust-sdk/"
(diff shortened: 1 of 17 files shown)


alexhancock:
Closing as this seems an empty diff and already addressed in https://github.com/modelcontextprotocol/rust-sdk/pull/940

PR #940, opened 2026-06-28
fix: don't respond to unparseable messages

### LINKED ISSUE #938: rmcp shouldn't respond to unparseable messages

tsouth89:
Fixes #938.

When rmcp receives invalid JSON on the stdio / async-read-write transport, it replies with `{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}}`. As reported in the issue, rmcp is the only official MCP SDK that responds to unparseable input, and responding can cause an error storm: if the peer treats the response itself as invalid data and answers with another error, the two sides bounce parse errors back and forth. The reporter hit this against a stdio server that accidentally wrote a log line to stdout.

This makes the transport ignore unparseable messages instead of responding. It still logs at `debug` and continues reading the next message, which matches the behavior of the other official SDKs. The now-unused `ErrorData` import is removed, and the existing test is updated to assert that no response is sent (and that the following valid message is still delivered).

Verified locally:
- `cargo test -p rmcp --features server,transport-async-rw` (the updated `receive_ignores_parse_error` test passes)
- `cargo clippy -p rmcp --all-features --all-targets -- -D warnings`
- `cargo +nightly fmt --check`

### DIFF
--- a/crates/rmcp/src/transport/async_rw.rs
+++ b/crates/rmcp/src/transport/async_rw.rs
@@ -13,10 +13,7 @@ use tokio_util::{
 };
 
 use super::{IntoTransport, Transport};
-use crate::{
-    model::ErrorData,
-    service::{RxJsonRpcMessage, ServiceRole, TxJsonRpcMessage},
-};
+use crate::service::{RxJsonRpcMessage, ServiceRole, TxJsonRpcMessage};
 
 #[non_exhaustive]
 pub enum TransportAdapterAsyncRW {}
@@ -143,16 +140,12 @@ where
                 Ok(Some(msg)) => return Some(msg),
                 Ok(None) => continue,
                 Err(JsonRpcMessageCodecError::Serde(e)) => {
-                    tracing::debug!("Parse error on incoming message: {e}");
-                    let mut write = self.write.lock().await;
-                    let framed = write.as_mut()?;
-                    let response = TxJsonRpcMessage::<Role>::error(
-                        ErrorData::parse_error("Parse error", None),
-                        None,
-                    );
-                    if framed.send(response).await.is_err() {
-                        return None;
-                    }
+                    // Don't respond to unparseable input. The message id is unknown, so a response
+                    // can't be correlated to a request, and replying to invalid data can trigger an
+                    // error storm if the peer echoes the response back as more invalid data. This
+                    // matches the other official MCP SDKs, which ignore unparseable messages.
+                    // See https://github.com/modelcontextprotocol/rust-sdk/issues/938
+                    tracing::debug!("Ignoring unparseable incoming message: {e}");
                 }
                 Err(e) => {
                     tracing::error!("Error reading from stream: {}", e);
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/async_rw.rs
```diff
@@ -143,16 +140,12 @@ where
                 Ok(Some(msg)) => return Some(msg),
                 Ok(None) => continue,
                 Err(JsonRpcMessageCodecError::Serde(e)) => {
```
DaleSeo, line 145: I think we should narrow the ignore path because `serde_json::Error` is not limited to invalid JSON syntax. It can also represent valid JSON that fails to deserialize into the role-specific `RxJsonRpcMessage<Role>` shape, which should remain a protocol error rather than disappearing.

DaleSeo (dismissed):
Thanks for the fix, @tsouth89. I left one comment.

btsouth:
Good catch, thanks. You're right that `serde_json::Error` conflates two cases. I've narrowed it: the receive loop now classifies the error, so only `Category::Syntax` and `Category::Eof` (genuinely unparsable input, where there's no id to correlate a reply to) are ignored per #938. `Category::Data` (valid JSON that doesn't match the `RxJsonRpcMessage<Role>` shape) is treated as a real protocol error and still gets an error response instead of disappearing. Added a test covering that path.

One open question: for the data-error case I kept the existing `-32700 Parse error` response for a minimal change, but `-32600 Invalid Request` is arguably the more accurate code for well-formed JSON with the wrong shape. Happy to switch to that if you'd prefer.

DaleSeo (dismissed):
> One open question: for the data-error case I kept the existing `-32700 Parse error` response for a minimal change, but `-32600 Invalid Request` is arguably the more accurate code for well-formed JSON with the wrong shape. Happy to switch to that if you'd prefer.

@tsouth89 That sounds like the right fix. Could you make that change before we merge this? Thanks!

btsouth:
Done. Switched the data-error case to `-32600 Invalid Request` and updated the test accordingly. Thanks for the review!

DaleSeo (approved):

Issue #869, opened 2026-06-02
Implement SEP-2575: Make MCP Stateless


alexhancock:
# SEP-2575: Make MCP Stateless — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575
**Track:** Specification · **Stage:** accepted · **Priority:** P0
**Needs code changes:** Yes (Large)

## Summary
Makes MCP **stateless-by-default** by removing the mandatory initialization handshake and
replacing it with discrete, "pay as you go" alternatives:
- **Per-request protocol version** via the `MCP-Protocol-Version` HTTP header and a `_meta` field, with negotiation through `UnsupportedVersionError` responses.
- **`server/discover` RPC** — optional endpoint for capabilities, supported versions, and metadata, replacing the capability exchange from `initialize`.
- **Per-request client capabilities** — clients declare capabilities in `_meta` per request instead of once at connection time.
- **`messages/listen` RPC** — dedicated endpoint for client-initiated streaming, replacing the Streamable HTTP GET behavior.

## Why this needs code changes in rust-sdk
Some of the groundwork already exists, which changes the shape of the work:
- `ProtocolVersion` (`crates/rmcp/src/model.rs`) is a small struct with `V_2025_11_25` (= `LATEST`), `KNOWN_VERSIONS`, and a forgiving `Deserialize` that already accepts unknown version strings — so adding a new version is a one-line + array-entry change.
- The streamable HTTP server **already parses `HEADER_MCP_PROTOCOL_VERSION`** per request in `streamable_http_server/tower.rs` (used today only for validation, not for per-request behavior switching).
- `StreamableHttpServerConfig` already has `stateful_mode: bool` (default `true`) and `with_stateful_mode(false)` to disable sessions — i.e. a basic stateless path exists, but it still assumes the `initialize` handshake and doesn't yet implement `server/discover` / `messages/listen` / per-request capabilities.

What's genuinely missing: the discrete handshake-replacement RPCs and per-request capability/version negotiation.

## Proposed work
- [ ] Add `ProtocolVersion::V_2026_07_28`, append it to `KNOWN_VERSIONS`, and set it as `LATEST`.
- [ ] Promote the existing `HEADER_MCP_PROTOCOL_VERSION` parse in `tower.rs` into a per-request *negotiated version* that drives behavior (not just validation); also read the per-request `_meta` version field.
[...]

PR #942, opened 2026-06-29
feat: add SEP-2575 meta helpers

howardjohn:
## Motivation and Context
Expose typed accessors for the per-request protocol version, client info, client capabilities, and log level entries carried in _meta without making those fields required for older peers.

This is needed for part of SEP-2575 which requires reading/writing these fields. Obviously its possible to set these manually today but given these are first class fields it seems nice to have typed APIs for them.

For #869

## How Has This Been Tested?
Testing in Agentgateway (unmerged)

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model/meta.rs
+++ b/crates/rmcp/src/model/meta.rs
@@ -4,8 +4,9 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
 use super::{
-    ClientNotification, ClientRequest, CustomNotification, CustomRequest, Extensions, JsonObject,
-    JsonRpcMessage, NumberOrString, ProgressToken, ServerNotification, ServerRequest, TaskMetadata,
+    ClientCapabilities, ClientNotification, ClientRequest, CustomNotification, CustomRequest,
+    Extensions, Implementation, JsonObject, JsonRpcMessage, LoggingLevel, NumberOrString,
+    ProgressToken, ProtocolVersion, ServerNotification, ServerRequest, TaskMetadata,
 };
 
 pub trait GetMeta {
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/model/meta.rs
```diff
@@ -200,6 +201,16 @@ variant_extension! {
...
 const PROGRESS_TOKEN_FIELD: &str = "progressToken";
+
+/// `_meta` key carrying the MCP protocol version for this request.
+pub const META_KEY_PROTOCOL_VERSION: &str = "io.modelcontextprotocol/protocolVersion";
+/// `_meta` key carrying the client implementation identity for this request.
+pub const META_KEY_CLIENT_INFO: &str = "io.modelcontextprotocol/clientInfo";
+/// `_meta` key carrying the client capabilities for this request.
+pub const META_KEY_CLIENT_CAPABILITIES: &str = "io.modelcontextprotocol/clientCapabilities";
+/// `_meta` key carrying the requested per-request log level.
+pub const META_KEY_LOG_LEVEL: &str = "io.modelcontextprotocol/logLevel";
```
DaleSeo: Do these constants need to be part of the public API? It seems like the getters and setters below should be sufficient for consumers. How about we keep them internal until there's a clear need for them downstream?

DaleSeo: I would also consider making these private associated constants on `Meta` instead of module-level constants. The raw keys are only used by the `Meta` helpers, so keeping them under `impl Meta` makes the ownership clearer.

crates/rmcp/src/model/meta.rs
```diff
@@ -249,11 +260,72 @@ impl Meta {
...
         }
     }
+
+    fn decode_value<T>(&self, key: &str) -> Option<T>
+    where
+        T: for<'de> Deserialize<'de>,
+    {
+        self.0
+            .get(key)
+            .and_then(|value| serde_json::from_value(value.clone()).ok())
```
DaleSeo: Serde can deserialize directly from `&Value` instead of cloning `Value` every time.

```suggestion
            .and_then(|value| T::deserialize(value).ok())
```

DaleSeo (commented):
Thanks for adding this, @howardjohn! I have a few comments.

DaleSeo:
Wrapping the long commit body will help the CI job pass.

2026-06-30 at 10 44 46

DaleSeo (approved):

PR #956, opened 2026-07-05
feat!: type Annotations.lastModified as a string

DaleSeo:
## Motivation and Context

[The MCP 2025-11-25 schema](https://modelcontextprotocol.io/specification/2025-11-25/schema#annotations) types `Annotations.lastModified` as a plain string. Some schema-valid values, such as date-only strings or offset-less date-times, are not strict RFC 3339 and could previously cause a whole message to fail deserialization when nested in content or resource annotations.

This PR changes `Annotations.last_modified` to preserve `lastModified` as the schema-defined string instead of eagerly parsing it as `DateTime<Utc>`. The DateTime-based writer helpers still accept `DateTime<Utc>` and serialize RFC 3339 strings, while callers that need a typed read can use `last_modified_datetime()`.

This is also consistent with `Task.createdAt` and `Task.lastUpdatedAt`, which are already represented as strings in rmcp. Keeping the raw string is lossless for proxying and forwarding, while the new accessor keeps DateTime ergonomics for callers that want best-effort typed parsing.

## How Has This Been Tested?

Added tests

## Breaking Changes

Yes. The public field `Annotations.last_modified` changes type from `Option<DateTime<Utc>>` to `Option<String>`.

The public API / SemVer check is expected to flag this field-type change. That is intentional for the v3 release.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model/annotated.rs
+++ b/crates/rmcp/src/model/annotated.rs
@@ -20,20 +20,23 @@ pub struct Annotations {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub priority: Option<f32>,
     #[serde(skip_serializing_if = "Option::is_none", rename = "lastModified")]
-    pub last_modified: Option<DateTime<Utc>>,
+    pub last_modified: Option<String>,
 }
 
 impl Annotations {
-    /// Creates a new Annotations instance specifically for resources
-    /// optional priority, and a timestamp (defaults to now if None)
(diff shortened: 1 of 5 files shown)


jamadeo (approved):
Agreed that this is more correct. I think there is some value in the API offering deserialization so that users aren't concerned with how the instant is serialized, but correctness matters more of course.

Maybe if there is interest, we could have fallible methods that deserialize in the future.

PR #957, opened 2026-07-05
fix: don't respond to cancelled requests

DaleSeo:
## Motivation and Context

This fixes the service loop so a locally produced response is not sent after the peer has already cancelled the matching request. The existing cancellation path removes the request id from `local_ct_pool`, so the response path now treats a missing id as the signal to suppress the late response instead of writing it to the transport. This keeps cancellation best-effort and race-safe while matching the MCP guidance that a receiver should not respond to a cancelled request.

## How Has This Been Tested?

Add integration tests

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -1103,10 +1103,6 @@ where
                         JsonRpcMessage::Error(error) => error.id.as_ref(),
                         _ => None,
                     } {
-                        // A response whose request is no longer tracked means the
-                        // request was cancelled (the cancelled-notification handler
-                        // removed it from the pool). Per the spec, a receiver SHOULD
-                        // NOT send a response for a cancelled request, so drop it.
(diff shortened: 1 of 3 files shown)


alexhancock (approved):

crates/rmcp/src/service.rs
```diff
@@ -1103,9 +1103,15 @@ where
                         JsonRpcMessage::Error(error) => error.id.as_ref(),
                         _ => None,
                     } {
-                        if let Some(ct) = local_ct_pool.remove(id) {
-                            ct.cancel();
-                        }
+                        // A response whose request is no longer tracked means the
```
alexhancock: I might remove this comment. The code is somewhat self-explanatory

alexhancock (dismissed):

alexhancock (approved):

PR #962, opened 2026-07-07
RFC 9728 resource is used instead of base url when possible

fkonecny:
## Motivation and Context

Library initiates OAuth2 workflow with predefined resource as the base url to abide RFC 8707. This violates a possibility where MCP belongs under umbrella of exposed functionalities. RFC 8707 permits it in [its second paragraph](https://www.rfc-editor.org/rfc/rfc8707.html#name-resource-parameter).

There was already implemented resource discovery from OAuth2 resource metadata document defined by RFC 9728 in [PR 937](https://github.com/modelcontextprotocol/rust-sdk/pull/937) which can access the exposed resource.

Use resource defined by OAuth2 resource metadata document so the creator of MCP and its provider can control what resource are tokens requested for.

## How Has This Been Tested?

2 tests were added:
- OAuth2 resource metadata document is missing,
- OAuth2 resource metadata document declares resources.

PR 937 already handles situations where document has missing resource.

## Breaking Changes

Breaking change: changes default behavior where base url is utilized always for resource and use defined resource by provider when available.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -816,6 +816,8 @@ pub struct AuthorizationManager {
     www_auth_scopes: RwLock<Vec<String>>,
     /// scopes_supported from protected resource metadata (RFC 9728)
     resource_scopes: RwLock<Vec<String>>,
+    /// resource indicator from protected resource metadata, used for RFC 8707 `resource`
+    discovered_resource: RwLock<Option<String>>,
     /// OIDC Dynamic Client Registration `application_type` (SEP-837)
     application_type: Option<String>,
 }
@@ -1050,6 +1052,7 @@ impl AuthorizationManager {
             scope_upgrade_config: ScopeUpgradeConfig::default(),
             www_auth_scopes: RwLock::new(Vec::new()),
             resource_scopes: RwLock::new(Vec::new()),
+            discovered_resource: RwLock::new(None),
             application_type: Some(DEFAULT_APPLICATION_TYPE.to_string()),
         };
 
@@ -1428,7 +1431,7 @@ impl AuthorizationManager {
         let mut auth_request = oauth_client
             .authorize_url(CsrfToken::new_random)
             .set_pkce_challenge(pkce_challenge)
-            .add_extra_param("resource", self.base_url.to_string());
+            .add_extra_param("resource", self.oauth_resource().await);
 
         // add request scopes
         for scope in scopes {
@@ -1464,6 +1467,14 @@ impl AuthorizationManager {
             .await?;
 
         Ok(auth_url.to_string())
+    }
+
+    async fn oauth_resource(&self) -> String {
+        self.discovered_resource
+            .read()
+            .await
+            .clone()
+            .unwrap_or_else(|| self.base_url.to_string())
     }
 
     /// get the current granted scopes
@@ -1709,7 +1720,7 @@ impl AuthorizationManager {
         let token_result = match oauth_client
             .exchange_code(AuthorizationCode::new(code.to_string()))
             .set_pkce_verifier(pkce_verifier)
-            .add_extra_param("resource", self.base_url.to_string())
+            .add_extra_param("resource", self.oauth_resource().await)
             .request_async(&OAuth2HttpClient {
                 client: self.http_client.as_ref(),
                 redirect_policy: OAuthHttpRedirectPolicy::Stop,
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -1351,6 +1354,14 @@ impl AuthorizationManager {
         Ok(auth_url.to_string())
     }
 
+    async fn oauth_resource(&self) -> String {
+        self.discovered_resource
+            .read()
```
DaleSeo, line 1474: Shouldn't we store this before reading it?

fkonecny: I am sorry (focused so much on PR text that I muddled stash pop).

I have patched it in [the next commit](https://github.com/modelcontextprotocol/rust-sdk/pull/962/commits/1e453b1e008e7fd8a6c880befe9efe3f7f784a2f).

I have kept your strict check about path equality.

crates/rmcp/tests/test_sampling.rs
```diff
@@ -375,7 +375,7 @@ fn test_tool_result_content_requires_content() {
 #[case::array(serde_json::json!([{ "city": "SF", "temp": 72 }, { "city": "NY", "temp": 65 }]))]
 #[case::string(serde_json::json!("sunny"))]
 #[case::integer(serde_json::json!(42))]
-#[case::float(serde_json::json!(3.14))]
+#[case::float(serde_json::json!(2.5))]
```
alexhancock: Why are we changing this in this diff? Can we remove this line change?

fkonecny: It's reverted

alexhancock (dismissed):

alexhancock:
@filip-konecny-tfs Can you take a look at the red checks? Once resolved we can move forward

fkonecny:
@alexhancock I did squash the commits and shortened the message so I can pass the lint message checks. Which erased the history of checks (sorry).

The public API check is confusing me as it refers things which I did not initiate, at least I think.

The semver check is confusing me, should I be the one bumping it? I can if you wish to do so.

P.S. @filip-konecny-tfs is my work account which I should not use for this type of activities :-).

crates/rmcp/src/transport/auth.rs
```diff
@@ -2162,43 +2178,51 @@ impl AuthorizationManager {
...
+        }
 
-        root_resource.path() == "/"
-            && root_resource.query().is_none()
-            && root_resource.fragment().is_none()
-            && path_resource.path() != "/"
-            && Self::is_same_origin(&root_resource, &path_resource)
+        expected_path.starts_with(actual_path)
+            && expected.query().is_none()
+            && actual.query().is_none()
```
DaleSeo, line 2225: Here, `expected` is `base_url`, and `actual` is the provider-declared resource. This means it accepts any declared path where the `base_url` path starts with that path. For example, if `base_url` is `https://mcp.example.com/mcp-tools` and the metadata declares `https://mcp.example.com/mcp`, then `"/mcp-tools".starts_with("/mcp")` is true.

crates/rmcp/src/transport/auth.rs
```diff
@@ -2162,43 +2178,51 @@ impl AuthorizationManager {
             ));
         };
 
-        if !Self::resource_identifiers_match(self.base_url.as_str(), resource) {
+        let Some(resource_url) = Url::parse(resource).ok() else {
+            return Err(AuthError::MetadataError(
+                "Protected resource metadata resource field is not a valid URL".to_string(),
+            ));
+        };
```
DaleSeo: nit

```suggestion
        let Ok(resource_url) = Url::parse(resource) else {
            return Err(AuthError::MetadataError(
                "Protected resource metadata resource field is not a valid URL".to_string(),
            ));
        };
```

crates/rmcp/src/transport/auth.rs
fkonecny: Fixed and added test for it. Thank you

alexhancock:
If you fix up the commit messages including a `!` to denote the breaking changes it will pass the red checks

DaleSeo:
@alexhancock I updated the commit messages and forced pushed them. We should be able to merge this.

DaleSeo (approved):

fkonecny:
@DaleSeo and @alexhancock, thank you for your patience and help

Issue #879, opened 2026-06-02
Implement SEP-2352: Clarify authorization server binding and migration


alexhancock:
# SEP-2352: Clarify authorization server binding and migration — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2352
**Track:** Specification · **Stage:** accepted · **Priority:** P0 · **Theme:** Enterprise Readiness
**Needs code changes:** Yes (Medium)

## Summary
Multi-AS guidance:
- Clients **SHOULD** maintain separate registration state per authorization server and **MUST NOT** assume cross-AS credential validity.
- New "Authorization Server Binding" rules: DCR / pre-registered credentials must be keyed by issuer; CIMD `client_id`s are portable; clients should **error on AS mismatch**.

## Why this needs code changes in rust-sdk
The OAuth client in `crates/rmcp/src/transport/auth.rs` persists credentials and registration state
via the `CredentialStore` trait (`StoredCredentials`, `InMemoryCredentialStore`) and the
`StateStore` trait (`StoredAuthorizationState`, `InMemoryStateStore`). Today these are effectively
**single-AS**: `StoredCredentials` and the in-memory stores are not keyed by issuer, and
`AuthorizationManager` holds one `metadata: Option<AuthorizationMetadata>` (whose `issuer` field
exists but isn't used to partition stored credentials). So credentials from one AS could be reused
against another, which SEP-2352 forbids.

## Proposed work
- [ ] Key `CredentialStore` / `StateStore` entries by issuer — either add an `issuer` discriminator to `StoredCredentials` / `load`/`save` keys, or make the in-memory stores `HashMap<issuer, …>` so per-AS state is isolated.
- [ ] On token use, compare the active `AuthorizationMetadata.issuer` against the issuer the stored credentials were minted for; on mismatch return a clear `AuthError` (add a variant, e.g. `AuthError::AuthorizationServerMismatch`).
- [ ] Treat CIMD `client_id`s as portable (don't tie those to a single issuer), but keep DCR/pre-registered credentials issuer-bound.
- [ ] Update `register_client` / `configure_client_id` / `get_credentials` paths to read/write the issuer-keyed store.
- [ ] Add tests: two ASes with distinct registrations don't cross-contaminate; mismatch is rejected.

## Affected areas
[...]

PR #965, opened 2026-07-08
feat(auth): bind stored credentials to authorization server issuer (SEP-2352)

alexhancock:
## Summary

Implements the authorization-server binding rule from **SEP-2352** (`Clarify authorization server binding and migration`). OAuth credentials are now bound to the `issuer` of the authorization server (AS) that minted them, and `get_access_token()` refuses to hand out a token when the manager is configured against a *different* known issuer. This prevents credentials obtained from one AS from being silently reused against another, which SEP-2352 forbids.

Refs #879.

## What changed

All changes are confined to `crates/rmcp/src/transport/auth.rs`.

- **`StoredCredentials.issuer: Option<String>`** — new field recording the AS issuer the credentials were minted for.
  - `#[serde(default)]` so credentials persisted before this change still deserialize (issuer becomes `None`).
  - `StoredCredentials::new()` signature is **unchanged** (still sets `issuer: None`); added a `with_issuer()` builder.
  - `Debug` impl updated to include the issuer (no secrets exposed).
- **`AuthorizationManager::stored_credentials()`** — new helper that stamps the currently-configured issuer (from discovered `AuthorizationMetadata`) onto persisted credentials. Wired into every token-save site: authorization-code exchange, refresh, client-credentials, client-credentials-JWT, and `OAuthState::set_credentials` (which now discovers metadata before persisting so the issuer is available).
- **`AuthorizationManager::check_issuer_binding()`** — guard invoked at the top of `get_access_token()`. On a known-issuer mismatch it returns the new `AuthError::AuthorizationServerMismatch { stored, active }`.
- **Re-register on mismatch (SEP-2352: MUST NOT reuse / MUST re-register)** — on a detected mismatch, `get_access_token()` clears the stale credentials before returning the error, so the next authorization flow re-registers with the new AS rather than reusing credentials invalid there.
- **CIMD portability carve-out** — CIMD client_ids (HTTPS URLs) are portable across authorization servers, so they are exempt from the binding check and never trigger a mismatch.

## Backward compatibility

Non-breaking:
- Legacy credentials with `issuer == None` are treated as unbound and pass through unchanged.
- If the active manager has no issuer in its metadata, no mismatch can be proven, so bound credentials still work.
- `StoredCredentials` and `AuthError` are both `#[non_exhaustive]`, and the public `new()` signature is preserved.

## Tests

Added 9 tests:
- mismatch is rejected with `AuthorizationServerMismatch`
- stale credentials are cleared on mismatch (re-registration path)
- portable CIMD client_id (HTTPS URL) passes through despite differing issuer
- matching issuer passes through
- legacy credentials (issuer `None`) pass through
- unknown active issuer passes through
- `stored_credentials()` stamps the active issuer
- no issuer stamped when metadata is absent
- `with_issuer()` builder sets the issuer

`cargo build`, `cargo test`, `cargo clippy`, and `cargo fmt` all pass (feature `auth`); 90 auth tests green.

## Scope / notes

Focused, self-contained slice of the SEP-2352 work: issuer binding, mismatch enforcement + re-registration on the token-use path, and the CIMD portability carve-out. The remaining broader work — per-AS keying of custom credential/state stores so multiple AS credentials can coexist (rather than a single active set) — is intentionally left for follow-up. Marked as **draft** for early review.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -200,6 +200,8 @@ pub struct StoredCredentials {
     pub granted_scopes: Vec<String>,
     #[serde(default)]
     pub token_received_at: Option<u64>,
+    #[serde(default)]
+    pub issuer: Option<String>,
 }
 
 impl std::fmt::Debug for StoredCredentials {
@@ -212,6 +214,7 @@ impl std::fmt::Debug for StoredCredentials {
             )
             .field("granted_scopes", &self.granted_scopes)
             .field("token_received_at", &self.token_received_at)
+            .field("issuer", &self.issuer)
             .finish()
     }
 }
@@ -229,7 +232,13 @@ impl StoredCredentials {
             token_response,
             granted_scopes,
             token_received_at,
+            issuer: None,
         }
+    }
+
+    pub fn with_issuer(mut self, issuer: impl Into<String>) -> Self {
+        self.issuer = Some(issuer.into());
+        self
     }
 }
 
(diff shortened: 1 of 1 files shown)


alexhancock:
Replaced with https://github.com/modelcontextprotocol/rust-sdk/pull/998

PR #970, opened 2026-07-10
fix: bound streamable HTTP memory usage

DaleSeo:
## Motivation and Context

The Streamable HTTP client previously allowed an SSE event to grow indefinitely while waiting for its terminating blank line. The server similarly buffered complete POST bodies without a size limit, allowing a peer to exhaust process memory.
This PR applies configurable limits of 16 MiB per SSE event and 4 MiB per POST body by default. Oversized SSE streams terminate with a typed error without reconnecting, and oversized POST bodies receive HTTP 413.

## How Has This Been Tested?
Added regression coverage for unterminated and multi-line SSE events.

## Breaking Changes

No source-level breaking changes are expected.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/common/auth/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/common/auth/streamable_http_client.rs
@@ -47,6 +47,33 @@ where
             .await
     }
 
+    async fn get_stream_with_max_sse_event_size(
+        &self,
+        uri: std::sync::Arc<str>,
+        session_id: std::sync::Arc<str>,
+        last_event_id: Option<String>,
+        mut auth_token: Option<String>,
+        custom_headers: HashMap<HeaderName, HeaderValue>,
+        max_sse_event_size: usize,
+    ) -> Result<
+        futures::stream::BoxStream<'static, Result<sse_stream::Sse, sse_stream::Error>>,
+        crate::transport::streamable_http_client::StreamableHttpError<Self::Error>,
+    > {
+        if auth_token.is_none() {
+            auth_token = Some(self.get_access_token().await?);
+        }
+        self.http_client
+            .get_stream_with_max_sse_event_size(
+                uri,
+                session_id,
+                last_event_id,
+                auth_token,
+                custom_headers,
+                max_sse_event_size,
+            )
+            .await
+    }
+
     async fn post_message(
         &self,
         uri: std::sync::Arc<str>,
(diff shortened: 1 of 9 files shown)


crates/rmcp/src/transport/common/auth/streamable_http_client.rs
```diff
@@ -47,6 +47,33 @@ where
...
+        custom_headers: HashMap<HeaderName, HeaderValue>,
+        max_sse_event_size: usize,
+    ) -> Result<
+        futures::stream::BoxStream<'static, Result<sse_stream::Sse, sse_stream::Error>>,
+        crate::transport::streamable_http_client::StreamableHttpError<Self::Error>,
+    > {
+        if auth_token.is_none() {
+            auth_token = Some(self.get_access_token().await?);
+        }
+        self.http_client
```
alexhancock, line 65: My only Q would be whether we can apply the new defaults at the HTTP client level for all requests, instead of providing these methods for per request type limit setting. If not, feel free to go ahead as-is.

alexhancock (dismissed):

alexhancock (approved):

crates/rmcp/src/transport/common/auth/streamable_http_client.rs
DaleSeo: @alexhancock The value here is actually a single client-wide setting that gets threaded through each call. I see the method signature make it look like a per-request-type limit. I'll add some comment to clear that up.

alexhancock: My question was more why we needed the new wrapper methods, and why the limit couldn't be applied internally in `self.http_client.get_stream_with_max_sse_event_size` for example 

Just not yet understanding

DaleSeo: @alexhancock The limit isn't stored on the client, so there's nothing for `AuthClient` to read internally. It lives on the transport config, and the clients are separate objects passed into `with_client(client, config)`. One of them is `reqwest::Client`, which is a foreign type we can't add a field to, so the value has to come from the worker at call time.

Also, `AuthClient` isn't where the bounding can happen. The cap is applied at the raw byte layer, where `response.bytes_stream()` gets wrapped into an `SseStream`. That only happens in the leaf client (reqwest or unix socket). `AuthClient` just adds the auth token and forwards to the inner client, so all it can do is pass `max_sse_event_size` along.

So the value flows from the worker, through `AuthClient`, down to the client that actually reads the bytes. Dropping the parameter would mean storing the limit on the client, but because it's implemented on the foreign `reqwest::Client` that needs a wrapper type and removing the bare impl. I'd rather keep that as a separate follow-up.

PR #973, opened 2026-07-10
feat!: add server discovery and negotiation (SEP-2575)

### LINKED ISSUE #869: Implement SEP-2575: Make MCP Stateless

DaleSeo:
## Motivation and Context
This implements the discovery and negotiation portion of [SEP-2575](https://modelcontextprotocol.io/seps/2575-stateless-mcp). It adds the `server/discover` RPC, per-request protocol version negotiation, and typed unsupported-version errors so clients can learn a server's versions and capabilities without initializing a session, while preserving the legacy initialization flow.

All six protocol union enums generated by `ts_union!` are now `#[non_exhaustive]`. This requires downstream exhaustive matches to add a wildcard arm once, but allows future protocol variants to be added without causing the same Rust API break.

## How Has This Been Tested?
Added tests with conformance coverage

## Breaking Changes
Yes. `ClientRequest`, `ClientNotification`, `ClientResult`, `ServerRequest`, `ServerNotification`, and `ServerResult` are now `#[non_exhaustive]`, so downstream exhaustive matches must add a wildcard arm. `DiscoverRequest` and `DiscoverResult` also add variants to `ClientRequest` and `ServerResult`; subsequent variants can be added without another exhaustive-match break.

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -524,6 +524,10 @@ pub struct JsonRpcNotification<N = Notification> {
 pub struct ErrorCode(pub i32);
 
 impl ErrorCode {
+    /// The request used a protocol version the server does not support.
+    pub const UNSUPPORTED_PROTOCOL_VERSION: Self = Self(-32022);
+    /// Processing the request requires a client capability that was not declared.
+    pub const MISSING_REQUIRED_CLIENT_CAPABILITY: Self = Self(-32021);
     pub const HEADER_MISMATCH: Self = Self(-32020);
     pub const RESOURCE_NOT_FOUND: Self = Self(-32002);
     pub const INVALID_REQUEST: Self = Self(-32600);
@@ -572,6 +576,30 @@ impl ErrorData {
     }
     pub fn header_mismatch(message: impl Into<Cow<'static, str>>, data: Option<Value>) -> Self {
         Self::new(ErrorCode::HEADER_MISMATCH, message, data)
+    }
+    /// Create an unsupported-protocol-version error.
+    pub fn unsupported_protocol_version(
+        requested: ProtocolVersion,
+        supported: &[ProtocolVersion],
+    ) -> Self {
+        Self::new(
+            ErrorCode::UNSUPPORTED_PROTOCOL_VERSION,
+            "Unsupported protocol version",
+            Some(serde_json::json!({
+                "requested": requested,
+                "supported": supported,
+            })),
+        )
+    }
+    /// Create a missing-required-capability error.
+    pub fn missing_required_client_capability(required: ClientCapabilities) -> Self {
+        Self::new(
+            ErrorCode::MISSING_REQUIRED_CLIENT_CAPABILITY,
+            "Missing required client capability",
+            Some(serde_json::json!({
+                "requiredCapabilities": required,
+            })),
+        )
     }
     pub fn parse_error(message: impl Into<Cow<'static, str>>, data: Option<Value>) -> Self {
         Self::new(ErrorCode::PARSE_ERROR, message, data)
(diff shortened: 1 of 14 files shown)


crates/rmcp/src/model.rs
```diff
@@ -519,6 +519,10 @@ pub struct JsonRpcNotification<N = Notification> {
 pub struct ErrorCode(pub i32);
 
 impl ErrorCode {
+    /// The request used a protocol version the server does not support.
+    pub const UNSUPPORTED_PROTOCOL_VERSION: Self = Self(-32022);
+    /// Processing the request requires a client capability that was not declared.
+    pub const MISSING_REQUIRED_CLIENT_CAPABILITY: Self = Self(-32021);
```
DaleSeo, line 530: I initially set them to `-32004` and `-32003` following the SEP, but it failed the conformance suite. Then, I noticed that the error codes in the SEP differ from the draft schema. I reported this issue in modelcontextprotocol/modelcontextprotocol/issues/3091

- SEP

Image

- Schema

Image

crates/rmcp/src/service/client.rs
```diff
@@ -147,6 +148,19 @@ where
 #[expect(clippy::exhaustive_structs, reason = "intentionally exhaustive")]
 pub struct RoleClient;
 
+/// Select the first client-preferred protocol version supported by the server.
+///
+/// Returns `None` when no version is shared.
```
alexhancock, line 154: What happens in this case? When a client only expresses a preference the server doesn't yet support?

Shouldn't it fall back to the newest one the server does support?

DaleSeo: This is wired up in the follow-up client lifecycle PR, #995. `preferred_versions` contains every version the client supports in preference order. When the server returns `UnsupportedProtocolVersionError`, the client selects the first mutually supported unattempted version and retries. If there is no overlap, it returns `NoCompatibleProtocolVersion`. It intentionally does not select a server version the client does not support. This should match the spec: https://modelcontextprotocol.io/specification/draft/basic/versioning#protocol-version-negotiation

alexhancock (dismissed):

alexhancock (approved):

PR #994, opened 2026-07-15
fix(auth): add an SDK path for pre-registered OAuth clients

### LINKED ISSUE #984: auth/pre-registration: new pre-registration-auth conformance check fails
Part of #977.

On the 2026-07-28 suite (`0.2.0-alpha.9`), `auth/pre-registration` fails 5/6 with check `pre-registration-auth`. The scenario passed 13/13 on the stable `0.1.16` suite, so this is a check added in the alpha suite (the check set also changed shape, 13 → 6 checks).

Work: inspect the failing check's expectations (pre-registered `client_id`/`client_secret` supplied via `MCP_CONFORMANCE_CONTEXT`) and fix the auth flow in the SDK or the conformance client.

Repro:

```bash
  --scenario auth/pre-registration -o results
```

jamadeo:
... and exercise it in conformance tests

fixes #984

## Motivation and Context

The sdk didn't have a high-level path for preregistered oauth clients, so the conformance client implemented it in a more custom way, which didn't pick up the `iss`-parameter preservation.

## How Has This Been Tested?
Ran the conformance client.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -2803,6 +2803,34 @@ impl AuthorizationSession {
         })
     }
 
+    /// create a session using pre-registered client credentials, skipping
+    /// dynamic client registration and URL-based client IDs.
+    ///
+    /// The manager must already have discovered authorization server metadata.
+    ///
+    /// On failure, the manager is returned alongside the error so callers can
+    /// retry without losing the original configuration and stores.
+    pub async fn with_preregistered_client(
+        mut auth_manager: AuthorizationManager,
+        config: OAuthClientConfig,
+    ) -> Result<Self, (AuthorizationManager, AuthError)> {
+        let redirect_uri = config.redirect_uri.clone();
+        let scopes = config.scopes.clone();
+        if let Err(e) = auth_manager.configure_client(config) {
+            return Err((auth_manager, e));
+        }
+        let scope_refs: Vec<&str> = scopes.iter().map(|s| s.as_str()).collect();
+        let auth_url = match auth_manager.get_authorization_url(&scope_refs).await {
+            Ok(url) => url,
+            Err(e) => return Err((auth_manager, e)),
+        };
+        Ok(Self {
+            auth_manager,
+            auth_url,
+            redirect_uri,
+        })
+    }
+
     /// create session for scope upgrade flow (existing manager + pre-computed auth url)
     pub fn for_scope_upgrade(
         auth_manager: AuthorizationManager,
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -3073,6 +3093,36 @@ impl OAuthState {
...
+    /// Use this when the client was registered with the authorization server
+    /// out of band and already holds a `client_id` (and optionally a
+    /// `client_secret`). If `config.scopes` is empty, scopes are selected
+    /// from the discovered authorization server metadata.
+    pub async fn start_authorization_with_preregistered_client(
+        &mut self,
+        mut config: OAuthClientConfig,
+    ) -> Result<(), AuthError> {
+        let placeholder = self.placeholder().await?;
+        if let OAuthState::Unauthorized(mut manager) = std::mem::replace(self, placeholder) {
```
DaleSeo: `std::mem::replace` installs the placeholder before discovery, client configuration, and authorization-URL generation. How should callers retry after a transient discovery/configuration failure without losing the original base URL, stores, and state?

crates/rmcp/src/transport/auth.rs
```diff
@@ -3073,6 +3093,36 @@ impl OAuthState {
         }
     }
 
+    /// start authorization using pre-registered client credentials,
+    /// skipping dynamic client registration.
+    ///
+    /// Use this when the client was registered with the authorization server
+    /// out of band and already holds a `client_id` (and optionally a
+    /// `client_secret`). If `config.scopes` is empty, scopes are selected
+    /// from the discovered authorization server metadata.
```
DaleSeo: `select_scopes` prefers scopes from `WWW-Authenticate` and PRM before ARM, so this is narrower than the implementation.

```suggestion
    /// using the SDK's normal scope-selection policy.
```

crates/rmcp/src/transport/auth.rs
```diff
@@ -2803,6 +2803,26 @@ impl AuthorizationSession {
         })
     }
 
+    /// create a session using pre-registered client credentials, skipping
+    /// dynamic client registration and URL-based client IDs.
+    ///
+    /// The manager must already have discovered authorization server metadata.
+    pub async fn with_preregistered_client(
```
DaleSeo, line 2813: The same impl currently exposes `new` and `for_scope_upgrade`. While `with_*` is valid Rust naming, the three constructors do not seem to communicate an obvious shared convention. I was just thinking we could clean up the API, since v3 allows us to make breaking changes.

crates/rmcp/src/transport/auth.rs
```diff
@@ -3073,6 +3093,36 @@ impl OAuthState {
...
     }
 
+    /// start authorization using pre-registered client credentials,
+    /// skipping dynamic client registration.
+    ///
+    /// Use this when the client was registered with the authorization server
+    /// out of band and already holds a `client_id` (and optionally a
+    /// `client_secret`). If `config.scopes` is empty, scopes are selected
+    /// from the discovered authorization server metadata.
+    pub async fn start_authorization_with_preregistered_client(
```
DaleSeo, line 3111: Could we add targeted tests using the existing recording HTTP client to verify that this path skips the registration endpoint, handles default and explicit scopes, and etc? Those tests would isolate the new orchestration from the broader conformance harness. e.g.

https://github.com/modelcontextprotocol/rust-sdk/blob/98bb2635f4814c0bef671168833820001e5bbf77/crates/rmcp/src/transport/auth.rs#L3870-L3871

DaleSeo, line 3111: [The draft authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization/client-registration) recommends the priority pre-registered client information → CIMD → DCR, but the current API exposes these choices through separate entry points. In particular, `start_authorization()` passes no metadata URL and therefore cannot take the preferred CIMD path, while `start_authorization_with_metadata_url()` is the method that actually implements CIMD with DCR fallback. This PR understandably extends that pattern with another mechanism-specific method, but we might want to consolidates these paths behind one `start_authorization` API. This is out of scope and non-blocking for this PR. Let me know what you think, and we can follow up on this separately.

crates/rmcp/src/transport/auth.rs
jamadeo: Yes, good point, we should take the opportunity. Probably best in a follow-up PR - what do you think? What's your preferred convention?

crates/rmcp/src/transport/auth.rs
jamadeo: That makes sense to me. Probably a good thing to do jointly along with cleaning up the API as you suggested in the above comment.

Opened: https://github.com/modelcontextprotocol/rust-sdk/issues/1006

DaleSeo: @jamadeo Sounds good! Let's fast-follow on this so we can include it in v3 alongside the other breaking changes.

crates/rmcp/src/transport/auth.rs
DaleSeo: My preference would be a single `AuthorizationSession::new(...)` plus chained builder methods for the optional parts. Same shape as `OAuthClientConfig`.

DaleSeo: Oh, I found this: https://rust-lang.github.io/api-guidelines/naming.html Just FYI

crates/rmcp/src/transport/auth.rs
DaleSeo, line 3076: Should the same recovery pattern be applied to the existing paths too?

DaleSeo (approved):

crates/rmcp/src/transport/auth.rs
jamadeo: Yep, that makes sense. Added to #1006

PR #995, opened 2026-07-15
feat: add modern client lifecycle modes (SEP-2575)

DaleSeo:
## Motivation and Context

This builds on PR #973 for discovery and negotiation by letting clients start with `server/discover` instead of the legacy `initialize` handshake. Modern requests carry self-contained client metadata, while auto mode can fall back for legacy servers and the existing legacy lifecycle remains the default.

## How Has This Been Tested?

The modern, auto-fallback, per-request metadata, and stateless Streamable HTTP paths are covered by integration tests.

## Breaking Changes

No breaking changes. Existing clients continue to use the legacy lifecycle flow unless they explicitly select a modern lifecycle mode.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -226,6 +226,14 @@ jobs:
             --scenario sep-2322-client-request-state \
             -o conformance-client-results/mrtr
 
+      - name: Run draft SEP-2575 client scenario
+        run: |
+          npx -y "@modelcontextprotocol/conformance@${DRAFT_CONFORMANCE_VERSION}" client \
+            --command "$(pwd)/target/debug/conformance-client" \
+            --scenario request-metadata \
+            --spec-version draft \
+            -o conformance-client-results/sep-2575
+
       - name: Upload results
         if: always()
         uses: actions/upload-artifact@v7
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -62,7 +62,7 @@ pub enum ClientInitializeError {
         server_supported: Vec<ProtocolVersion>,
     },
 
-    #[error("modern startup requires at least one preferred protocol version")]
+    #[error("discover startup requires at least one preferred protocol version")]
     NoPreferredProtocolVersion,
 
     #[error("Cancelled")]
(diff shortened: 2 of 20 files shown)


.github/workflows/conformance.yml
```diff
@@ -222,6 +222,14 @@ jobs:
             --scenario sep-2322-client-request-state \
             -o conformance-client-results/mrtr
 
+      - name: Run draft SEP-2575 client scenario
```
alexhancock, line 229: How does this combine/relate to https://github.com/modelcontextprotocol/rust-sdk/pull/991/changes#diff-bd4df60e4b8977afa547a4b2976c2aed79b2113b0fffad61810242961e7fc930R925 ?
cc @jamadeo

Ideally we'd just set the conformance tests to run all the scenarios relevant for the new version and see the list of failures get smaller as we check in support for things.

crates/rmcp/src/service/client.rs
```diff
@@ -285,7 +419,114 @@ where
     transport.send(notification).await.map_err(|error| {
         ClientInitializeError::transport::<T>(error, "send initialized notification")
     })?;
-    Ok(serve_inner(service, transport, peer, peer_rx, ct))
+    Ok(())
+}
+
+async fn modern_startup<S, T>(
```
alexhancock: Thoughts renaming the old code `legacy_` instead of the new code `modern_`?

It strikes me it would be easier to deprecate and remove the legacy stuff if done this way vs having to ship all this and then rename it all when we don't want it to be called "modern" any more.

.github/workflows/conformance.yml
DaleSeo: @alexhancock That PR added an interim plumbing so the scenarios could run at all without SDK changes. This PR actually replaces that harness, allowing all routes to go through the real public API.

Regarding running the whole draft suite, I thought about it, but the runner exits with a non-zero status on any failure. So, running all draft scenarios today would keep CI red until the remaining SEPs are completed. I'm fine with that, but we would need to be a bit careful not to miss any regressions.

crates/rmcp/src/service/client.rs
DaleSeo: Makes sense! Thanks for the suggestion. Will rename it.

alexhancock (dismissed):

alexhancock (approved):

PR #997, opened 2026-07-15
feat(conformance): add SEP-2243 header validation tool

### LINKED ISSUE #979: Implement SEP-2243: server-side custom header validation
Part of #977.

SEP-2243 (HTTP Standardization) server-side support. The `http-custom-header-server-validation` scenario fails 0/5 on the 2026-07-28 suite (`0.2.0-alpha.9`):

- `sep-2243-server-no-xmcp-tool` — the conformance server exposes no tool with `x-mcp-header` annotations, which makes the remaining checks untestable
- `sep-2243-server-decode-base64` — decode base64-encoded header values
- `sep-2243-server-validate-param-match` — validate that header values match the designated tool parameters
- `sep-2243-server-reject-invalid-param-chars` — reject invalid characters in header params
- `sep-2243-server-reject-param-mismatch` — reject mismatched header/param values

Work: SDK support for `x-mcp-header` tool annotations on the server side, plus a test tool in `conformance/src/bin/server.rs`.

Repro:

```bash
STATELESS=1 PORT=8002 ./target/debug/conformance-server &
npx -y @modelcontextprotocol/conformance@0.2.0-alpha.9 server \
  --url http://127.0.0.1:8002/mcp \
  --scenario http-custom-header-server-validation -o results
```

jstar0:
## Summary

Expose a conformance tool that exercises server-side SEP-2243 custom parameter header validation.

The Streamable HTTP transport already validates `Mcp-Param-*` headers using the called tool's input schema. The conformance server did not advertise any tool with an `x-mcp-header` annotation, so the `http-custom-header-server-validation` scenario could not exercise that implementation.

## Changes

- Add a `test_custom_header` tool with a required string parameter promoted to `Mcp-Param-Value`.
- Return the same tool definition from `get_tool` so the transport can cache its schema before handler dispatch.
- Expose the tool through `tools/list` and echo the validated value from `tools/call`.
- Add a regression test for the transport-facing tool schema.
- Add a transport regression for a valid Base64 header whose decoded value mismatches the body parameter.
- Run the server unit test and the pinned custom-header scenario in the 2026-07-28 conformance workflow.

The patch reuses the existing transport validation and does not add another header parser or Base64 decoder.

## Verification

```bash
cargo test -p mcp-conformance --all-features
cargo test -p rmcp --test test_streamable_http_standard_headers --features server,client,transport-streamable-http-server,reqwest
cargo test --all-features
cargo fmt --all -- --check
```

The pinned `@modelcontextprotocol/conformance@0.2.0-alpha.9` `http-custom-header-server-validation` scenario also passes all 9 emitted checks, including valid and invalid Base64 values, literal prefix/suffix boundaries, missing custom headers, and JSON-RPC error code `-32020`. The repository integration test separately covers a syntactically valid Base64 value that decodes to a different body value.

Closes #979

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -34,6 +34,9 @@ jobs:
       - name: Build conformance binaries
         run: cargo build -p mcp-conformance
 
+      - name: Test conformance server
+        run: cargo test -p mcp-conformance --bin conformance-server
+
       - name: Start conformance server
         run: |
           PORT=8001 ./target/debug/conformance-server &
@@ -80,7 +83,12 @@ jobs:
 
       - name: Run draft SEP scenarios
         run: |
-          for scenario in sep-2164-resource-not-found caching http-header-validation; do
+          for scenario in \
+            sep-2164-resource-not-found \
+            caching \
+            http-header-validation \
+            http-custom-header-server-validation \
+          ; do
             npx -y "@modelcontextprotocol/conformance@${DRAFT_CONFORMANCE_VERSION}" server \
               --url http://127.0.0.1:8002/mcp \
               --scenario "$scenario" \
--- a/conformance/src/bin/server.rs
+++ b/conformance/src/bin/server.rs
@@ -28,6 +28,20 @@ fn json_object(v: Value) -> JsonObject {
     }
 }
 
+fn custom_header_tool() -> Tool {
+    Tool::new(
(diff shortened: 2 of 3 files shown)


alexhancock (approved):

alexhancock:
Thanks for the contribution of the code needed to pass the conformance test

PR #998, opened 2026-07-16
feat(auth): bind DCR client credentials to issuing authorization server (SEP-2352)

### LINKED ISSUE #879: Implement SEP-2352: Clarify authorization server binding and migration

alexhancock:
## Motivation and Context

Per https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2352 clients SHOULD maintain separate registration state per auth server and MUST NOT assume cross-AS credential validity

## How Has This Been Tested?
Conformance tests
```
npx -y @modelcontextprotocol/conformance@0.2.0-alpha.9 client --command ./target/debug/conformance-client --scenario auth/authorization-server-migration -o results
```

## Breaking Changes
None. `StoredCredentials` is #[non_exhaustive] and has a `new` constructor

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

## Additional context
Closes https://github.com/modelcontextprotocol/rust-sdk/issues/879

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -201,6 +201,8 @@ pub struct StoredCredentials {
     pub granted_scopes: Vec<String>,
     #[serde(default)]
     pub token_received_at: Option<u64>,
+    #[serde(default)]
+    pub issuer: Option<String>,
 }
 
 impl std::fmt::Debug for StoredCredentials {
@@ -213,6 +215,7 @@ impl std::fmt::Debug for StoredCredentials {
             )
             .field("granted_scopes", &self.granted_scopes)
             .field("token_received_at", &self.token_received_at)
+            .field("issuer", &self.issuer)
             .finish()
     }
 }
@@ -230,6 +233,7 @@ impl StoredCredentials {
             token_response,
             granted_scopes,
             token_received_at,
+            issuer: None,
         }
     }
 }
@@ -1088,6 +1092,15 @@ impl AuthorizationManager {
                     self.metadata = Some(metadata);
                 }
 
+                if let (Some(stored_issuer), Some(current_issuer)) =
+                    (stored.issuer.as_deref(), self.metadata_issuer().as_deref())
+                {
+                    if stored_issuer != current_issuer {
+                        self.credential_store.clear().await?;
+                        return Ok(false);
+                    }
+                }
+
                 self.configure_client_id(&stored.client_id)?;
                 return Ok(true);
             }
@@ -1703,6 +1716,7 @@ impl AuthorizationManager {
             token_response: Some(token_result.clone()),
             granted_scopes,
             token_received_at: Some(Self::now_epoch_secs()),
+            issuer: self.metadata_issuer(),
         };
         self.credential_store.save(stored).await?;
 
@@ -1714,6 +1728,10 @@ impl AuthorizationManager {
             .duration_since(UNIX_EPOCH)
             .unwrap_or_default()
             .as_secs()
+    }
+
+    fn metadata_issuer(&self) -> Option<String> {
+        self.metadata.as_ref().and_then(|m| m.issuer.clone())
     }
 
     /// Proactive refresh buffer: refresh tokens this many seconds before they expire
@@ -1838,6 +1856,7 @@ impl AuthorizationManager {
             token_response: Some(token_result.clone()),
             granted_scopes,
             token_received_at: Some(Self::now_epoch_secs()),
+            issuer: self.metadata_issuer(),
         };
         self.credential_store.save(stored).await?;
 
@@ -2658,6 +2677,7 @@ impl AuthorizationManager {
             token_response: Some(token_result.clone()),
             granted_scopes,
             token_received_at: Some(Self::now_epoch_secs()),
+            issuer: self.metadata_issuer(),
         };
         self.credential_store.save(stored).await?;
 
@@ -2780,6 +2800,7 @@ impl AuthorizationManager {
             token_response: Some(token_result.clone()),
             granted_scopes,
             token_received_at: Some(Self::now_epoch_secs()),
+            issuer: self.metadata_issuer(),
         };
         self.credential_store.save(stored).await?;
 
@@ -3147,6 +3168,7 @@ impl OAuthState {
                 token_response: Some(credentials),
                 granted_scopes,
                 token_received_at: Some(AuthorizationManager::now_epoch_secs()),
+                issuer: None,
             };
             manager.credential_store.save(stored).await?;
 
@@ -4617,6 +4639,7 @@ mod tests {
             token_response: Some(token_response),
             granted_scopes: vec![],
             token_received_at: None,
+            issuer: None,
         };
         let debug_output = format!("{:?}", creds);
 
@@ -5594,6 +5617,7 @@ mod tests {
             token_response: Some(make_token_response("my-access-token", Some(3600))),
             granted_scopes: vec![],
             token_received_at: Some(AuthorizationManager::now_epoch_secs()),
+            issuer: None,
         };
         manager.credential_store.save(stored).await.unwrap();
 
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -2517,6 +2536,7 @@ impl AuthorizationManager {
             token_response: Some(token_result.clone()),
             granted_scopes,
             token_received_at: Some(Self::now_epoch_secs()),
+            issuer: self.metadata_issuer(),
```
DaleSeo, line 2719: We now stamp `issuer: self.metadata_issuer()` here, but nothing reads it back. After a migration, what stops a static pre-registered credential from being silently sent to the new AS, the case where SEP-2352 says clients "SHOULD surface an error"?

crates/rmcp/src/transport/auth.rs
```diff
@@ -3006,6 +3027,7 @@ impl OAuthState {
                 token_response: Some(credentials),
                 granted_scopes,
                 token_received_at: Some(AuthorizationManager::now_epoch_secs()),
+                issuer: None,
```
DaleSeo: Since `discover_metadata()` runs right after the save, the issuer is actually available in this function. Was keeping the save before discovery intentional, or could the stamp be filled in after discovery succeeds?

crates/rmcp/src/transport/auth.rs
```diff
@@ -1068,6 +1072,15 @@ impl AuthorizationManager {
                     self.metadata = Some(metadata);
                 }
 
+                if let (Some(stored_issuer), Some(current_issuer)) =
+                    (stored.issuer.as_deref(), self.metadata_issuer().as_deref())
```
DaleSeo, line 1101: SEP-2352 explicitly says CIMD client IDs "are portable across authorization servers … No re-registration is required." The mismatch check here doesn't distinguish credential types.

crates/rmcp/src/transport/auth.rs
```diff
@@ -229,6 +232,7 @@ impl StoredCredentials {
             token_response,
             granted_scopes,
             token_received_at,
+            issuer: None,
```
DaleSeo, line 236: How can external `CredentialStore` implementors set `issuer`?

crates/rmcp/src/transport/auth.rs
```diff
@@ -1068,6 +1072,15 @@ impl AuthorizationManager {
                     self.metadata = Some(metadata);
                 }
 
+                if let (Some(stored_issuer), Some(current_issuer)) =
+                    (stored.issuer.as_deref(), self.metadata_issuer().as_deref())
+                {
+                    if stored_issuer != current_issuer {
+                        self.credential_store.clear().await?;
+                        return Ok(false);
+                    }
```
DaleSeo, line 1140: Clearing stored credentials is destructive. Should we add a warning log here?

crates/rmcp/src/transport/auth.rs
alexhancock: good catch. `initialize_from_store` now returns an AuthorizationServerMismatch error on issuer change

crates/rmcp/src/transport/auth.rs
alexhancock: done

crates/rmcp/src/transport/auth.rs
alexhancock: Done. I added a check for if the client id is url based, which is the case for CIMD

crates/rmcp/src/transport/auth.rs
alexhancock: Added a `with_issuer`

crates/rmcp/src/transport/auth.rs
alexhancock: Done

crates/rmcp/src/transport/auth.rs
```diff
@@ -1088,6 +1097,25 @@ impl AuthorizationManager {
...
+                            stored_issuer,
+                            current_issuer,
+                            "authorization server issuer changed; clearing stored credentials bound to the previous issuer"
+                        );
+                        self.credential_store.clear().await?;
+                        return Err(AuthError::AuthorizationServerMismatch {
+                            expected_issuer: stored_issuer.to_string(),
+                            received_issuer: current_issuer.to_string(),
+                        });
+                    }
```
DaleSeo: Shouldn't we keep the client ID if it's a url (CIMD), since it's supposed to be portable?

```suggestion
                    if stored_issuer != current_issuer {
                        if is_https_url(&stored.client_id) {
                            // A CIMD client ID is the client's metadata URL, so it is
                            // portable across authorization servers — but the tokens
                            // were minted by the previous AS and must not be reused.
                            tracing::warn!(
                                stored_issuer,
                                current_issuer,
                                "authorization server issuer changed; discarding tokens but keeping portable CIMD client ID"
                            );
                            self.credential_store
                                .save(
                                    StoredCredentials::new(
                                        stored.client_id.clone(),
                                        None,
                                        vec![],
                                        None,
                                    )
                                    .with_issuer(self.metadata_issuer()),
                                )
                                .await?;
                            self.configure_client_id(&stored.client_id)?;
                            return Ok(false);
                        }

                        tracing::warn!(
                            stored_issuer,
                            current_issuer,
                            "authorization server issuer changed; clearing stored credentials bound to the previous issuer"
                        );
                        self.credential_store.clear().await?;
                        return Err(AuthError::AuthorizationServerMismatch {
                            expected_issuer: stored_issuer.to_string(),
                            received_issuer: current_issuer.to_string(),
                        });
                    }
```

alexhancock: I had thought calling `self.configure_client_id(&stored.client_id)?` was enough - let me look into it later tonight. Maybe it doesn't call the store.save?

DaleSeo: I think this is more about UX (avoiding re-logins) rather than security, so I'll approve it.

DaleSeo (dismissed):

alexhancock:
I applied your suggestion @DaleSeo but had to squash to satisfy commitlint. Can you re-stamp? TY!

DaleSeo (approved):

PR #999, opened 2026-07-16
fix(server): serve draft-version requests statelessly per SEP-2567

### LINKED ISSUE #870: Implement SEP-2567: Sessionless MCP via Explicit State Handles
# SEP-2567: Sessionless MCP via Explicit State Handles — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567
**Track:** Specification · **Stage:** final · **Priority:** P0 · **Theme:** Transport Evolution and Scalability
**Needs code changes:** Yes (Large)

## Summary
Removes protocol-level sessions, replacing implicit session-scoped state with explicit,
server-minted **state handles**. Builds on SEP-1442 (stateless-by-default) but argues the opt-in
stateful path should not exist at all.

What changes:
- No `session/create`, `session/destroy`, or `Mcp-Session-Id` header.
- `tools/list` / `resources/list` / `prompts/list` **MUST NOT** depend on per-connection or prior-tool-call state (so list endpoints become cacheable at `(deployment, auth)` granularity).
- Stateful workflows use `create_*() -> handle` + threaded parameters (guidance, not a protocol construct).

The streamable HTTP transport mints and threads a session id today, but a disable path already
exists:
- Server: `StreamableHttpServerConfig.stateful_mode` (default `true`); when `false` the server "disables sessions entirely (stateless mode)" per the doc comment, backed by the `SessionManager`/`SessionStore` trait (`session.rs`, with `Local…`/`Never…` stores and `RestoreOutcome`/`SessionRestoreMarker`). The session id rides on `HEADER_SESSION_ID`.
- Client: `streamable_http_client.rs` carries `session_id: Arc<str>` through requests and SSE reconnects.
- List handlers live behind `service/server.rs` + `handler/server/router.rs`.

So "no sessions" is partly reachable via `stateful_mode = false`. The remaining work is: making that
path the default/clean one for `V_2026_07_28`, ensuring list endpoints are state-independent, and
removing the client's hard dependency on a session id.

## Proposed work
- [ ] Ensure the `stateful_mode = false` server path neither requires nor emits `HEADER_SESSION_ID`, and verify the `Never…`-store path is the one selected.
- [ ] Audit `tools/list` / `resources/list` / `prompts/list` handlers (`handler/server/router.rs`) so results don't depend on per-connection or prior-call state in the sessionless path (precondition for the SEP-2549 caching story).
- [ ] Make the client's `session_id` optional (`Option<Arc<str>>`) so it can operate without one when the negotiated version is `V_2026_07_28`; keep sending it for older negotiated versions.
- [ ] Keep `HEADER_SESSION_ID` handling for back-compat behind older protocol versions.
- [ ] Add an example demonstrating the `create_*() -> handle` pattern (state handles are app-level, not a protocol construct).

## Affected areas
`transport/streamable_http_server/session.rs`, `transport/streamable_http_server/tower.rs`,
`transport/streamable_http_client.rs`, `service/server.rs`, `handler/server/router.rs`.

## Notes / risks
- Overlaps heavily with SEP-2575. Treat as one transport-statelessness workstream; this SEP is the application-state half, 2575 is the handshake half.
- Making client `session_id` optional touches several `Arc<str>` call sites in `streamable_http_client.rs` (reconnect/get-stream paths) — mechanical but broad.

---
## Related existing issues
- #863 — Expose negotiated `Mcp-Session-Id` (sessionless work changes how session identity is surfaced)
- #526 — SEP-1442 Make MCP Stateless (same transport-statelessness workstream)

alexhancock:
## Motivation and Context

Per https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567 protocol-level sessions (and the `Mcp-Session-Id` header) are removed from the Streamable HTTP transport as of the draft protocol version (`2026-07-28`); servers speaking that version MUST serve requests statelessly.

The rust-sdk server assigned an `Mcp-Session-Id` in its default `stateful_mode` even when the negotiated protocol version was `2026-07-28`. This gates the session code path on the negotiated version: sessions are used only for pre-draft versions (`< 2026-07-28`), and draft-version requests are served statelessly even in `stateful_mode`. Older versions retain full session behavior.

## How Has This Been Tested?

No dedicated SEP-2567 conformance scenario exists (the conformance PR that would add one was closed unmerged); per the maintainer, every existing draft (`2026-07-28`) scenario already sends sessionless requests. Verified against the existing draft server suite:

```
STATELESS= ./target/debug/conformance-server &   # default stateful_mode
npx -y @modelcontextprotocol/conformance@0.2.0-alpha.9 server --url http://127.0.0.1:8001/mcp --suite draft -o results
```

The default (stateful) server now produces results byte-for-byte identical to an explicit stateless server (48 passed / 31 failed; the 31 remaining failures are pre-existing SEP-2575 `server/discover` gaps, out of scope). `rmcp` lib unit tests (232) and streamable-HTTP integration tests (session store, stale session, protocol version, JSON response — 19) pass locally.

## Breaking Changes

None. The session code path is unchanged for all pre-`2026-07-28` versions; only the draft version changes behavior, bringing it into spec compliance.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Additional context

SEP-2575 (also in progress) removes the `initialize` handshake for the draft version. Once that lands, the version detection here can be simplified to a lifecycle check (an `initialize` request implies the pre-draft/stateful lifecycle; draft requests never carry `initialize`).

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -59,8 +59,13 @@ pub struct StreamableHttpServerConfig {
     pub sse_retry: Option<Duration>,
     /// If true, the server will create a session for each request and keep it alive.
     /// When enabled, SSE priming events are sent to enable client reconnection.
-    pub stateful_mode: bool,
-    /// When true and `stateful_mode` is false, the server prefers
+    ///
+    /// Only applies to legacy protocol versions (`< 2026-07-28`). Per SEP-2567,
+    /// sessions are removed from the `2026-07-28` draft version, so requests
+    /// negotiating that version are always served statelessly regardless of
+    /// this setting.
+    pub legacy_session_mode: bool,
+    /// When true and `legacy_session_mode` is false, the server prefers
     /// `Content-Type: application/json` for simple request-response tools.
     /// If the handler emits a notification or request before the final response,
     /// the server falls back to `text/event-stream` so no message is lost.
@@ -130,7 +135,7 @@ impl Default for StreamableHttpServerConfig {
         Self {
             sse_keep_alive: Some(Duration::from_secs(15)),
             sse_retry: Some(Duration::from_secs(3)),
-            stateful_mode: true,
+            legacy_session_mode: true,
             json_response: false,
             cancellation_token: CancellationToken::new(),
             allowed_hosts: vec!["localhost".into(), "127.0.0.1".into(), "::1".into()],
@@ -176,8 +181,8 @@ impl StreamableHttpServerConfig {
         self
     }
 
-    pub fn with_stateful_mode(mut self, stateful: bool) -> Self {
-        self.stateful_mode = stateful;
+    pub fn with_legacy_session_mode(mut self, legacy_session_mode: bool) -> Self {
+        self.legacy_session_mode = legacy_session_mode;
         self
     }
 
@@ -690,7 +695,7 @@ fn validate_origin_header(
 ///
 /// ## Session management
 ///
-/// When [`StreamableHttpServerConfig::stateful_mode`] is `true` (the default),
+/// When [`StreamableHttpServerConfig::legacy_session_mode`] is `true` (the default),
 /// the server creates a session for each client that sends an `initialize`
 /// request. The session ID is returned in the `Mcp-Session-Id` response header
 /// and the client must include it on all subsequent requests.
@@ -1186,13 +1191,13 @@ where
             return response;
         }
         let method = request.method().clone();
-        let allowed_methods = match self.config.stateful_mode {
+        let allowed_methods = match self.config.legacy_session_mode {
             true => "GET, POST, DELETE",
             false => "POST",
         };
-        let result = match (method, self.config.stateful_mode) {
+        let result = match (method, self.config.legacy_session_mode) {
             (Method::POST, _) => self.handle_post(request).await,
-            // if we're not in stateful mode, we don't support GET or DELETE because there is no session
+            // if legacy session mode is disabled, we don't support GET or DELETE because there is no session
             (Method::GET, true) => self.handle_get(request).await,
             (Method::DELETE, true) => self.handle_delete(request).await,
             _ => {
@@ -1372,7 +1377,7 @@ where
         };
 
         let use_session =
-            self.config.stateful_mode && is_legacy_request(Some(&message), &part.headers);
+            self.config.legacy_session_mode && is_legacy_request(Some(&message), &part.headers);
 
         if use_session {
             // do we have a session id?
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -1340,7 +1340,31 @@ where
             Err(response) => return Ok(response),
         };
 
-        if self.config.stateful_mode {
+        // SEP-2567: sessions are removed at the protocol level from 2026-07-28.
+        // Serve such requests statelessly even in stateful mode, never assigning
+        // or requiring an Mcp-Session-Id. `initialize` declares the version in its
+        // body params; every other request in the `MCP-Protocol-Version` header
+        // (defaulting to `2025-03-26` when absent).
+        let init_version = match &message {
+            ClientJsonRpcMessage::Request(req) => match &req.request {
+                ClientRequest::InitializeRequest(init) => {
+                    Some(init.params.protocol_version.clone())
+                }
+                _ => None,
+            },
+            _ => None,
+        };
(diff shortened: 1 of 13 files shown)


crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -1309,7 +1162,31 @@ where
...
+        };
+        let declared_version = init_version.unwrap_or_else(|| {
+            part.headers
+                .get(HEADER_MCP_PROTOCOL_VERSION)
+                .and_then(|value| value.to_str().ok())
+                .and_then(|s| serde_json::from_value(serde_json::Value::String(s.to_owned())).ok())
+                .unwrap_or(ProtocolVersion::V_2025_03_26)
+        });
+        let use_session =
+            self.config.stateful_mode && declared_version < ProtocolVersion::V_2026_07_28;
```
DaleSeo: Since #973 has landed, could we classify the request as legacy or modern before touching the session manager, then derive `use_session` from that classification instead of ordering `ProtocolVersion`? I'm thinking of something like `stateful_mode && lifecycle.is_legacy()`. This would also let malformed or unsupported versions fail before any session lookup or restore.

crates/rmcp/src/transport/streamable_http_server/tower.rs
DaleSeo, line 1357: GET and DELETE still enter legacy session handlers?

DaleSeo, line 1196: Since `2026-07-28` requests are always stateless, should we deprecate `stateful_mode` in favor of a legacy-scoped option such as `legacy_session_mode`? After this change, the current option only affects pre-`2026-07-28` traffic, but its name suggests that it controls all requests. Since v3 already allows breaking changes, this may be a good opportunity to clarify the API.

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -1204,6 +1237,9 @@ where
                 )
                 .expect("valid response"));
         }
+        if !is_legacy_request(None, request.headers()) {
+            return Ok(method_not_allowed_response());
+        }
```
DaleSeo: We should move this check to the start of `handle_get()` to make draft GET behavior independent of legacy SSE validation.

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -250,6 +255,34 @@ fn message_has_per_request_protocol_version(message: &ClientJsonRpcMessage) -> b
     }
 }
 
+// SEP-2567: sessions are removed from 2026-07-28; older versions are legacy.
+fn is_legacy_request(message: Option<&ClientJsonRpcMessage>, headers: &HeaderMap) -> bool {
```
DaleSeo: Could this helper return `Result` and perform the existing header/meta validation before any caller touches the session manager?

https://github.com/modelcontextprotocol/rust-sdk/blob/80e9e93213786a441eb22ecbb36312c9d4f538e9/crates/rmcp/src/transport/streamable_http_server/tower.rs#L215

https://github.com/modelcontextprotocol/rust-sdk/blob/80e9e93213786a441eb22ecbb36312c9d4f538e9/crates/rmcp/src/transport/streamable_http_server/tower.rs#L317

https://github.com/modelcontextprotocol/rust-sdk/blob/80e9e93213786a441eb22ecbb36312c9d4f538e9/crates/rmcp/src/transport/streamable_http_server/tower.rs#L351

crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -59,8 +59,13 @@ pub struct StreamableHttpServerConfig {
     pub sse_retry: Option<Duration>,
     /// If true, the server will create a session for each request and keep it alive.
     /// When enabled, SSE priming events are sent to enable client reconnection.
-    pub stateful_mode: bool,
```
DaleSeo, line 62: Since this is a breaking change, the commit message should contains `!` to pass CI like `refactor!: rename stateful_mode to legacy_session_mode`.

crates/rmcp/src/transport/streamable_http_server/tower.rs
alexhancock: Take a look at d6eb1404ed2250f4a77d7dce0709d939ee0fa586 and let me know if it matches your expectation of how it would work

DaleSeo: Perfect. Thanks!

DaleSeo (approved):

PR #1011, opened 2026-07-18
fix: re-register after auth server change

DaleSeo:
## Motivation and Context

This is a quick follow-up to #1008. It routed `auth/authorization-server-migration` through the generic auth flow, which made its dedicated conformance path unreachable. Restoring that path exposed a second issue in #998: after clearing the DCR credentials tied to the previous authorization server, `initialize_from_store` returned an issuer-mismatch error instead of allowing registration with the new server. This completes the authorization-server migration flow and adds a focused regression test.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/conformance/expected-failures-2026-07-28.yaml
+++ b/conformance/expected-failures-2026-07-28.yaml
@@ -23,6 +23,3 @@ client:
   # Auth feature gaps in the 2026-07-28 auth scenarios.
   # tracked in #1002
   - auth/scope-step-up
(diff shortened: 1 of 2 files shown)


alexhancock (approved):

alexhancock:
Only thing I can see is that now callers can no longer distinguish "AS changed, creds discarded" from "no creds present" as both now return `Ok(false)` but seems ok

PR #1010, opened 2026-07-18
chore: run full conformance CI suite

jamadeo:
Runs the full conformance suite, enumerating expected failures.

closes #985 

## Motivation and Context
Track conformance for new spec.

## How Has This Been Tested?
Run the conformance tests locally and on CI.

## Breaking Changes
none

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1013, opened 2026-07-19
fix(conformance): use discover lifecycle for 2026 clients

### LINKED ISSUE #1002: Conformance client: legacy-lifecycle scenarios fail with -32020 missing MCP-Protocol-Version header at 2026-07-28
Several 2026-07-28 client scenarios drive the regular streamable HTTP client through the legacy `initialize` lifecycle. The suite's mock server requires the `MCP-Protocol-Version` header on every POST at this spec version, and the legacy lifecycle does not send it before negotiation completes, so the run dies at `initialize`:

```
{"jsonrpc":"2.0","id":0,"error":{"code":-32020,"message":"Missing MCP-Protocol-Version header"}}
```

Affected scenarios (verified at current main 2e2c791; identical at pre-#973 839922d, so not a regression from the SEP-2575 merges):

- `tools_call` — 0/1 (passed in the 2026-07-13 baseline in #977)
- `auth/scope-step-up` — 11/13; dies before the escalation round, so `scope-step-up-escalation` and `sep-2350-scope-union-on-reauth` never run. **This means the suite cannot currently verify #888 (SEP-2350, merged 2026-07-16).**
- `http-standard-headers` — 4/5; fails `sep-2243-client-includes-standard-headers` ("Missing Mcp-Method header on initialize request")

Likely fix: switch these scenario paths in `conformance/src/bin/client.rs` to the modern discover lifecycle added by #995 (`ClientLifecycleMode::Discover` / `Auto`) when running at 2026-07-28, which sends per-request metadata and the version header on every request. `request-metadata` already passes 5/5 through that path.

Repro:

```bash
  --scenario auth/scope-step-up --spec-version 2026-07-28
```

Part of #977. Related: #978 (client-side SEP-2243 header work).

0xzr:
## Summary

- select the conformance client lifecycle from the runner's protocol version, using discovery for 2026-07-28 while preserving initialization for legacy runners
- exercise the discovery lifecycle in the affected tools, scope step-up, and standard-header scenarios
- request the initial `mcp:basic` scope required by the draft scope step-up server
- run the three affected draft scenarios in the conformance workflow

Closes #1002

## Testing

- `cargo test -p mcp-conformance --bin conformance-client` (3 passed)
- 2025-11-25 client suite with conformance 0.1.16 (222/222 checks passed)
- 2026-07-28 `tools_call` with conformance 0.2.0-alpha.9 (1/1 check passed)
- 2026-07-28 `auth/scope-step-up` with conformance 0.2.0-alpha.9 (30/30 checks passed)
- 2026-07-28 `http-standard-headers` with conformance 0.2.0-alpha.9 (3/3 applicable checks passed)

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -234,6 +234,15 @@ jobs:
             --spec-version draft \
             -o conformance-client-results/sep-2575
 
+      - name: Run draft lifecycle client scenarios
+        run: |
+          for scenario in tools_call auth/scope-step-up http-standard-headers; do
+            npx -y "@modelcontextprotocol/conformance@${DRAFT_CONFORMANCE_VERSION}" client \
+              --command "$(pwd)/target/debug/conformance-client" \
+              --scenario "$scenario" \
+              --spec-version 2026-07-28
+          done
+
       - name: Upload results
         if: always()
         uses: actions/upload-artifact@v7
--- a/conformance/src/bin/client.rs
+++ b/conformance/src/bin/client.rs
@@ -189,6 +189,7 @@ impl ClientHandler for FullClientHandler {
 
 const CIMD_CLIENT_METADATA_URL: &str = "https://conformance-test.local/client-metadata.json";
 const REDIRECT_URI: &str = "http://localhost:3000/callback";
+const SCOPE_STEP_UP_INITIAL_SCOPES: &[&str] = &["mcp:basic"];
 const SCOPE_STEP_UP_ESCALATED_SCOPES: &[&str] = &["mcp:basic", "mcp:write"];
 
 /// Perform the headless OAuth authorization-code flow.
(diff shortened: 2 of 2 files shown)


.github/workflows/conformance.yml
```diff
@@ -234,6 +234,15 @@ jobs:
             --spec-version draft \
             -o conformance-client-results/sep-2575
 
+      - name: Run draft lifecycle client scenarios
```
alexhancock, line 237: We're going to make a change shortly where we run the whole client suite

https://github.com/modelcontextprotocol/rust-sdk/pull/1010/changes#diff-e5ed24bc2f5084a1664419e35473c15466f87e7552ef1fad4d13cdde0eaed400

Mind backing this part out of the change and just doing the `.rs` changes?

alexhancock (commented):
Can you look into the merge conflicts?

alexhancock:
Given no reply on the feedback above I made some related changes in https://github.com/modelcontextprotocol/rust-sdk/pull/1022

Thanks for finding this originally!

PR #1022, opened 2026-07-21
fix: conformance tests for tools_call and auth/scope-step-up

### LINKED ISSUE #1002: Conformance client: legacy-lifecycle scenarios fail with -32020 missing MCP-Protocol-Version header at 2026-07-28

alexhancock:
* Updates the lifecycle used for `tools_call` and `auth/scope-step-up` conformance tests
* Removes them from `expected-failures-2026-07-28.yaml`
* Fixes #1002.


conformance/src/bin/client.rs
```diff
@@ -930,7 +934,21 @@ fn conformance_protocol_version() -> ProtocolVersion {
     std::env::var("MCP_CONFORMANCE_PROTOCOL_VERSION")
         .ok()
         .and_then(|version| serde_json::from_value(Value::String(version)).ok())
-        .unwrap_or(ProtocolVersion::V_2026_07_28)
+        .unwrap_or(ProtocolVersion::V_2025_11_25)
```
alexhancock, line 937: The node module defaults to this in its code for this ENV var

https://github.com/modelcontextprotocol/conformance/blob/da56f6631c83bec2b3b7e40a82f392ae360b790c/src/types.ts#L43

https://github.com/modelcontextprotocol/conformance/blob/da56f6631c83bec2b3b7e40a82f392ae360b790c/src/runner/client.ts#L45

It is set to the new version when the command is run with `--spec-version 2026-07-28` so we still have coverage of the new way

conformance/src/bin/client.rs
```diff
@@ -398,7 +400,9 @@ async fn run_auth_scope_step_up_client(
                     auth_client2,
                     StreamableHttpClientTransportConfig::with_uri(server_url),
                 );
-                let client2 = BasicClientHandler.serve(transport2).await?;
+                let client2 = BasicClientHandler
+                    .serve_with_lifecycle(transport2, conformance_lifecycle())
+                    .await?;
                 let _ = client2
                     .call_tool(call_tool_params(tool.name.clone(), args))
                     .await;
```
DaleSeo: We should propagate the error here as well so the scenario validates the final observable outcome.

```suggestion
client2
.call_tool(call_tool_params(tool.name.clone(), args))
.await?;
```

conformance/expected-failures-2026-07-28.yaml
DaleSeo, line 1: How about we delete this file and remove the following lines?

https://github.com/modelcontextprotocol/rust-sdk/blob/8803d39034b6a7590151ab8fbf9b9cbe7c337bb4/.github/workflows/conformance.yml#L94

https://github.com/modelcontextprotocol/rust-sdk/blob/8803d39034b6a7590151ab8fbf9b9cbe7c337bb4/.github/workflows/conformance.yml#L168

conformance/src/bin/client.rs
```diff
@@ -930,7 +934,21 @@ fn conformance_protocol_version() -> ProtocolVersion {
     std::env::var("MCP_CONFORMANCE_PROTOCOL_VERSION")
         .ok()
         .and_then(|version| serde_json::from_value(Value::String(version)).ok())
-        .unwrap_or(ProtocolVersion::V_2026_07_28)
+        .unwrap_or(ProtocolVersion::V_2025_11_25)
+}
+
+fn uses_discover_lifecycle() -> bool {
+    conformance_protocol_version().as_str() >= ProtocolVersion::V_2026_07_28.as_str()
+}
```
DaleSeo: nit: I guess this isn't really worth having its own function?

DaleSeo (dismissed):
I have minor suggestions.

conformance/src/bin/client.rs
alexhancock: Applied

conformance/expected-failures-2026-07-28.yaml
alexhancock: Done

conformance/src/bin/client.rs
alexhancock: Good call - done

alexhancock:
@DaleSeo All feedback addressed

jamadeo (approved):

PR #975, opened 2026-07-12
Feat/issue 974 client response cache

starsaintf:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1025, opened 2026-07-22
feat: add client-side TTL-honoring response cache (SEP-2549)

### LINKED ISSUE #974: Implement client-side TTL-honoring cache (SEP-2549 follow-up)
# Client-side TTL-honoring cache — SEP-2549 follow-up

**Parent:** #875 (model fields landed in #889)
**Spec:** https://modelcontextprotocol.io/specification/draft/server/utilities/caching
**SEP:** https://modelcontextprotocol.io/seps/2549-TTL-for-list-results
**Track:** Specification · **Stage:** final · **Priority:** P2 (optional fast follow)
**Needs code changes:** Yes (Medium) — additive client behavior

## Summary

#889 added `ttlMs` / `cacheScope` on cacheable results (`tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`, `resources/read`). Clients can already *read* those fields; this issue covers *honoring* them with an automatic response cache in the client service layer.

This is optional per the SEP (TTL is a freshness hint). TypeScript and Python SDKs already ship a built-in client cache; rust-sdk does not yet.

## Why

Without a client cache, every `list_*` / `read_resource` call still hits the wire even when the server declared a positive TTL. A TTL-honoring cache:

- Reduces round trips for stable feature lists
- Complements (does not replace) `list_changed` / `resources/updated` notifications
- Improves behavior on transports where long-lived SSE is unavailable

## Proposed work

- [ ] Add a client response cache in `crates/rmcp/src/service/client.rs` (or a dedicated module) that stores cacheable results keyed by **method + request params** that affect the result (e.g. `uri` for `resources/read`, `cursor` for paginated lists).
- [ ] Honor `ttlMs` freshness: fresh while `now < t_received + ttlMs`; treat absent / negative / `0` as immediately stale.
- [ ] Honor `cacheScope`:
  - `"public"` — may be shared across callers
  - `"private"` — MUST be partitioned by authorization context; MUST NOT serve across different auth contexts
- [ ] Invalidate on relevant notifications (`tools/list_changed`, `prompts/list_changed`, `resources/list_changed`, `resources/updated`) even if TTL has not expired.
- [ ] Do **not** cache multi-round-trip retry results (`inputResponses` / `requestState`).
- [ ] Treat each paginated page as an independently cacheable entry; discard all pages for a list if a cursor becomes invalid.
- [ ] Opt-in / configurable defaults (e.g. default TTL when server omits `ttlMs`, enable/disable cache) — align with TS/Python SDK knobs where practical.
- [ ] Tests covering hit/miss, expiry, notification invalidation, `private` isolation, and pagination keys.
- [ ] Wire / extend conformance `caching` scenario coverage as needed.

## Non-goals

- Changing wire types (already done in #889)
- Requiring servers to set non-zero TTLs
- Background polling on TTL expiry (spec: check freshness on access; do not treat TTL as a poll interval)

## References

- Spec: https://modelcontextprotocol.io/specification/draft/server/utilities/caching
- SEP-2549: https://modelcontextprotocol.io/seps/2549-TTL-for-list-results
- Prior art: [TypeScript client caching](https://ts.sdk.modelcontextprotocol.io/v2/clients/caching.html), [Python client caching](https://py.sdk.modelcontextprotocol.io/v2/client/caching/)
- Model landing PR: #889
- Parent issue: #875

alexhancock:
Add a configurable client response cache in the service layer that honors SEP-2549 caching hints (ttlMs / cacheScope) for tools/list, prompts/list, resources/list, resources/templates/list, and resources/read.

Closes #974

## Motivation and Context
Needed to support https://modelcontextprotocol.io/specification/draft/server/utilities/caching

## How Has This Been Tested?
CI/Tests

## Breaking Changes
N/A

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

## Additional context
Original PR (heavily adapted for this): https://github.com/modelcontextprotocol/rust-sdk/pull/975
https://github.com/starsaintf Marked as Co-Author

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -1342,13 +1342,14 @@ impl Peer<RoleClient> {
         let result = match result {
             Ok(result) => result,
             Err(error) => {
+                if uses_cursor {
+                    self.invalidate_prompt_cache().await;
+                    return Err(error);
+                }
                 if let Some(ServerResult::ListPromptsResult(result)) =
                     self.stale_cached_response(&cache_key).await
                 {
                     return Ok(result);
-                }
-                if uses_cursor {
-                    self.invalidate_prompt_cache().await;
                 }
                 return Err(error);
             }
@@ -1391,14 +1392,15 @@ impl Peer<RoleClient> {
         let result = match result {
             Ok(result) => result,
             Err(error) => {
+                if uses_cursor {
+                    self.invalidate_cached_responses(RESOURCE_LIST_CACHE_PREFIX)
+                        .await;
+                    return Err(error);
+                }
                 if let Some(ServerResult::ListResourcesResult(result)) =
                     self.stale_cached_response(&cache_key).await
                 {
                     return Ok(result);
-                }
-                if uses_cursor {
-                    self.invalidate_cached_responses(RESOURCE_LIST_CACHE_PREFIX)
-                        .await;
                 }
                 return Err(error);
             }
@@ -1443,14 +1445,15 @@ impl Peer<RoleClient> {
         let result = match result {
             Ok(result) => result,
             Err(error) => {
+                if uses_cursor {
+                    self.invalidate_cached_responses(RESOURCE_TEMPLATE_LIST_CACHE_PREFIX)
+                        .await;
+                    return Err(error);
+                }
                 if let Some(ServerResult::ListResourceTemplatesResult(result)) =
                     self.stale_cached_response(&cache_key).await
                 {
                     return Ok(result);
-                }
-                if uses_cursor {
-                    self.invalidate_cached_responses(RESOURCE_TEMPLATE_LIST_CACHE_PREFIX)
-                        .await;
                 }
                 return Err(error);
             }
@@ -1502,13 +1505,14 @@ impl Peer<RoleClient> {
         let result = match result {
             Ok(result) => result,
             Err(error) => {
+                if uses_cursor {
+                    self.invalidate_tool_cache().await;
+                    return Err(error);
+                }
                 if let Some(ServerResult::ListToolsResult(result)) =
                     self.stale_cached_response(&cache_key).await
                 {
                     return Ok(result);
-                }
-                if uses_cursor {
-                    self.invalidate_tool_cache().await;
                 }
                 return Err(error);
             }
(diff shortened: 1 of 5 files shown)


crates/rmcp/src/service/client.rs
```diff
@@ -820,6 +843,52 @@ where
...
+    DISCOVER_CACHE_PREFIX.to_string()
+}
+
+fn list_response_cache_key(prefix: &str, params: &Option<PaginatedRequestParams>) -> String {
+    // Only the pagination cursor affects which page is returned.
+    let cursor = params.as_ref().and_then(|params| params.cursor.as_deref());
+    let cursor =
+        serde_json::to_string(&cursor).expect("serializing a pagination cursor cannot fail");
+    format!("{prefix}{cursor}")
+}
```
DaleSeo, line 870: `RequestMetaObject` permits arbitrary extension keys that handlers can use to vary a result, so two calls with the same cursor but different `_meta` can currently share one cached response. How should result-affecting metadata participate in cache identity without fragmenting entries on transport-only fields such as progress and trace context?

crates/rmcp/src/service/client.rs
```diff
@@ -1118,7 +1276,215 @@ impl Peer<RoleClient> {
...
+            Ok(result) => result,
+            Err(error) => {
+                if let Some(ServerResult::ListToolsResult(result)) =
+                    self.stale_cached_response(&cache_key).await
+                {
+                    return Ok(result);
+                }
+                if uses_cursor {
+                    self.invalidate_tool_cache().await;
+                }
```
DaleSeo: With the default stale-on-error setting, an expired cursor page remains available when the server rejects that cursor, so this early return bypasses the page-set invalidation below.

crates/rmcp/src/service/client.rs
```diff
@@ -1590,3 +1956,148 @@ where
...
+            .cache_response(
+                key.clone(),
+                ServerResult::ListToolsResult(tools_result(Some(5_000), Some(CacheScope::Private))),
+                Some(5_000),
+                Some(CacheScope::Private),
+            )
+            .await;
+
+        assert!(first.cached_response(&key).await.is_some());
+        assert!(second.cached_response(&key).await.is_none());
```
DaleSeo: Separate peers always own separate cache maps, so this test passes even if `private_partition` is ignored. Could it exercise two authorization partitions on the same peer to cover the isolation guarantee introduced by this cache?

crates/rmcp/src/service/client.rs
DaleSeo: Agreed not to address it offline for now.

DaleSeo (approved):

Issue #873, opened 2026-06-02
Implement SEP-2260: Server requests must associate with a client request


alexhancock:
# SEP-2260: Require Server requests to be associated with a Client request — rust-sdk implementation

**Spec PR:** https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2260
**Track:** Specification · **Stage:** final · **Priority:** P0 · **Theme:** Transport Evolution and Scalability
**Needs code changes:** Yes (Medium) — **Breaking** (forbids unsolicited server→client requests)

## Summary
Tightens scoping so that server→client requests must be associated with an in-flight client
request, and serves notice that "unsolicited" server-to-client requests will not be supported by
future stateless transports. Spec-only text change, but it changes allowed runtime behavior.

## Why this needs code changes in rust-sdk
Server→client requests in rmcp go through the `Peer` API (the `send_request` machinery in
`service.rs`), exposed on the server side via `service/server.rs` (e.g. `create_message` for
sampling, `create_elicitation`, `list_roots`, `ping`). Today these can be invoked from a server
`ServerHandler` at any time — including outside the handling of a specific client request — because
the `Peer` doesn't carry an "originating client request" context. To conform, that context needs to
be threaded into the server's outbound request path.

The `RequestContext` passed to handler methods is the natural carrier for the originating request
id; the issue is that the free-standing `Peer` handle outlives any single request, so unsolicited
sends are currently possible.

## Proposed work
- [ ] Enumerate the server→client request entry points on the server `Peer` / `service/server.rs` (`create_message`, elicitation, `list_roots`, `ping`, and any notification-vs-request paths).
- [ ] Thread the originating client request id (available in `RequestContext`) through to outbound server→client requests so each is associated with an in-flight client request.
- [ ] When no originating request exists (a bare `Peer` send outside request handling), guard it: under `V_2026_07_28` return/log an error; under older versions keep the legacy behavior.
- [ ] Carry the association over the wire using the SEP-2322 MRTR framing (don't invent a second mechanism).
- [ ] Gate the stricter behavior behind `ProtocolVersion::V_2026_07_28`.
- [ ] Update examples/docs that rely on unsolicited server-initiated requests.

## Affected areas
`crates/rmcp/src/service.rs` (`Peer` send-request path, `RequestContext`), `service/server.rs`
(sampling/elicitation/roots/ping helpers), `transport/streamable_http_*` (association on the wire).

## Notes / risks
[...]

PR #1027, opened 2026-07-22
feat: Implement SEP-2260 require server requests to associate with client requests

alexhancock:
Implements [SEP-2260](https://modelcontextprotocol.io/seps/2260-Require-Server-requests-to-be-associated-with-Client-requests.md).

Tracking issue: #873

## What

Servers **MUST** send `sampling/createMessage`, `elicitation/create`, and `roots/list` requests only in association with an originating client request (e.g. during `tools/call`, `resources/read`, or `prompts/get` processing). Standalone server-initiated requests of these types are not supported. `ping` is excepted.

## How

- Added a `tokio::task_local!` (`ORIGINATING_REQUEST`) that is scoped around each inbound client-request handler in the serve loop, so any server→client request issued while handling a client request is recognized as associated.
- Added `Peer<RoleServer>::ensure_request_association`, called from the server-side outbound request helpers (`create_message`, `create_elicitation`, `create_elicitation_with_timeout`, `list_roots`). When the negotiated protocol version is `2026-07-28` or newer and the call is made outside a request-handler scope, it returns an error instead of sending an unassociated request.
- `ping` uses a separate path and is not affected.
- Gated behind `ProtocolVersion::V_2026_07_28` so older peers keep legacy behavior.

## Notes

- Client-side behavior is unchanged (the SEP requires no client changes).
- The canonical nested pattern (sampling/elicitation/roots inside a tool/resource/prompt handler) remains fully supported.

## Test

Added `test_sep_2260_request_association` verifying nested sampling inside `call_tool` succeeds while a standalone (spawned) sampling request outside the handler scope is rejected under `2026-07-28`.

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -149,6 +149,15 @@ pub trait ServiceRole: std::fmt::Debug + Send + Sync + 'static + Copy + Clone {
     ) -> impl Future<Output = ()> + MaybeSendFuture {
         async {}
     }
+
+    #[doc(hidden)]
+    fn enforce_request_association(
+        _request: &Self::Req,
+        _peer_info: Option<&Self::PeerInfo>,
+        _in_request_handler_scope: bool,
+    ) -> Result<(), ServiceError> {
+        Ok(())
+    }
 }
 
 pub(crate) fn uses_legacy_lifecycle(
@@ -157,6 +166,14 @@ pub(crate) fn uses_legacy_lifecycle(
 ) -> bool {
     !uses_discover_lifecycle
         && protocol_version.is_none_or(|version| version < &ProtocolVersion::V_2026_07_28)
+}
+
+tokio::task_local! {
+    pub(crate) static ORIGINATING_REQUEST: RequestId;
+}
+
+pub(crate) fn in_request_handler_scope() -> bool {
+    ORIGINATING_REQUEST.try_with(|_| ()).is_ok()
 }
 
 pub type TxJsonRpcMessage<R> =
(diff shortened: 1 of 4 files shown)


crates/rmcp/src/service/server.rs
```diff
@@ -49,6 +49,31 @@ impl ServiceRole for RoleServer {
...
+            ServerRequest::CreateMessageRequest(_)
+                | ServerRequest::ListRootsRequest(_)
+                | ServerRequest::ElicitRequest(_)
+        );
+        if !restricted {
+            return Ok(());
+        }
+        let strict =
+            peer_info.is_some_and(|info| info.protocol_version >= ProtocolVersion::V_2026_07_28);
+        if strict && !in_request_handler_scope {
```
DaleSeo, line 69: Handshake-free services can enter `serve_inner` with `peer_info == None` and take the protocol version from each request's metadata

https://github.com/modelcontextprotocol/rust-sdk/blob/d772e4cb0386c4f4820d48c5d74946f2abc26f69/crates/rmcp/src/service/server.rs#L548-L564

Here that makes `strict` false even after the stateless lifecycle has been selected. How should an unassociated send be rejected on that supported path when no handshake-derived `PeerInfo` exists?

crates/rmcp/src/task_manager.rs
```diff
@@ -349,8 +349,11 @@ impl TaskManager {
         let future = make_future(context);
         let inner = self.inner.clone();
         let id_for_task = task_id.clone();
+        let originating_request = crate::service::ORIGINATING_REQUEST
+            .try_with(|id| id.clone())
+            .ok();
         let handle = tokio::spawn(async move {
-            let result = future.await;
+            let result = run_task_operation(originating_request, future).await;
```
DaleSeo, line 356: `TaskManager::spawn` returns the task to the handler immediately, so this child future can con
tinue after the originating JSON-RPC response has closed. Re-establishing the task-local makes
 a later direct sampling or elicitation pass the guard anyway.

crates/rmcp/src/service.rs
```diff
@@ -1398,9 +1420,10 @@ where
                             extensions,
                         };
                         let current_span = tracing::Span::current();
+                        let handler_id = id.clone();
                         spawn_service_task(async move {
-                            let result = service
-                                .handle_request(request, context)
+                            let result = ORIGINATING_REQUEST
+                                .scope(handler_id, service.handle_request(request, context))
```
DaleSeo, line 1426: Since Tokio task-local values are not inherited by `tokio::spawn`, a handler that spawns sampling work and awaits its `JoinHandle` before returning is rejected even though the parent client request is still in flight?

alexhancock:
These changes are now included in https://github.com/modelcontextprotocol/rust-sdk/pull/1029 with more of what we need! Closing this one

PR #1028, opened 2026-07-22
fix(auth): preserve OAuth discovery transport errors

tjni:
## Motivation and Context

OAuth discovery currently treats network and server failures as missing metadata, which can incorrectly make a server appear not to support OAuth even when the root cause could be a client-side issue or a server-side server error.

This change propagates such errors to clients so that they can be displayed. I think it aligns more with how this SDK in other languages (I checked Go) behave.

## How Has This Been Tested?

Currently, in Codex, adding Linear MCP server in a node:22-bookworm-slim container without ca-certificates installed shows the Linear MCP server as being unavailable even though the root cause is missing CA certificates on the system. With this change, the error is now clearer (even though Codex needs some change too in order to surface it properly).

## Breaking Changes

N/A

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -78,16 +78,32 @@ impl OAuthHttpRequest {
 
 /// Error returned by a custom OAuth HTTP client.
 #[derive(Debug, Error)]
-#[error("{message}")]
+#[error(transparent)]
 pub struct OAuthHttpClientError {
-    message: String,
+    inner: OAuthHttpClientErrorKind,
+}
+
+#[derive(Debug, Error)]
+enum OAuthHttpClientErrorKind {
+    #[error("{0}")]
+    Message(String),
+
+    #[error("{0}")]
+    Source(#[source] Box<dyn std::error::Error + Send + Sync>),
 }
 
 impl OAuthHttpClientError {
-    /// Create an error from a transport-provided message.
+    /// Create an error from a message.
     pub fn new(message: impl Into<String>) -> Self {
         Self {
-            message: message.into(),
+            inner: OAuthHttpClientErrorKind::Message(message.into()),
+        }
+    }
+
+    /// Create an error from its underlying cause.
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -1984,13 +2000,10 @@ impl AuthorizationManager {
...
-            Ok(r) => r,
-            Err(e) => {
-                debug!("discovery request failed: {}", e);
-                return Ok(None);
-            }
-        };
+        let response = self
+            .discovery_get(discovery_url)
+            .await
+            .map_err(|error| Self::discovery_failed(discovery_url, error))?;
```
DaleSeo, line 2006: Previously, if one candidate failed, the code moved on to the next well-known URL. Now it returns `Err`, and the existing `?` in `try_discover_oauth_server`'s candidate loop propagates the error and skips all remaining candidates. Is this intended?

tjni: Yeah, I think so. What's happening is that now we immediately bubble up errors during the request itself or if the response returns a 5xx (the latter is converted to an error in the new `discovery_request` method). The other status codes still return the same response and fall into the next checks for whether that response is a 200, 403, or 404.

I feel this is a good compromise, where if requesting one of the candidates returns an unexpected error, we still surface that to the user, but something like 200, 403, or 404 we know how to handle.

DaleSeo:
Hi @tjni, I'm sorry it took me so long to get back to your PR.
We're preparing the v3 release today and have very little time left to land this, so I opened #1069 to carry the work forward. I preserved your original commit and authorship as the first commit, then added a follow-up commit that revises the public error API while we can still make breaking changes.

crates/rmcp/src/transport/auth.rs
DaleSeo: Makes sense. Thanks for the clarification.

tjni:
Thank you so much for taking care of the all the details involved here.

PR #1029, opened 2026-07-23
feat: route SEP-2260 associated server requests to the originating SSE stream

### LINKED ISSUE #873: Implement SEP-2260: Server requests must associate with a client request

gocamille:
Builds on #1027 to implement [SEP-2260](https://modelcontextprotocol.io/seps/2260-Require-Server-requests-to-be-associated-with-Client-requests.md)'s transport-layer requirement: server -> client requests associated with a client request are delivered on the originating request's SSE stream, not the standalone GET stream.

Tracking issue: #873

## Motivation and Context

#1027 enforces that `sampling/createMessage`, `elicitation/create`, and `roots/list` are only sent while handling a client request. On streamable HTTP, the transport still delivered all server-initiated requests on the standalone GET stream, so the association was not observable by clients. SEP-2260 requires associated requests to ride the originating request's stream.

## How

- Added `OriginatingRequestId`, a marker in the outbound request's non-serialized `Extensions`, attached at the send choke point (`Peer::send_request_with_option`) whenever `ORIGINATING_REQUEST` (from #1027) is in scope. SEP-2260 defines no wire field, so nothing changes on the wire.
- `LocalSessionWorker::resolve_outbound_channel` reads the marker and routes the request to the originating request's SSE channel via the existing `resource_router` with `close: false` so the stream stays open for the response.
- If the originating request has already completed, falls back to the standalone stream with a `tracing::warn!` rather than dropping the request.
- Documented on `SessionManager` that implementations serializing messages between processes lose the marker and need their own association mechanism; documented on the restricted request helpers that the association does not cross `tokio::spawn`.
- Client receive-side (per review): restricted server-to-client requests received while the client has no outbound request in flight are rejected with `-32602` (Invalid Params), gated on negotiated protocol `2026-07-28+`, `ping` exempt — the "Clients receiving server-to-client requests with no associated outbound request SHOULD respond with a -32602" line from SEP-2260. With a request in flight we cannot tell *which* one the server request belongs to (no wire field), so this is a deliberate under-approximation; exact stream-based enforcement is a follow-up.

## How Has This Been Tested?

- Unit tests for marker attachment in/out of handler scope (`service.rs`) and for the three routing outcomes (`local.rs`).
- End-to-end `test_sep_2260_stream_routing`: real streamable HTTP server; asserts an in-handler `elicitation/create` appears on the originating `tools/call` POST SSE stream and never on the standalone GET stream. Pins protocol `2025-11-25` since SEP-2567 serves `2026-07-28+` statelessly with no standalone GET stream to test against.
- Client receive-side tests drive a raw JSON-RPC server over a duplex pipe (a compliant rmcp server can't emit an unassociated request at `2026-07-28+`): unassociated `sampling/createMessage` rejected with `-32602`, permissive behavior preserved at `2025-11-25`, unassociated `ping` still answered.

## Breaking Changes

None. New public type `rmcp::service::OriginatingRequestId`; routing behavior changes only in the bundled `LocalSessionManager`. Custom `SessionManager` implementations are unaffected (documented limitation).

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

The marker is role-agnostic. The completed-originating-request fallback is deliberate as enforcement already happened at the service layer and dropping a committed in-flight request would hang the handler awaiting a reply, so we deliver on the standalone stream and `warn` instead.

One design note: the TypeScript SDK threads this association explicitly as `relatedRequestId` on transport send options. The equivalent here means a breaking change to `Transport`/`PeerSinkMessage` to carry data only this transport consumes, so this PR uses the existing `Extensions` side-channel instead. But if those traits ever take a breaking revision, promoting the association to an explicit field would be a follow-up (and it would also let serializing session managers comply with SEP-2260).

### DIFF
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -769,6 +769,17 @@ impl Peer<RoleServer> {
         }
     }
 
+    /// # SEP-2260: request association
+    ///
+    /// From protocol version `2026-07-28`, this request must be issued while
+    /// handling a client request — call it on the [`Peer`] from a
+    /// [`RequestContext`], from the handler's
+    /// own task. Calling it outside a handler returns an
+    /// `invalid_request` error. Note that task-local association does not
+    /// cross `tokio::spawn`: a task spawned inside a handler is treated as
+    /// outside the handler — use the task manager for long-running work.
+    /// Requests issued in-handler are delivered on the originating request's
+    /// SSE stream in the streamable HTTP transport.
     #[deprecated(
         since = "1.8.0",
         note = "Sampling is deprecated by SEP-2577 and will be removed in a future release. See https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577"
@@ -803,6 +814,17 @@ impl Peer<RoleServer> {
         }
     }
     method!(
+        /// # SEP-2260: request association
+        ///
+        /// From protocol version `2026-07-28`, this request must be issued while
+        /// handling a client request — call it on the [`Peer`] from a
+        /// [`RequestContext`], from the handler's
+        /// own task. Calling it outside a handler returns an
+        /// `invalid_request` error. Note that task-local association does not
+        /// cross `tokio::spawn`: a task spawned inside a handler is treated as
+        /// outside the handler — use the task manager for long-running work.
+        /// Requests issued in-handler are delivered on the originating request's
+        /// SSE stream in the streamable HTTP transport.
         #[deprecated(
             since = "1.8.0",
             note = "Roots is deprecated by SEP-2577 and will be removed in a future release. See https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577"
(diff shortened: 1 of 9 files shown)


crates/rmcp/src/service/server.rs
```diff
@@ -49,6 +49,31 @@ impl ServiceRole for RoleServer {
             _ => None,
         }
     }
+
+    fn enforce_request_association(
```
alexhancock, line 53: Is there a reason we don't also do this in `service/client.rs`? Trying to figure out where this part of the spec is implemented:

> Clients recieving server-to-client requests with no associated outbound request SHOULD respond with a -32602 (Invalid Params) error.

I think it may be missing?

alexhancock (commented):
Left a comment about a SHOULD in the spec, but looking good to me.

crates/rmcp/src/service/server.rs
gocamille: Yes @alexhancock it wasn't implemented anywhere, good catch. SEP-2260 defines no wire field, so the exact association is only seen at the transport layer and our streamable HTTP client merges the standalone GET stream and POST-response SSE streams into one flat stream before the service layer sees anything. On `stdio` there's no channel separation at all.

In f0b7fd3 if the client receives one of these requests with no outbound request in flight, it now responds with -32602. This only applies when the negotiated protocol is 2026-07-28+, and `ping` is exempt. The check lives in a new receive-side `ServiceRole::enforce_peer_request_association` hook, mirroring the send-side one from #1027.

For now the check is intentionally loose: with a request in flight we can't tell which outbound request the server request belongs to, so we accept it. Doing this properly (rejecting these requests when they arrive on the standalone GET stream, even while unrelated requests are in flight) needs the client transport to keep track of which stream each message came in on. If it's helpful I can file a follow-up so this PR stays scoped to routing plus the coarse check.

alexhancock: That sounds good to me for now, yes. Thanks for the updated commit.

alexhancock (approved):

alexhancock:
Very good. Thanks @gocamille !

PR #1038, opened 2026-07-23
fix!: omit resultType for legacy protocol sessions

### LINKED ISSUE #1036: rmcp 3.0 emits `resultType: "complete"` for legacy protocol sessions
## Summary

rmcp 3.0 unconditionally serializes `resultType: "complete"` on ordinary
response types, including `CallToolResult`, paginated list results, and
resource results.

That behavior is correct for MCP 2026-07-28, but it also changes the wire
format when a server is implementing or negotiating an older MCP version.
Consequently, upgrading rmcp can break existing legacy servers or clients even
when they do not enable the new protocol.

## Background

The discriminator was introduced intentionally by:

- [#915: Add MRTR model types (SEP-2322)](https://github.com/modelcontextprotocol/rust-sdk/pull/915).
- [#929: Add MRTR behavior support (SEP-2322)](https://github.com/modelcontextprotocol/rust-sdk/pull/929).

The SDK documentation says:

> Servers implementing this protocol version MUST include `resultType` in
> every result. For backward compatibility, clients MUST treat an absent field
> as `"complete"`.

However, backward compatibility here appears to cover deserialization by new
clients, not serialization by new servers that are still speaking a legacy
protocol.

## Reproduction

```rust
use rmcp::model::CallToolResult;

let result = CallToolResult::success(Vec::new());
let value = serde_json::to_value(result)?;

assert_eq!(
    value,
    serde_json::json!({
        "resultType": "complete",
        "content": [],
        "isError": false,
    })
);
```

This serialization is identical regardless of whether the peer negotiated MCP
2026-07-28 or an older version such as 2025-06-18.

For a legacy session, the historical wire representation was:

```json
{
  "content": [],
  "isError": false
}
```

The same issue also applies to other result types to which #915 added a
`result_type` field.

## Why this matters

Applications can upgrade rmcp globally while continuing to implement legacy MCP
server behavior. Adding a new field to their existing responses can break:

- Clients that strictly validate legacy response schemas.
- Clients with exact response-shape expectations.
- Applications promising no wire-format changes during an SDK-only upgrade.

We currently work around this by serializing the response to JSON and removing
`resultType` when its value is `"complete"` before sending responses from our
legacy MCP server.

## Existing upstream signal

The SDK already acknowledges this compatibility concern for empty results:

```rust
#[test]
fn empty_result_serializes_without_result_type() {
    assert_eq!(serde_json::to_value(EmptyResult {})?, serde_json::json!({}));
}
```

The accompanying comment explains that older or strict peers may reject
additional fields. Ordinary result types have the same issue when used in a
legacy protocol session.

## Expected behavior

Serialization should respect the negotiated protocol version:

- MCP 2026-07-28: include `resultType: "complete"`.
- Older MCP versions: preserve the legacy representation without `resultType`.
- Continue accepting absent `resultType` as `"complete"` when deserializing
  legacy responses.

A session-aware serializer, transport-level compatibility adapter, or explicit
legacy serialization API would all address this without weakening 2026-07-28
compliance.

## Question

Is unconditional serialization intentional even for legacy sessions, or should
rmcp provide a supported protocol-version-aware compatibility path?

DaleSeo:
Fixes #1036

## Motivation and Context

Stops emitting the SEP-2322 `resultType: "complete"` discriminator on responses to peers that negotiated a protocol version older than `2026-07-28`, restoring the historical wire shape for legacy sessions while keeping `2026-07-28` sessions spec-compliant.

## How Has This Been Tested?

Added integration tests

## Breaking Changes

Yes, the `result_type` field on the eight legacy-visible result types changed from `ResultType` to `Option<ResultType>`. It's okay to make a breaking change before v3 is released.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/handler/server.rs
+++ b/crates/rmcp/src/handler/server.rs
@@ -55,7 +55,8 @@ impl<H: ServerHandler> Service<RoleServer> for H {
     ) -> Result<<RoleServer as ServiceRole>::Resp, McpError> {
         // `context` is moved into the dispatch below, so read the negotiated version first.
         let protocol_version = context.protocol_version();
-        let mrtr_supported = protocol_version
+        // SEP-2322 (`resultType` discriminator, MRTR) exists from 2026-07-28.
+        let sep_2322_supported = protocol_version
             .as_ref()
             .is_some_and(|v| v.as_str() >= ProtocolVersion::V_2026_07_28.as_str());
         let requested_version = context.meta.protocol_version();
@@ -240,13 +241,18 @@ impl<H: ServerHandler> Service<RoleServer> for H {
                     .map(ServerResult::task_ack)
             }
         };
-        let result = result.and_then(|result| {
-            if matches!(result, ServerResult::InputRequiredResult(_)) && !mrtr_supported {
+        let result = result.and_then(|mut result| {
+            if matches!(result, ServerResult::InputRequiredResult(_)) && !sep_2322_supported {
                 Err(McpError::invalid_request(
                     "InputRequiredResult requires negotiated protocol version 2026-07-28 or newer",
                     None,
                 ))
             } else {
+                // Peers on protocol versions older than 2026-07-28 keep the
+                // legacy wire shape without `resultType: "complete"`.
+                if !sep_2322_supported {
+                    result.strip_result_type_for_legacy_peer();
+                }
                 Ok(result)
             }
         });
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -776,6 +776,13 @@ impl From<EmptyResult> for () {
 /// so unknown values are preserved rather than rejected. Servers implementing this
 /// protocol version MUST include `resultType` in every result. For backward
 /// compatibility, clients MUST treat an absent field as `"complete"`.
+///
+/// Ordinary results model the field as `Option<ResultType>`: `None` means the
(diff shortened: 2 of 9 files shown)


crates/rmcp/src/model.rs
```diff
@@ -3741,11 +3796,23 @@ impl<'de> Deserialize<'de> for CallToolResult {
     }
 }
 
+impl Default for CallToolResult {
+    fn default() -> Self {
+        CallToolResult {
+            result_type: Some(ResultType::COMPLETE),
```
DaleSeo, line 3802: With `Option`, a handler could manually set `result_type: None` and ship a 2026-07-28 response missing a required field. The MUST is now guaranteed by convention instead since constructors always produce `Some(COMPLETE)`.

crates/rmcp/src/handler/server.rs
```diff
@@ -240,13 +241,18 @@ impl<H: ServerHandler> Service<RoleServer> for H {
...
                 Err(McpError::invalid_request(
                     "InputRequiredResult requires negotiated protocol version 2026-07-28 or newer",
                     None,
                 ))
             } else {
+                // Peers on protocol versions older than 2026-07-28 keep the
+                // legacy wire shape without `resultType: "complete"`.
+                if !sep_2322_supported {
+                    result.strip_result_type_for_legacy_peer();
+                }
```
DaleSeo, line 255: `"resultType": "complete"` gets striped out here for legacy clients.

jamadeo (approved):

PR #1041, opened 2026-07-24
feat: expose oauth discovery metadata source

### LINKED ISSUE #1037: OAuth discovery should distinguish discovered authorization metadata from synthesized legacy endpoints
## Summary

`AuthorizationManager::discover_metadata()` currently returns a successful
`AuthorizationMetadata` result even when every discovery request fails and the
server provides no evidence that it supports OAuth.

In that situation, rmcp synthesizes `/authorize`, `/token`, and `/register`
endpoints from the MCP server's base URL. The resulting value is
indistinguishable from metadata actually discovered from an authorization
server.

Legacy endpoint fallback is useful for the
`auth/2025-03-26-oauth-endpoint-fallback` conformance scenario, but callers also
use `discover_metadata()` to determine whether a server supports OAuth at all.
Those callers can no longer distinguish an unauthenticated server from a legacy
OAuth server.

## Current behavior

After protected-resource metadata and authorization-server metadata discovery
both fail, rmcp does the following:

```rust
pub async fn discover_metadata(&self) -> Result<AuthorizationMetadata, AuthError> {
    if let Some(metadata) = self.discover_oauth_server_via_resource_metadata().await? {
        return Ok(metadata);
    }

    if let Some(metadata) = self.try_discover_oauth_server(&self.base_url).await? {
        return Ok(metadata);
    }

    Ok(Self::legacy_authorization_metadata(&self.base_url))
}
```

The synthesized result is equivalent to:

```rust
AuthorizationMetadata {
    authorization_endpoint: "https://example.com/authorize".into(),
    token_endpoint: "https://example.com/token".into(),
    registration_endpoint: Some("https://example.com/register".into()),
    ..Default::default()
}
```

The upstream test
`discover_metadata_falls_back_to_legacy_default_endpoints` explicitly asserts
that **five HTTP 404 responses** still produce `Ok(AuthorizationMetadata)`.

## Why this is a problem

An application that checks whether authentication is available or required
might reasonably write:

```rust
match authorization_manager.discover_metadata().await {
    Ok(metadata) => show_oauth_login(metadata),
    Err(AuthError::NoAuthorizationSupport) => connect_without_oauth(),
    Err(error) => report_discovery_error(error),
}
```

Before the legacy fallback was reintroduced, an MCP server with no OAuth
metadata reached `NoAuthorizationSupport`. It now reaches the OAuth-login path,
even though no authorization server, OAuth metadata, or functioning OAuth
endpoint was discovered.

Observable consequences include:

- Public MCP servers can incorrectly appear to require an interactive login.
- Clients can issue unnecessary authorization, token, or registration requests
  to guessed endpoints.
- Applications cannot distinguish verified discovery from an unverified
  compatibility fallback.
- Consumers must reverse-engineer rmcp's current synthesized metadata shape to
  recover the missing information. Such checks are brittle and can silently
  break when rmcp changes its fallback representation.

For example, a downstream workaround currently has to identify the exact
combination of `/authorize`, `/token`, `/register`, absent issuer, absent
optional metadata fields, and empty extension fields. This information should
be provided explicitly by the SDK instead.

## Expected behavior

Callers should be able to distinguish at least these outcomes:

1. Authorization metadata was actually discovered.
2. No authorization metadata was found, but legacy endpoints were synthesized
   as a compatibility fallback.
3. No authorization support was found.

The legacy fallback should remain available where required for older MCP OAuth
flows and conformance. The request is to expose its provenance or make its use
explicit, not necessarily to remove it.

### Option A: Expose discovery provenance

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum AuthorizationMetadataSource {
    ProtectedResourceMetadata,
    AuthorizationServerMetadata,
    LegacyEndpointFallback,
}

pub struct DiscoveredAuthorizationMetadata {
    pub metadata: AuthorizationMetadata,
    pub source: AuthorizationMetadataSource,
}

impl AuthorizationManager {
    pub async fn discover_metadata_with_source(
        &self,
    ) -> Result<DiscoveredAuthorizationMetadata, AuthError> {
        // ...
    }
}
```

Existing `discover_metadata()` behavior could remain unchanged for API and
conformance compatibility, while callers that need accurate capability
detection use the provenance-aware method.

### Option B: Add strict versus legacy-compatible discovery modes

```rust
pub enum OAuthDiscoveryPolicy {
    Strict,
    AllowLegacyEndpointFallback,
}

impl AuthorizationManager {
    pub async fn discover_metadata_with_policy(
        &self,
        policy: OAuthDiscoveryPolicy,
    ) -> Result<AuthorizationMetadata, AuthError> {
        // ...
    }
}
```

With `Strict`, absent metadata should return
`AuthError::NoAuthorizationSupport`; with `AllowLegacyEndpointFallback`, the
current behavior can be retained.

A dedicated `discover_metadata_strict()` method would also address this use
case.

### Option C: Only synthesize endpoints when OAuth is otherwise indicated

Another possibility is to require evidence of authorization support, such as a
relevant bearer challenge, or an explicitly requested OAuth login before
constructing legacy endpoints. This would avoid treating a server that returns
only 404 responses as OAuth-capable.

This option should be evaluated against the existing legacy conformance
scenario; explicit provenance or an opt-in policy may be more broadly useful.

## Regression history

The SDK previously removed guessed authorization metadata in
[PR #507](https://github.com/modelcontextprotocol/rust-sdk/pull/507):

```rust
// No valid authorization metadata found - return error instead of guessing
// OAuth endpoints must be discovered from the server, not constructed by the client
Err(AuthError::NoAuthorizationSupport)
```

The fallback was reintroduced in
[PR #960](https://github.com/modelcontextprotocol/rust-sdk/pull/960) to satisfy
client conformance, including
`auth/2025-03-26-oauth-endpoint-fallback`.

Preserving that compatibility makes sense, but silently collapsing real
discovery and guessed fallback into the same successful result creates a
regression for callers that use discovery to determine authentication support.

## Suggested regression coverage

- A server that returns only 404 responses is distinguishable from a server
  that publishes real authorization metadata.
- Strict/provenance-aware discovery identifies the all-404 case as unsupported
  or synthesized, rather than verified OAuth support.
- The existing `auth/2025-03-26-oauth-endpoint-fallback` conformance scenario
  continues to pass when legacy fallback is explicitly enabled or consumed.
- Existing callers of `discover_metadata()` retain compatibility if the new
  API is introduced additively.

DaleSeo:
Closes #1037

## Motivation and Context

The previous `AuthorizationManager::discover_metadata()` method fell back to the legacy default endpoints, `/authorize`, `/token`, and `/register`, based on the base URL when discovery returned nothing. This fallback is required for the `auth/2025-03-26-oauth-endpoint-fallback` conformance scenario. However, it made a successful fallback look the same as metadata that the server actually published. As a result, callers could not tell whether OAuth metadata was discovered or synthesized for compatibility. That behavior was originally introduced in #507 and changed by #960.

This replaces that method with `AuthorizationManager::resolve_metadata()`, which returns `AuthorizationMetadataResolution` containing both the `AuthorizationMetadata` and its `AuthorizationMetadataSource`. “Resolve” reflects that the method may either discover server-published metadata or synthesize legacy endpoints. The source distinguishes protected resource metadata, authorization server metadata, and the legacy endpoint fallback; `is_discovered()` reports whether the server published metadata.

Because v3 is still in beta, this makes provenance part of the primary metadata-resolution contract now instead of adding a parallel method that would need to be retained long-term. Internal SDK and conformance call sites explicitly extract `.metadata`; the discovery order, legacy endpoint synthesis, and conformance behavior remain unchanged.

## How Has This Been Tested?

Unit tests cover all three sources

## Breaking Changes

`AuthorizationManager::discover_metadata()` has been renamed to `resolve_metadata()` and now returns `Result<AuthorizationMetadataResolution, AuthError>` instead of `Result<AuthorizationMetadata, AuthError>`.

Callers must use the new method and extract the `metadata` field when they need `AuthorizationMetadata` directly:

```rust
// Before
let metadata = manager.discover_metadata().await?;
manager.set_metadata(metadata);

// After
let resolution = manager.resolve_metadata().await?;
manager.set_metadata(resolution.metadata);
```

Callers can inspect `resolution.source` or call `resolution.source.is_discovered()` before consuming the metadata.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -578,6 +578,38 @@ pub struct AuthorizationMetadata {
     pub additional_fields: HashMap<String, serde_json::Value>,
 }
 
+/// How [`AuthorizationMetadata`] was obtained during discovery.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum AuthorizationMetadataSource {
+    /// Discovered through RFC 9728 protected resource metadata.
+    ProtectedResourceMetadata,
+    /// Discovered through RFC 8414 / OpenID Connect metadata at the server's
+    /// base URL.
+    AuthorizationServerMetadata,
+    /// Nothing was discovered; the endpoints were synthesized from the base
+    /// URL (`/authorize`, `/token`, `/register`) for compatibility with the
+    /// 2025-03-26 MCP spec's default-endpoint fallback. The server gave no
+    /// evidence that it supports OAuth.
+    LegacyEndpointFallback,
+}
+
+impl AuthorizationMetadataSource {
+    /// Whether the metadata was actually published by the server, as opposed
+    /// to synthesized by the client as a legacy compatibility fallback.
+    pub fn is_discovered(self) -> bool {
+        !matches!(self, Self::LegacyEndpointFallback)
+    }
+}
+
+/// [`AuthorizationMetadata`] together with its discovery provenance.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct DiscoveredAuthorizationMetadata {
+    pub metadata: AuthorizationMetadata,
+    pub source: AuthorizationMetadataSource,
+}
+
 #[derive(Debug, Clone, Deserialize)]
 struct ResourceServerMetadata {
     resource: Option<String>,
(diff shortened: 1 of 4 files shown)


jamadeo (dismissed):

crates/rmcp/src/transport/auth.rs
```diff
@@ -578,6 +578,38 @@ pub struct AuthorizationMetadata {
...
+pub enum AuthorizationMetadataSource {
+    /// Discovered through RFC 9728 protected resource metadata.
+    ProtectedResourceMetadata,
+    /// Discovered through RFC 8414 / OpenID Connect metadata at the server's
+    /// base URL.
+    AuthorizationServerMetadata,
+    /// Nothing was discovered; the endpoints were synthesized from the base
+    /// URL (`/authorize`, `/token`, `/register`) for compatibility with the
+    /// 2025-03-26 MCP spec's default-endpoint fallback. The server gave no
+    /// evidence that it supports OAuth.
```
alexhancock, line 593: I would include something here if we're going to have a descriptive comment that metadata discovery is now mandatory in newer versions of the spec, and there is no endpoint-synthesis fallback any more.

https://modelcontextprotocol.io/specification/draft/basic/authorization/authorization-server-discovery#protected-resource-metadata-discovery-requirements

alexhancock (approved):

PR #368, opened 2025-08-12
fix(rmcp): return serialized json with structured content

ccamel55:
## Motivation and Context

Issue #355 seems to be about using this crate as a client, however the contrary also needs to be addressed.

The `2025-06-18` specification states the following recommendation:

>For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

This issue can be observed when trying to use tasks which return `structured_content` with clients that don't support `2025-06-18` yet. 

## How Has This Been Tested?

Unit tests check:
- both `content` and `structured_content` exist for structured results
- `content` contains serialized version of `structured_content`

## Breaking Changes

No breaking changes.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

- https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content
- https://github.com/modelcontextprotocol/rust-sdk/pull/357 PR for fixing issue previously mentioned

PR #1046, opened 2026-07-24
feat: add structured-only tool results that skip the text mirror

olaservo:
`CallToolResult::structured` and `structured_error` always mirror the structured value into `content` as serialized text, doubling large payloads. The spec marks that mirror as a [SHOULD](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#structured-content), not a MUST.

This PR makes the mirror optional. Opting out is explicit, since the mirror is the compatibility path for clients that do not read `structuredContent`:

- `CallToolResult::structured_only` / `structured_error_only` — constructors that leave `content` empty
- `StructuredOnly<T>` — return-type wrapper, a sibling of `Json<T>`, for `#[tool]` methods; the tool macro derives `outputSchema` from it the same way it does for `Json<T>`
- `CallToolResult::with_content` — replaces the `content` blocks, so a structured result can pair `structuredContent` with a custom rendering (e.g. a short text summary) instead of the verbatim mirror; the two fields need not carry the same data

### DIFF
--- a/crates/rmcp-macros/src/tool.rs
+++ b/crates/rmcp-macros/src/tool.rs
@@ -5,8 +5,8 @@ use syn::{Expr, Ident, ImplItemFn, LitStr, ReturnType, parse_quote};
 
 use crate::common::extract_doc_line;
 
-/// Check if a type is Json<T> or StructuredOnly<T> and extract the inner type T
-fn extract_json_inner_type(ty: &syn::Type) -> Option<&syn::Type> {
+/// Extract `T` from a `Json<T>` or `StructuredOnly<T>` return type.
+fn extract_structured_inner_type(ty: &syn::Type) -> Option<&syn::Type> {
     if let syn::Type::Path(type_path) = ty
         && let Some(last_segment) = type_path.path.segments.last()
         && (last_segment.ident == "Json" || last_segment.ident == "StructuredOnly")
@@ -18,19 +18,15 @@ fn extract_json_inner_type(ty: &syn::Type) -> Option<&syn::Type> {
     None
 }
 
-/// Extract schema expression from a function's return type
-/// Handles patterns like Json<T> and Result<Json<T>, E>, and the same
-/// shapes with StructuredOnly<T>
+/// Extract a schema expression from a structured function return type.
 fn extract_schema_from_return_type(ret_type: &syn::Type) -> Option<Expr> {
-    // First, try direct Json<T>
-    if let Some(inner_type) = extract_json_inner_type(ret_type) {
+    if let Some(inner_type) = extract_structured_inner_type(ret_type) {
         return syn::parse2::<Expr>(quote! {
             rmcp::handler::server::tool::schema_for_output::<#inner_type>()
         })
         .ok();
     }
 
-    // Then, try Result<Json<T>, E>
     let type_path = match ret_type {
         syn::Type::Path(path) => path,
         _ => return None,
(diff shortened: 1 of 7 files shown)


alexhancock:
Thanks for posting this @olaservo I agree if we want to make an opt-out path explicit and do it by default this looks like a good way to do it!

Given the spec says

> For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

I'd ask the question how often is this backwards compatibility relevant for callers out there in the ecosystem? How long has it been the case that tools could have relevant info in structured content.

Wondering when it would be reasonable to drop the automatic mirroring all together from an SDK perspective...

DaleSeo:
@alexhancock I was wondering the same thing. One thing we should consider here is that the mirror itself is only a month old. `structured` shipped without it for about a year until #927 added it, so I doubt many servers depend on it yet.

Given that, and with 3.0 GA targeted for 7/28, what do we think about flipping the default in this PR instead of adding an opt-out? `structured` and `Json<T>` would stop mirroring, and the mirror becomes the explicit opt-in. If we add `_only` variants now and flip the default later anyway, we end up shipping API that's legacy the day it stabilizes. FWIW, this is also where the TypeScript SDK already is. It never had automatic mirroring, and its docs just show tool authors including the text block themselves.

It would also make the PR smaller since the second wrapper and the macro change wouldn't be needed. Happy to help with the rework if that makes sense.

olaservo:
@alexhancock @DaleSeo good question, I surveyed this in a sample of popular open source clients and servers, and I discovered a wide range of interpretations and uses for one field vs the other.  I think its never been super clear when/if/how each field should be used, outside of how MCP Apps does it.

I think since the spec still has the 'mirroring' as a SHOULD for backwards compatibility, we should make it more flexible, but probably still make the automatic fallback the default?  In future spec versions, I think the plan is to design a way to explicitly choose one form or the other (especially in relation to the intended destination for the output, i.e. client-parsed only vs putting the output into model context).  At least this PR makes it closer to the other SDKs.

I'm also not sure how common it is for MCP servers to be written in Rust (vs clients), so in reality it may not come into play a lot in the wild?  I haven't done a survey for specific examples of popular open source Rust servers yet.

DaleSeo:
@olaservo @alexhancock I put together the alternative approach in #1085 so we can compare it concretely.

One correction to my earlier comment: I found that automatic mirroring was originally added in #368, not #927, so my claim that it was "only a month old" was incorrect.

Would appreciate your thoughts on the API and whether this default is worth considering.

DaleSeo:
@olaservo @alexhancock I rebased this branch onto the latest main and pushed a follow-up commit to align it with the current spec semantics. The update preserves the StructuredOnly opt-out for object-shaped outputs. It also keeps the required serialized TextContent fallback for arrays, primitives, and null, as specified in SEP-2106.

alexhancock:
Comparing to the TS SDK it seems like they don't auto-mirror for object-shaped data. Should we consider making one last change to make mirroring opt-in for this case vs the `StructuredOnly` approach?

cc @olaservo @DaleSeo

DaleSeo:
@alexhancock That's actually the question that led me to create PR #1085, so we could evaluate the opt-in alternative in a concrete way. I still think that model has a cleaner long-term shape, but @olaservo's client survey kind of changed my mind: https://gist.github.com/olaservo/3bb819673c444e4fe282c3af44a1ae01

Looking at the other Tier 1 SDKs, I'm not sure this is settled enough to justify either approach. So, this is tricky. 😅

PR #1050, opened 2026-07-24
fix: avoid orphaned sessions from oauth discovery post probe

### LINKED ISSUE #1048: OAuth discovery `initialize` POST probe can create server sessions that are never DELETEd
## Summary

During OAuth discovery, when a `GET` to the MCP endpoint returns `404` or `405`, `AuthorizationManager` falls back to POSTing a synthetic JSON-RPC `initialize` request to the endpoint in order to elicit a `401` response with a `WWW-Authenticate` header. If the server actually accepts that `initialize` instead of rejecting it (e.g. authorization is optional, or `initialize` is served before authentication), the server may allocate a session and return an `Mcp-Session-Id`. The client discards any non-`401` probe response without ever sending an HTTP `DELETE` for that session, so the session is orphaned on the server.

Separately, the MCP spec does not clearly indicate that clients should send `initialize` as part of OAuth discovery at all. The probe exists as a pragmatic workaround for streamable HTTP servers that reject a session-less `GET`, but it means discovery — which callers reasonably expect to be a read-only operation — can have server-side side effects.

## Current behavior

At `6839cfd`, resource metadata discovery works like this:

1. [`discover_resource_metadata_url()`](https://github.com/modelcontextprotocol/rust-sdk/blob/6839cfd61314c2a3c2bc51bf5563c42b1eee2fd2/crates/rmcp/src/transport/auth.rs#L2357-L2382) first issues a `GET` against the base URL via [`fetch_resource_metadata_url()`](https://github.com/modelcontextprotocol/rust-sdk/blob/6839cfd61314c2a3c2bc51bf5563c42b1eee2fd2/crates/rmcp/src/transport/auth.rs#L2386-L2412), hoping for a `401` + `WWW-Authenticate`.
2. Many streamable HTTP servers answer a session-less `GET` with `404`/`405`, in which case [`fetch_resource_metadata_url_with_post_probe()`](https://github.com/modelcontextprotocol/rust-sdk/blob/6839cfd61314c2a3c2bc51bf5563c42b1eee2fd2/crates/rmcp/src/transport/auth.rs#L2414-L2451) POSTs [`RESOURCE_METADATA_POST_PROBE_BODY`](https://github.com/modelcontextprotocol/rust-sdk/blob/6839cfd61314c2a3c2bc51bf5563c42b1eee2fd2/crates/rmcp/src/transport/auth.rs#L34-L39) — a synthetic `initialize` request with `protocolVersion: 2024-11-05`.
3. Any response other than `401` is [logged and discarded](https://github.com/modelcontextprotocol/rust-sdk/blob/6839cfd61314c2a3c2bc51bf5563c42b1eee2fd2/crates/rmcp/src/transport/auth.rs#L2440-L2446). If the server processed the `initialize` and returned an `Mcp-Session-Id`, no `DELETE` is ever sent for it.

The probe was introduced in #960 to pass the client conformance suite against servers that reject session-less `GET`s.

## Why this is a problem

- Servers where authorization is optional (or where `initialize` is allowed pre-auth) accumulate orphaned sessions: one per discovery attempt, never cleaned up.
- Callers that use discovery to check whether a server supports OAuth (the same usage pattern described in #1037) trigger this side effect on servers that don't require auth at all — exactly the servers that will process the probe.
- The spec describes discovery via `401` + `WWW-Authenticate` (RFC 9728) and well-known metadata paths; it does not sanction `initialize` as a discovery mechanism, so servers have no reason to expect or special-case this probe.

## Possible fixes

1. **Best-effort cleanup**: when the probe response is not `401` and carries an `Mcp-Session-Id` header, send an HTTP `DELETE` for that session before discarding the response.
2. **Make the probe a last resort**: try the well-known `oauth-protected-resource` paths *before* falling back to the POST probe (today the probe runs before the well-known lookup), so most servers never receive a synthetic `initialize`.
3. **Upstream spec clarification**: raise with the spec repo whether clients are intended to send `initialize` as part of OAuth discovery at all, and what server-side session semantics apply to unauthenticated probe requests.

Options 1 and 2 are complementary and could both be implemented in the SDK regardless of the spec discussion.

## Related

- #960 (introduced the POST probe for conformance)
- #1037 / #1041 (metadata provenance — orthogonal, but same discovery code path)

DaleSeo:
Fixes #1048.

## Motivation and Context

When a GET request to the MCP endpoint returns a 404 or 405, which is expected for servers without SSE support, AuthorizationManager falls back to sending a synthetic JSON-RPC initialize request. This is meant to trigger a 401 response with a WWW-Authenticate header.

However, on servers where authorization is optional, or where initialize is available before authentication, the request is accepted. A session is created, and an Mcp-Session-Id is returned. The client then discards the session without sending a DELETE request, leaving one orphaned session after every discovery attempt.

This PR fixes the issue by sending an HTTP DELETE request for the session whenever a probe response includes an Mcp-Session-Id header. The client does this before processing the response. Any cleanup failures are logged at the debug level and never cause discovery to fail.

## How Has This Been Tested?

Add tests

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -26,7 +26,7 @@ use thiserror::Error;
 use tokio::sync::{Mutex, RwLock};
 use tracing::{debug, warn};
 
-use crate::transport::common::http_header::HEADER_MCP_PROTOCOL_VERSION;
+use crate::transport::common::http_header::{HEADER_MCP_PROTOCOL_VERSION, HEADER_SESSION_ID};
 
 const DEFAULT_HTTP_TIMEOUT: Duration = Duration::from_secs(30);
 const MAX_OAUTH_HTTP_RESPONSE_BODY_BYTES: usize = 1024 * 1024;
@@ -1000,6 +1000,19 @@ pub struct AuthorizationManager {
     resource_scopes: RwLock<Vec<String>>,
     /// OIDC Dynamic Client Registration `application_type` (SEP-837)
     application_type: Option<String>,
+}
+
+/// Outcome of a resource metadata discovery GET issued by
+/// [`AuthorizationManager::probe_resource_metadata_url`].
+enum ResourceMetadataProbeOutcome {
+    /// The probe located a resource metadata URL: either a 200 on the probed
+    /// URL itself, or a 401 whose WWW-Authenticate header pointed at one.
+    Found(Url),
+    /// The probe got a 404/405, typical of streamable HTTP servers that
+    /// reject session-less GETs, so the POST probe fallback is worth trying.
+    PostProbeEligible,
+    /// The probe yielded nothing usable.
+    Unavailable,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


DaleSeo:
Close this in favor of #1052

PR #1051, opened 2026-07-24
fix: default to allowing missing `issuer`

### LINKED ISSUE #1047: OAuth discovery rejects metadata without `issuer`, breaking compatibility with rmcp 2.2
### Description

rmcp 3.0.0-beta.1 rejects OAuth authorization-server metadata that omits `issuer`, although the same metadata works with rmcp 2.2.0.

For example:

```json
{
  "authorization_endpoint": "https://auth.example.com/authorize",
  "token_endpoint": "https://auth.example.com/token"
}
```

When returned from an OAuth authorization-server metadata endpoint, this document causes `AuthorizationManager::discover_metadata()` to fail with `AuthorizationServerMissingIssuer`.

This appears to be an unintended regression introduced by #996. Its description explicitly states that issuer validation should reject mismatches while “preserving compatibility for metadata that omits issuer,” but the current implementation rejects missing issuers unconditionally.

Although omitting `issuer` is not compliant with modern OAuth metadata requirements, existing MCP servers use this shape and are supported by the current stable release.

**Expected behavior:** Metadata with an explicitly provided but incorrect issuer should be rejected, while legacy metadata omitting `issuer` should continue to work as it does in rmcp 2.2.0.

**Actual behavior:** Both incorrect and missing issuers are rejected.

**Versions:**

- rmcp 2.2.0: accepted.
- rmcp 3.0.0-beta.1: rejected.

jamadeo:
fixes #1047

## Motivation and Context
Strictly compliant servers must send `issuer`, but to preserver legacy compatibility, default to allowing missing.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -963,6 +963,7 @@ pub struct AuthorizationManager {
     resource_scopes: RwLock<Vec<String>>,
     /// OIDC Dynamic Client Registration `application_type` (SEP-837)
     application_type: Option<String>,
+    strict_issuer_validation: bool,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -1208,6 +1209,7 @@ impl AuthorizationManager {
             www_auth_scopes: RwLock::new(Vec::new()),
             resource_scopes: RwLock::new(Vec::new()),
             application_type: Some(DEFAULT_APPLICATION_TYPE.to_string()),
+            strict_issuer_validation: false,
         };
 
         Ok(manager)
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -1255,6 +1220,17 @@ impl AuthorizationManager {
...
     }
 
+    /// Configure whether authorization server metadata discovery requires the
+    /// metadata `issuer` field.
+    ///
+    /// The default is `false` to preserve compatibility with legacy
+    /// authorization servers that omit `issuer`. Set this to `true` to enforce
+    /// the RFC 8414/OIDC requirement that discovered metadata include `issuer`
+    /// whenever the expected issuer can be derived from the discovery URL.
+    pub fn set_strict_issuer_validation(&mut self, strict: bool) {
```
alexhancock, line 1230: Should we provide a public setter, or just announce the change and change the default at some point in the future? Providing a setter suggests to people they could set it to false and leave it that way when my understanding of this is that we'll want to eventually always require it.

alexhancock (approved):

crates/rmcp/src/transport/auth.rs
jamadeo: I think once we are at the point where we remove non-compliant server support, we'd remove the setter completely and that forces users to change their code anyway.

Issue #1033, opened 2026-07-23
SEP-2260 follow-up: stream-based enforcement of client receive-side request association


gocamille:
Follow-up to #873 / PR #1029.

PR #1029 implements the SEP-2260 receive-side SHOULD ("Clients receiving server-to-client requests with no associated outbound request SHOULD respond with a `-32602` (Invalid Params) error") via a `ServiceRole::enforce_peer_request_association` hook: the `RoleClient` implementation rejects restricted requests (sampling/roots/elicitation; `ping` exempt) with `-32602` when the client has no outbound request in flight, gated on negotiated protocol ≥ 2026-07-28.

What it intentionally leaves out is checking which request a server request belongs to. SEP-2260 defines no wire field for this and you can only tell at the transport layer, by which HTTP response stream a message arrived on. The streamable HTTP client merges the standalone GET stream and POST-response SSE streams into one flat stream before the service layer sees anything, so today:

- A restricted request arriving on the standalone GET stream while an unrelated outbound request is in flight is incorrectly accepted
- We can't verify a request arrived on the SSE stream of the specific originating POST

Work: have the streamable HTTP client transport track which stream each inbound message arrived on (standalone GET vs. a specific POST's SSE stream), so the receive-side check can reject restricted requests that didn't come in on the right stream. Stdio has no stream separation, so the coarse in-flight check remains the best we can do there.

This is not required for 2026-07-28 conformance (the receive-side check is a SHOULD, and the coarse check in #1029 covers the conformance scenarios), so this is not on the 2026-07-28 milestone.

alexhancock:
Thanks for opening to track @gocamille! Do you want to tackle the change for it? Not urgent I don't think

gocamille:
Hi @alexhancock I'm happy to work on this! I'll be putting out a PR for this shortly for review.

gocamille:
Landed in #1055 (1d0473b). The streamable HTTP client now marks inbound requests with the stream they arrived on, and `RoleClient::enforce_peer_request_association` rejects restricted requests that arrive unassociated — so a request on the standalone GET stream is refused with `-32602` even while unrelated requests are in flight.

The remaining piece is tracked in #1108: at protocol `2026-07-28` the client shouldn't keep a session id or open the GET stream at all, which makes the unassociated case unreachable rather than merely rejected.

Closing as completed.

PR #1055, opened 2026-07-27
feat: SEP-2260 stream-based enforcement of client receive-side request association

gocamille:
Implements the stream-based enforcement described in [SEP-2260](https://modelcontextprotocol.io/seps/2260-Require-Server-requests-to-be-associated-with-Client-requests.md).

Tracking issue: #1033 (follow-up to #873 / #1029)

## Motivation and Context

PR #1029 implemented the receive-side SHOULD ("Clients receiving server-to-client requests with no associated outbound request SHOULD respond with a `-32602` error") with a coarse check: reject restricted requests (`sampling/createMessage`, `elicitation/create`, `roots/list`; `ping` exempt) only when the client has no outbound request in flight. It couldn't tell which request a server request belongs to. SEP-2260 defines no wire field, so association is only observable at the transport layer via which HTTP response stream a message arrived on.

This PR closes that gap for streamable HTTP. Now, a restricted request arriving on the GET stream (or on the SSE stream of a POST whose request is no longer in flight) is rejected with`-32602` even when unrelated requests are in flight.

## How

- The streamable HTTP client transport attaches an `InboundStreamOrigin` marker (`Unassociated` | `OutboundRequest(RequestId)`) to each inbound request's non-serialized `Extensions`, recording whether it arrived on the standalone GET stream or a specific POST's SSE response stream. The marker survives SSE reconnection/resumption.
- The service layer translates the marker plus the in-flight responder pool into a `PeerRequestAssociation` enum (`Associated` | `Unassociated` | `Unknown { has_pending_outbound_request }`) passed to `ServiceRole::enforce_peer_request_association`.
- `RoleClient::enforce_peer_request_association` rejects `Unassociated` restricted requests with `-32602`; `Associated` is accepted; `Unknown` falls back to the coarse in-flight check from #1029.
- Still gated on negotiated protocol ≥ `2026-07-28`; older peers keep legacy behavior.
- Transports without stream separation (stdio, in-process) attach no marker and yield `Unknown`.

## How Has This Been Tested?

- Unit tests for the `InboundStreamOrigin` → `PeerRequestAssociation` mapping and for `enforce_peer_request_association` across all three association states.
- Transport tests verifying `execute_sse_stream` marks inbound requests with their stream origin (responses are untouched) and that the origin marker survives SSE resumption/reconnect.
- End-to-end test over streamable HTTP (`test_sep_2260_stream_enforcement.rs`): a restricted request on the originating POST's SSE stream reaches the handler; the same request on the standalone GET stream is rejected with `-32602` while an unrelated request is in flight. The scripted server negotiates `2026-07-28` and creates a session id. It's deliberately non-conforming (SEP-2567 removes sessions and the GET endpoint at that version), since receive-side enforcement exists to defend against exactly such servers and rmcp's client tolerates the session id and opens the GET stream.

## Breaking Changes

None. New public API: `PeerRequestAssociation` and the `ServiceRole::enforce_peer_request_association` hook (with a permissive default). Behavior changes only for streamable HTTP clients negotiating `2026-07-28+` against servers that send restricted requests on the wrong stream.

## Types of changes

- [ ] Documentation update

## Checklist

  ## Additional context

Possible follow-up: skip session tracking / the standalone GET stream entirely when the negotiated version is ≥ `2026-07-28`, making the unassociated-GET-stream state unreachable.

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -180,7 +180,7 @@ pub trait ServiceRole: std::fmt::Debug + Send + Sync + 'static + Copy + Clone {
 /// SEP-2260 defines no wire field for association, so only stream-separating
 /// transports (streamable HTTP) can observe it; other transports yield
 /// [`Self::Unknown`].
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[expect(clippy::exhaustive_enums, reason = "intentionally exhaustive")]
 pub enum PeerRequestAssociation {
     /// Arrived on the response stream of an in-flight outbound request.
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -163,8 +163,7 @@ pub trait ServiceRole: std::fmt::Debug + Send + Sync + 'static + Copy + Clone {
     /// SEP-2260 says clients receiving a server-to-client request with no
     /// associated outbound request should reject it with invalid params. An
     /// error return is sent back to the peer instead of dispatching to the
-    /// handler. The `association` argument is the transport-observed stream
-    /// association; see [`PeerRequestAssociation`].
+    /// handler.
     #[doc(hidden)]
     fn enforce_peer_request_association(
         _peer_request: &Self::PeerReq,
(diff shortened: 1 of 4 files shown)


alexhancock:
Nice. I should be able to review tomorrow if @DaleSeo @jamadeo don't get to it first!

crates/rmcp/src/service.rs
```diff
@@ -168,12 +168,31 @@ pub trait ServiceRole: std::fmt::Debug + Send + Sync + 'static + Copy + Clone {
...
     }
 }
 
+/// How an inbound peer request relates to this side's in-flight outbound
+/// requests (SEP-2260).
+///
+/// SEP-2260 defines no wire field for association, so only stream-separating
+/// transports (streamable HTTP) can observe it; other transports yield
+/// [`Self::Unknown`].
+#[derive(Debug, Clone, PartialEq, Eq)]
```
DaleSeo: nit: Deriving `Copy` matches how we handle other small enums, and it lets callers reuse the state without having to clone it explicitly.

```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```

DaleSeo (dismissed):
Thanks for implementing this, @gocamille! I only have a nitpick.

crates/rmcp/src/service.rs
gocamille: Oh, good catch @DaleSeo ! Thank you!

gocamille:
Looks like adding suggestions requires another review, so re-requesting -- thank you again @DaleSeo !

DaleSeo (approved):

PR #1065, opened 2026-07-27
fix: remove server_info from DiscoverResult

### LINKED ISSUE #1064: DiscoverResult contains invalid server_info field
Pulling this out from a comment

> Is this right? looking at the latest spec https://modelcontextprotocol.io/specification/draft/schema#discoverresult this is not even a field in the type anymore after a late spec change https://github.com/modelcontextprotocol/modelcontextprotocol/pull/3002.
> 
>  

 _Originally posted by @howardjohn in [#1044](https://github.com/modelcontextprotocol/rust-sdk/pull/1044/changes#r3661276541)_

howardjohn:
## Motivation and Context
Fixes https://github.com/modelcontextprotocol/rust-sdk/issues/1064

Replaces https://github.com/modelcontextprotocol/rust-sdk/pull/1044

This removes the `server_info` field which is not part of the spec:
 https://modelcontextprotocol.io/specification/draft/schema#discoverresult.

1044 add a synthetic server_info field; this PR removes it. This aligns with other types in this repo (using _meta as the source of truth, not a projection) and with other SDKS (Go and TypeScript do not have a `server_info`) field. Instead, users can use the helper accessors.

## How Has This Been Tested?
Tested proxying to copilot MCP

## Breaking Changes
Breaking from beta3 - the server_info field is removed

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -25,7 +25,7 @@ use crate::{
         NumberOrString, PaginatedRequestParams, ProgressNotification, ProgressNotificationParam,
         ProtocolVersion, ReadResourceRequest, ReadResourceRequestParams, ReadResourceResponse,
         ReadResourceResult, Reference, RequestId, RequestMetaObject, RootsListChangedNotification,
-        ServerInfo, ServerJsonRpcMessage, ServerNotification, ServerRequest, ServerResult,
+        ServerJsonRpcMessage, ServerNotification, ServerPeerInfo, ServerRequest, ServerResult,
         SetLevelRequest, SetLevelRequestParams, SubscribeRequest, SubscribeRequestParams,
         SubscriptionFilter, SubscriptionsListenRequest, SubscriptionsListenRequestParams,
         SubscriptionsListenResult, UnsubscribeRequest, UnsubscribeRequestParams, UpdateTaskParams,
@@ -213,7 +213,7 @@ impl ServiceRole for RoleClient {
     type PeerResp = ServerResult;
     type PeerNot = ServerNotification;
     type Info = ClientInfo;
-    type PeerInfo = ServerInfo;
+    type PeerInfo = ServerPeerInfo;
     type InitializeError = ClientInitializeError;
     const IS_CLIENT: bool = true;
 
@@ -776,7 +776,7 @@ where
     let ServerResult::InitializeResult(initialize_result) = response else {
         return Err(ClientInitializeError::ExpectedInitResult(Some(response)));
     };
-    peer.set_peer_info(initialize_result);
+    peer.set_peer_info(initialize_result.into());
 
     // send notification
     let notification = ClientJsonRpcMessage::notification(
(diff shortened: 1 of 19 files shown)


crates/rmcp/tests/test_server_discover.rs
```diff
@@ -121,25 +122,43 @@ fn discover_result_accepts_server_info_in_namespaced_metadata() {
...
     );
+    assert_eq!(
+        serialized["_meta"]["unrelated"],
+        json!({ "preserved": true })
+    );
 }
 
 #[test]
-fn discover_result_prefers_top_level_server_info_over_namespaced_metadata() {
+fn discover_result_ignores_legacy_top_level_server_info() {
```
tsarlandie-oai, line 160: We should not change this test. I think it's a useful test to make sure we remain compatible with legacy servers.

tsarlandie-oai (changes requested):
Moving `serverInfo` into `_meta` for serialization makes sense, but clients must remain able to consume legacy top-level `serverInfo`, and must not fail when server identity is absent: the specification makes this metadata optional. The current `MissingServerInfo` path already causes 15 client conformance failures and bypasses the conservative fallback in #1045. We should preserve the legacy deserialization test, emit only the modern representation, and accept both representations on input.

DaleSeo:
@howardjohn Can you take a look at the failing checks?

howardjohn:
> @howardjohn Can you take a look at the failing checks?

sorry should be good to go now!

crates/rmcp/src/service/client.rs
```diff
@@ -846,13 +846,15 @@ where
...
-                });
+                if let Some(server_info) = result.server_info() {
+                    peer.set_peer_info(ServerInfo {
+                        protocol_version: selected.clone(),
+                        capabilities: result.capabilities,
+                        server_info,
+                        instructions: result.instructions,
+                        meta: result.meta,
+                    });
+                }
```
DaleSeo: Since `serverInfo` is optional, this branch leaves `peer_info` unset even for a valid discovery response. That means the version, capabilities, and instructions were negotiated, but `RoleClient::enforce_peer_request_association` still treats the connection as legacy and skips the 2026-07-28 guard against unsolicited sampling, roots, or elicitation requests.

DaleSeo:
For context, I introduced the top-level field in #973 following the accepted SEP-2575 schema. About 15 hours before it merged, modelcontextprotocol/modelcontextprotocol/pull/3002 moved server identity into the optional result `_meta` field. 😅
I opened modelcontextprotocol/modelcontextprotocol/pull/3159 to make that post-Final change clear to future implementers.

crates/rmcp/src/service/client.rs
howardjohn: Good call.. working on it

howardjohn: I think we have 3 options:
1. Make PeerInfo a new type with server_info optional (breaking type change)
2. Make an Implementation with just version, a empty/fake `name`, and the rest `None`. No breaking change but pretty wonky
3. Just pass the version through for enforce_peer_request_association but do not give any peer info (they lose instructions etc)

(1) seems the best option but wanted to check we are ok with the break?

DaleSeo: @howardjohn I agree that 1 is the cleanest option, and I'm okay with the break since this is already a breaking change.

howardjohn: :+1: sent a change with (1)

alexhancock (approved):

PR #1068, opened 2026-07-28
fix: gate client handler bounds for local

### LINKED ISSUE #1062: `ClientHandler`'s `Send + Sync` bound is not gated by the `local` feature (unlike `ServerHandler` and `Service`)
`ClientHandler` appears to have been missed when the `local` feature was introduced: it is the only handler/service trait whose `Send + Sync` bound is not gated on it.

`ServerHandler` has both variants (`src/handler/server.rs:431-438`):

```rust
#[cfg(not(feature = "local"))]
pub trait ServerHandler: Sized + Send + Sync + 'static { ... }

#[cfg(feature = "local")]
pub trait ServerHandler: Sized + 'static { ... }
```

`Service<R>` — the trait `ClientHandler` blankets into via `impl<H: ClientHandler> Service<RoleClient> for H` — has both too (`src/service.rs:136` and `:151`).

`ClientHandler` has only one (`src/handler/client.rs:89`):

```rust
pub trait ClientHandler: Sized + Send + Sync + 'static {
```

So with `local` enabled, a client handler still has to be `Send + Sync`, even though everything downstream of it no longer requires that.

### Why it matters

This is the last blocker for a browser-wasm MCP **client** in our case. With `features = ["client", "local"]` on `wasm32-unknown-unknown`:

- `rmcp` itself compiles cleanly for the target.
- `MaybeSendFuture` correctly stops requiring `Send` on the handler's futures.
- The build then fails only on this bound, because our handler owns a tool registry holding `Arc<dyn Tool>` values that are deliberately not `Send`/`Sync` on wasm — browser tools capture `JsValue`, which is `!Send` by design.

That is the same `!Send` JS-interop constraint described in #379. Worth noting relative to that thread: the conclusion there was that wasm support would be a very large changeset, but for the client path specifically the `local` feature already covers essentially all of it — this one bound is what remains. (#379 is currently marked stale and scheduled to auto-close, which is part of why I am filing this separately: it is a small, self-contained item rather than the umbrella effort.)

### Reproduction

```rust
// Cargo.toml
// [target.'cfg(target_arch = "wasm32")'.dependencies]
// rmcp = { version = "2.2", features = ["client", "local"] }

use std::rc::Rc;

struct Handler {
    // Any !Send state — a JsValue, an Rc, a browser tool registry.
    _local: Rc<()>,
}

impl rmcp::handler::client::ClientHandler for Handler {
    fn get_info(&self) -> rmcp::model::ClientInfo {
        Default::default()
    }
}
```

```
$ cargo check --target wasm32-unknown-unknown
error[E0277]: `Rc<()>` cannot be shared between threads safely
  --> src/lib.rs:7:47
note: required by a bound in `ClientHandler`
  --> rmcp-2.2.0/src/handler/client.rs:89:41

error[E0277]: `Rc<()>` cannot be sent between threads safely
  --> src/lib.rs:7:47
note: required by a bound in `ClientHandler`
  --> rmcp-2.2.0/src/handler/client.rs:89:34
```

(Verified against rmcp 2.2.0 on rustc 1.94.0.)

### Suggested fix

Mirror `ServerHandler`:

```rust
#[cfg(not(feature = "local"))]
pub trait ClientHandler: Sized + Send + Sync + 'static { ... }

#[cfg(feature = "local")]
pub trait ClientHandler: Sized + 'static { ... }
```

`MaybeSend` (`src/service.rs:18`) already expands to `Send + Sync` without `local` and to nothing with it, so `Sized + MaybeSend + 'static` would work as a single definition if you prefer avoiding the duplicated method bodies — `ServerHandler` uses a `server_handler_methods!()` macro for that reason, and `ClientHandler`'s body is small enough to go either way.

No behaviour change without `local`; the bound is identical there.

### Scope

I have only verified that this unblocks *compiling* a client handler. I have not yet checked whether the streamable-HTTP transports build for `wasm32-unknown-unknown`, so there may be further work behind this for an end-to-end browser client. This bound is simply the first thing in the way, and it looks like an oversight rather than a decision.

DaleSeo:
Fixes #1062.

## Motivation and Context

This gives `ClientHandler` feature-gated trait bounds, allowing handlers with single-threaded state when the `local` feature is enabled while retaining the existing `Send + Sync` signature otherwise. It also adds a compile-time regression test using `Rc`.

## How Has This Been Tested?
Added a regression test

## Breaking Changes
None. Builds without `local` retain the existing `Send + Sync` requirements.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -239,6 +239,11 @@ required-features = [
 path = "tests/test_with_js.rs"
 
 [[test]]
+name = "test_local_client_handler"
+required-features = ["client", "local"]
+path = "tests/test_local_client_handler.rs"
+
+[[test]]
 name = "test_notification"
 required-features = ["server", "client"]
 path = "tests/test_notification.rs"
--- a/crates/rmcp/src/handler/client.rs
+++ b/crates/rmcp/src/handler/client.rs
@@ -85,189 +85,202 @@ impl<H: ClientHandler> Service<RoleClient> for H {
     }
 }
 
+macro_rules! client_handler_methods {
+    () => {
+        fn ping(
+            &self,
+            context: RequestContext<RoleClient>,
+        ) -> impl Future<Output = Result<(), McpError>> + MaybeSendFuture + '_ {
+            std::future::ready(Ok(()))
+        }
+
+        fn create_message(
+            &self,
+            params: CreateMessageRequestParams,
+            context: RequestContext<RoleClient>,
+        ) -> impl Future<Output = Result<CreateMessageResult, McpError>> + MaybeSendFuture + '_ {
+            std::future::ready(Err(
+                McpError::method_not_found::<CreateMessageRequestMethod>(),
+            ))
+        }
+
+        fn list_roots(
+            &self,
+            context: RequestContext<RoleClient>,
+        ) -> impl Future<Output = Result<ListRootsResult, McpError>> + MaybeSendFuture + '_ {
+            std::future::ready(Ok(ListRootsResult::default()))
+        }
+
+        /// Handle an elicitation request from a server asking for user input.
+        ///
+        /// This method is called when a server needs interactive input from the user
+        /// during tool execution. Implementations should present the message to the user,
+        /// collect their input according to the requested schema, and return the result.
+        ///
+        /// # Arguments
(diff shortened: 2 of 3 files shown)


alexhancock (approved):

gold-silver-copper:
awesome thx for the quick fix

PR #1069, opened 2026-07-28
fix!: preserve OAuth discovery transport errors

DaleSeo:
## Motivation and Context

OAuth metadata discovery previously converted transport failures into a successful fallback path, hiding actionable causes such as connection and certificate errors. This PR reports those failures with the discovery URL and complete source chain while retaining fallback behavior for valid non-matching discovery responses.

PR #1028 identified this through a concrete failure: using the Linear MCP server from a `node:22-bookworm-slim` container without CA certificates made the server appear not to support OAuth, while the underlying TLS error was discarded. This PR retains the original intended semantics: transport failures and HTTP 5xx responses stop discovery and surface the cause, while recognized non-success responses can still fall through to other discovery candidates.

Because this work already changes error propagation and v3 is approaching, this is the right point to settle the public API before it becomes stable. Rather than expand the public `OAuthHttpClientError` wrapper with separate constructors for messages and source errors, it becomes a domain-specific alias for `Box<dyn Error + Send + Sync>`. This matches existing SDK error aliases, lets custom clients propagate native errors with `?`, and confines the concrete compatibility wrapper required by `oauth2` to the private implementation.

## How Has This Been Tested?

Added coverage

## Breaking Changes

`OAuthHttpClientError` changes from a public struct with an `OAuthHttpClientError::new` constructor to `Box<dyn Error + Send + Sync>`. Custom `OAuthHttpClient` implementations should return their native errors with `.into()` or box them directly.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -70,36 +70,19 @@ impl OAuthHttpRequest {
     }
 }
 
-/// Error returned by a custom OAuth HTTP client.
-#[derive(Debug, Error)]
-#[error(transparent)]
-pub struct OAuthHttpClientError {
-    inner: OAuthHttpClientErrorKind,
-}
+/// Type-erased error returned by an [`OAuthHttpClient`].
+pub type OAuthHttpClientError = Box<dyn std::error::Error + Send + Sync>;
 
 #[derive(Debug, Error)]
-enum OAuthHttpClientErrorKind {
-    #[error("{0}")]
-    Message(String),
-
-    #[error("{0}")]
-    Source(#[source] Box<dyn std::error::Error + Send + Sync>),
-}
-
-impl OAuthHttpClientError {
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -2525,37 +2541,33 @@ impl AuthorizationManager {
             discovery_url.set_fragment(None);
             discovery_url.set_path(&candidate_path);
             if let Some(resource_metadata_url) =
```
alexhancock, line 2543: Do we need to handle the error case here?

DaleSeo: The error case is intentionally propagated by `?`. I had the same question in https://github.com/modelcontextprotocol/rust-sdk/pull/1028#discussion_r3635138981.

alexhancock (approved):

PR #1071, opened 2026-07-28
fix: preserve transient OAuth discovery HTTP errors

tsarlandie-oai:
## Summary

Follow-up to #1069.

OAuth metadata discovery now preserves transport failures and HTTP 5xx responses, but HTTP `408 Request Timeout`, `425 Too Early`, and `429 Too Many Requests` still fall through as though a metadata endpoint were absent. As a result, a temporary timeout, early-data rejection, or rate limit can cause discovery to continue to another endpoint or report a legacy fallback instead of surfacing the operational failure.

This change classifies HTTP 408, 425, and 429 alongside 5xx in the existing `discovery_request` helper. The resulting metadata error preserves both the actual response status and the failing discovery URL.

## Compatibility

- Preserve HTTP 401 challenge handling.
- Preserve HTTP 404 and 405 discovery fallback, with separate regression cases for both.
- Preserve the existing HTTP 5xx behavior.
- Do not introduce automatic retries, new dependencies, or public API changes.

## Regression coverage

Add nine parameterized cases covering HTTP 408, 425, and 429 at each of:

1. The protected resource endpoint.
2. The protected resource metadata endpoint.
3. The authorization server metadata endpoint.

Each case verifies that discovery reports the correct failing URL and status and does not issue additional discovery requests. The regression was first reproduced against upstream `main` before applying the production fix.

## Validation

- `cargo +1.96 test -p rmcp --lib --all-features` (482 tests passing).
- `cargo +1.96 clippy --all-targets --all-features -- -D warnings`.
- `cargo +nightly-2025-09-18 fmt --all -- --check`.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -2635,10 +2635,14 @@ impl AuthorizationManager {
         request: OAuthHttpRequest,
     ) -> Result<HttpResponse, OAuthHttpClientError> {
         let response = self.http_client.execute(request).await?;
-        if response.status().is_server_error() {
-            return Err(Box::new(OAuthHttpError::UnexpectedStatus(
-                response.status(),
-            )));
+        let status = response.status();
+        if status.is_server_error()
+            || matches!(
+                status,
+                StatusCode::REQUEST_TIMEOUT | StatusCode::TOO_MANY_REQUESTS
+            )
+        {
+            return Err(Box::new(OAuthHttpError::UnexpectedStatus(status)));
         }
         Ok(response)
     }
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -2635,10 +2635,14 @@ impl AuthorizationManager {
...
         let response = self.http_client.execute(request).await?;
-        if response.status().is_server_error() {
-            return Err(Box::new(OAuthHttpError::UnexpectedStatus(
-                response.status(),
-            )));
+        let status = response.status();
+        if status.is_server_error()
+            || matches!(
+                status,
+                StatusCode::REQUEST_TIMEOUT | StatusCode::TOO_MANY_REQUESTS
```
DaleSeo: Would `425` would make the transient classification more complete?

DaleSeo (dismissed):
Thanks for tightening the discovery error boundary!

crates/rmcp/src/transport/auth.rs
tsarlandie-oai: added

DaleSeo (approved):

PR #1072, opened 2026-07-28
fix(auth): propagate retryable OAuth discovery errors

tsarlandie-oai:
## Motivation and Context

Follow-up to #1069.

OAuth metadata discovery now correctly propagates transport failures and HTTP
5xx responses, but HTTP 408 Request Timeout and 429 Too Many Requests still pass
through as ordinary responses. The protected-resource and authorization-server
metadata probes then interpret those temporary failures as absent metadata. If
no later candidate succeeds, discovery can incorrectly return legacy fallback
metadata and downstream clients can conclude that OAuth is unsupported.

Treat 408 and 429 as discovery errors in the same central request path as 5xx.
This preserves the failing URL and HTTP status and stops discovery immediately,
without introducing retry policy or changing the public API.

HTTP 401 remains available for `WWW-Authenticate` challenge processing. HTTP 404
and 405 continue to select subsequent discovery candidates and preserve legacy
endpoint fallback.

## How Has This Been Tested?

- Added six parameterized regression cases for 408 and 429 at each discovery
  stage: the resource endpoint, protected-resource metadata, and authorization-
  server metadata.
- Verified all six cases fail on upstream `main` before the production fix.
- Asserted that each failure identifies the actual URL and status and stops
  before making another request.
- Parameterized the existing legacy fallback regression to independently cover
  both 404 and 405 while preserving the exact discovery request sequence.
- Ran the upstream all-features test suite:

  ```sh
  rustup run 1.96 cargo test --all-features
  ```

- Ran the upstream all-targets, all-features Clippy check:

  ```sh
  rustup run 1.96 cargo clippy --all-targets --all-features -- -D warnings
  ```

- Checked formatting using nightly rustfmt:

  ```sh
  rustup run nightly-2025-09-18 cargo fmt --all -- --check
  ```

## Breaking Changes

No public API or wire-format changes. OAuth metadata discovery now reports HTTP
408 and 429 as temporary failures instead of incorrectly treating them as
missing metadata.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -2635,10 +2635,12 @@ impl AuthorizationManager {
         request: OAuthHttpRequest,
     ) -> Result<HttpResponse, OAuthHttpClientError> {
         let response = self.http_client.execute(request).await?;
-        if response.status().is_server_error() {
-            return Err(Box::new(OAuthHttpError::UnexpectedStatus(
-                response.status(),
-            )));
+        let status = response.status();
+        if status.is_server_error()
+            || status == StatusCode::REQUEST_TIMEOUT
+            || status == StatusCode::TOO_MANY_REQUESTS
+        {
+            return Err(Box::new(OAuthHttpError::UnexpectedStatus(status)));
         }
         Ok(response)
     }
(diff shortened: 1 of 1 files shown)


tsarlandie-oai:
Closing as a duplicate of #1071 so review, CI, and follow-up stay in one canonical draft.

PR #1000, opened 2026-07-16
feat: add subscription listen streams (SEP-2575)

DaleSeo:
## Motivation and Context

[SEP-2575](https://modelcontextprotocol.io/seps/2575-stateless-mcp#sep-2575-make-mcp-stateless) replaces the old way of delivering notifications through `resources/subscribe`, `resources/unsubscribe`, and standalone Streamable HTTP GET streams. It introduces a new transport-neutral request called `subscriptions/listen`. This update brings in typed models and high-level server/client APIs that ensure acknowledgment ordering, subscription ID tagging, opt-in filtering, concurrent stream isolation, cancellation, and smooth or abrupt termination for both stdio and Streamable HTTP. 

We still support legacy resource subscriptions, HTTP sessions, GET/DELETE, and resumable streams for older protocol versions.

## How Has This Been Tested?

Add tests. The conformance suite passes

## Breaking Changes

No existing stable-protocol behavior is removed. The legacy `subscribe()` and `unsubscribe()` APIs are deprecated for the 2026 protocol revision but retain their existing wire behavior for older negotiated versions.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1078, opened 2026-07-29
fix: stamp server info on graceful subscription results

mrcs64:
Stamp the server implementation identity into the final result produced when a
`subscriptions/listen` handler completes gracefully.

## Motivation and Context

The MCP 2026-07-28 specification says servers SHOULD include
[`io.modelcontextprotocol/serverInfo` in every result's `_meta`](https://modelcontextprotocol.io/specification/2026-07-28/basic#general-fields).
`ServerHandler::listen` returns `Result<(), McpError>`, and rmcp constructs the
final `SubscriptionsListenResult` only after the handler returns. Applications
therefore cannot add this metadata themselves.

This PR intentionally covers only the graceful `subscriptions/listen` result
synthesized by rmcp after the application handler returns. SDK-wide stamping of
application-produced results is a separate concern.

## How Has This Been Tested?

- Added a stdio regression test that asserts the graceful final result contains
  the server implementation identity.
- Added the same assertion to the real Streamable HTTP/SSE graceful-close test.
- Ran the full repository test recipe, including the all-features and
  non-local-feature passes.
- Ran `cargo clippy --all-targets --all-features -- -D warnings`.
- Ran `cargo +nightly fmt --all -- --check`.

## Breaking Changes

None. This adds missing response metadata without changing the handler API.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The request path captures `ServerInfo` once before establishing the
subscription, so capability filtering and the eventual server identity come
from the same snapshot. Cancellation, abrupt transport closure, and handler
errors keep their existing behavior.

### DIFF
--- a/crates/rmcp/src/handler/server.rs
+++ b/crates/rmcp/src/handler/server.rs
@@ -177,11 +177,7 @@ impl<H: ServerHandler> Service<RoleServer> for H {
                     // server teardown; explicit stdio cancellation remains a notification.
                     self.listen(subscription).await.map(|()| {
                         let mut result = SubscriptionsListenResult::complete(subscription_id);
-                        result.meta.insert(
-                            SERVER_INFO_META_KEY.to_owned(),
-                            serde_json::to_value(server_implementation)
-                                .expect("Implementation serialization cannot fail"),
-                        );
+                        result.meta.set_server_info(server_implementation);
                         ServerResult::SubscriptionsListenResult(result)
                     })
                 }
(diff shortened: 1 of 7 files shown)


DaleSeo (approved):
Thanks for the fix, @mrcs64!

DaleSeo:
Just for context, I initially implemented the shape documented in SEP-2575 in #1000. However, the final spec added a graceful result for `subscriptions/listen` and made `serverInfo` optional in the result metadata. Since the SEP page was not updated, those changes were easy to miss, so I opened modelcontextprotocol/modelcontextprotocol#3159 to document these post-final SEP changes for future implementers.

PR #1085, opened 2026-07-29
feat: make structured mirroring explicit

DaleSeo:
## Motivation and Context

Automatic mirroring can double large structured payloads and makes a compatibility fallback an implicit SDK policy. This implements the alternative discussed in #1046: structured output is the default, and callers that need text content choose it explicitly without adding another return-type wrapper or macro behavior.

## How Has This Been Tested?

Added integration tests

## Breaking changes

Technically none because the API hasn't changed.

However, code that relies on `structured`, `structured_error`, or `Json<T>` to automatically populate `content` now needs to add the fallback explicitly with `with_content` or construct the `CallToolResult` directly.

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/handler/server/wrapper/json.rs
+++ b/crates/rmcp/src/handler/server/wrapper/json.rs
@@ -13,7 +13,7 @@ use crate::{
 /// When used with tools, this wrapper indicates that the value should be
 /// serialized as structured JSON content with an associated schema.
 /// The framework will place the JSON in the `structured_content` field
-/// of the tool result rather than the regular `content` field.
+/// of the tool result and leave the regular `content` field empty.
 #[expect(clippy::exhaustive_structs, reason = "intentionally exhaustive")]
 pub struct Json<T>(pub T);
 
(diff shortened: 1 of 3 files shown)


alexhancock:
cc @olaservo can you review?

olaservo:
Hi @alexhancock @DaleSeo , thanks for putting together this alternative!  After reviewing the surveyed clients again and considering that mirroring is a `SHOULD` in the spec, IMO it would be best to add an opt out for mirroring at this point, instead of an opt-in.  And hopefully by version 4.0 we will figure out better protocol semantics for explicitly choosing the outputs and destinations.  :)

For reference, here is a snapshot of what a sample of clients do with the different payloads: https://gist.github.com/olaservo/3bb819673c444e4fe282c3af44a1ae01

DaleSeo:
Thanks, @olaservo, for the review and client survey. After comparing the current Tier 1 SDKs, I agree that keeping mirroring on by default with an explicit opt-out is the better direction for v3.
I'll close this in favor of your PR and help update it so we can move forward with the opt-out.

PR #1088, opened 2026-07-29
fix: fall back to legacy initialize on general errors

### LINKED ISSUE #1040: `ClientLifecycleMode::Auto` does not fall back for deployed legacy-server responses
## Summary

`ClientLifecycleMode::Auto` only retries legacy `initialize` when
`server/discover` fails with JSON-RPC `METHOD_NOT_FOUND` (`-32601`). Existing
production MCP servers communicate unsupported modern discovery through other
recognizable responses, so enabling modern negotiation breaks servers that
previously worked.

## Observed against deployed servers

Testing protocol `2026-07-28` against public legacy MCP services reproduced the
problem with:

- OpenAI documentation, which exposes five tools through legacy initialization.
- Microsoft Learn, which exposes three tools.
- DeepWiki, which exposes three tools.

All three worked before modern discovery was enabled. With the modern lifecycle
enabled, startup failed with:

```text
Unsupported protocol version: 2026-07-28
```

Codex’s existing rmcp 1.8 implementation connected to the same servers by retrying initialize with protocol 2025-06-18.

## Current behavior

```rust
match discover_startup(...).await {
    Ok(()) => {}
        if error.code == ErrorCode::METHOD_NOT_FOUND =>
    {
        legacy_startup(...).await?;
    }
    Err(error) => return Err(error),
}
```

Production-shaped HTTP prevalidation responses observed during interoperability
testing include:

```json
{
  "jsonrpc": "2.0",
  "id": null,
  "error": {
    "code": -32000,
    "message": "Bad Request: Unsupported protocol version: 2026-07-28 (supported versions: 2025-11-25, 2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07)"
  }
}
```

```json
{
  "jsonrpc": "2.0",
  "id": null,
  "error": {
    "code": -32000,
    "message": "Bad Request: No valid session ID provided"
  }
}
```

Both occur as initial HTTP 400 responses before a session exists. Other useful
legacy-only signals include correlated unsupported-version errors, discovery
results advertising exclusively historical protocol versions, and initial HTTP
404/405 responses.

## Observed versus expected lifecycle

Actual:

```text
→ server/discover, protocol 2026-07-28
← recognizable legacy-only rejection
✗ startup terminates without attempting initialize
```

Expected:

```text
→ server/discover, protocol 2026-07-28
← recognizable legacy-only rejection
→ initialize, protocol 2025-06-18
← successful initialization
→ notifications/initialized
→ tools/list
```

Integration tests now reproduce these production-shaped responses and verify
the successful sequence
`["server/discover", "initialize", "notifications/initialized"]`.

## Downgrade safety

Fallback must remain conservative. Unrelated response IDs, arbitrary null-ID
errors, unknown or future protocol versions, HTTP 401/403 responses, existing-
session 404 responses, and ordinary modern-server rejections must not trigger
downgrade.

The observed public-server failures and successful reference-client comparison
demonstrate an existing interoperability regression rather than a hypothetical
extension to automatic lifecycle negotiation.

jamadeo:
fixes #1040

## Motivation and Context
The fallback logic was too strict, only falling back on a single METHOD_NOT_FOUND error, while the spec states we should fall back on any error that does not indicate a modern server.

## How Has This Been Tested?
Tested against FastMCP+goose

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -568,7 +568,8 @@ pub enum ClientLifecycleMode {
     Discover {
         preferred_versions: Vec<ProtocolVersion>,
     },
-    /// Probe with `server/discover`, falling back only when the peer proves it is legacy.
+    /// Probe with `server/discover`, falling back to legacy initialization on failures that do
+    /// not identify a modern server.
     Auto {
         preferred_versions: Vec<ProtocolVersion>,
         legacy_version: Option<ProtocolVersion>,
@@ -718,9 +719,17 @@ where
             .await;
             match discover_result {
                 Ok(()) => {}
-                Err(ClientInitializeError::JsonRpcError(error))
-                    if error.code == crate::model::ErrorCode::METHOD_NOT_FOUND =>
-                {
+                // `UnsupportedProtocolVersionError` is handled by `discover_startup`, which
+                // retries with an advertised compatible version. Reaching this point with no
+                // compatible version therefore positively identifies a modern server and must
+                // not trigger the legacy handshake.
+                Err(error @ ClientInitializeError::NoCompatibleProtocolVersion { .. }) => {
+                    return Err(error);
+                }
+                // On stdio, legacy servers may reject an unknown pre-initialize request with
+                // any implementation-defined error (or not respond at all). Treat every
+                // failure not recognized as modern version negotiation as a legacy peer.
+                Err(_) => {
                     let mut legacy_info = client_info;
                     if let Some(version) = legacy_version {
                         legacy_info.protocol_version = version;
@@ -728,7 +737,6 @@ where
                     legacy_startup(&service, &mut transport, &id_provider, &peer, legacy_info)
                         .await?;
                 }
-                Err(error) => return Err(error),
             }
         }
     }
--- a/crates/rmcp/tests/test_client_lifecycle_modes.rs
+++ b/crates/rmcp/tests/test_client_lifecycle_modes.rs
@@ -338,6 +338,66 @@ async fn auto_startup_falls_back_after_discover_method_not_found() {
 }
 
 #[tokio::test]
+async fn auto_startup_falls_back_after_discover_invalid_params() {
+    let (server_transport, client_transport) = tokio::io::duplex(4096);
+    let mut server = IntoTransport::<rmcp::RoleServer, _, _>::into_transport(server_transport);
+    let server_task = tokio::spawn(async move {
+        let ClientJsonRpcMessage::Request(discover) =
+            server.receive().await.expect("expected discover request")
+        else {
+            panic!("expected request");
+        };
+        assert!(matches!(
+            discover.request,
+            ClientRequest::DiscoverRequest(_)
+        ));
+        server
+            .send(ServerJsonRpcMessage::error(
+                ErrorData::new(ErrorCode::INVALID_PARAMS, "Invalid params", None),
+                Some(discover.id),
+            ))
+            .await
+            .expect("send invalid-params");
+
+        let ClientJsonRpcMessage::Request(initialize) =
+            server.receive().await.expect("expected initialize request")
+        else {
+            panic!("expected request");
+        };
+        assert!(matches!(
+            initialize.request,
+            ClientRequest::InitializeRequest(_)
+        ));
+        server
+            .send(ServerJsonRpcMessage::response(
+                ServerResult::InitializeResult(
+                    InitializeResult::new(ServerCapabilities::default()),
+                ),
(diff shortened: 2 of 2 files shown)


crates/rmcp/src/service/client.rs
```diff
@@ -718,17 +719,24 @@ where
...
+                // retries with an advertised compatible version. Reaching this point with no
+                // compatible version therefore positively identifies a modern server and must
+                // not trigger the legacy handshake.
+                Err(error @ ClientInitializeError::NoCompatibleProtocolVersion { .. }) => {
+                    return Err(error);
+                }
+                // On stdio, legacy servers may reject an unknown pre-initialize request with
+                // any implementation-defined error (or not respond at all). Treat every
+                // failure not recognized as modern version negotiation as a legacy peer.
+                Err(_) => {
```
DaleSeo, line 732: Should we also consider this in the issue?
Image

crates/rmcp/tests/test_client_lifecycle_modes.rs
```diff
@@ -337,6 +337,66 @@ async fn auto_startup_falls_back_after_discover_method_not_found() {
     server_task.await.expect("server task");
 }
 
+#[tokio::test]
+async fn auto_startup_falls_back_after_discover_invalid_params() {
```
DaleSeo, line 341: Would it be worth adding a test to catch a future `ClientInitializeError` variant being silently swept into `Err(_)`?

crates/rmcp/src/service/client.rs
jamadeo: The fix might not be quite as straightforward as I'd hoped. The core of the issue is

> A client that needs to interoperate with both kinds of servers detects the server’s era with *transport-specific mechanics*, specified in the binding pages:
> - [stdio](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/stdio#backward-compatibility): probe with server/discover and fall back on any error that is not a recognized modern error.
> - [Streamable HTTP](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#backward-compatibility): attempt a modern request and inspect the body of a 400 Bad Request before falling back.

from [here](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#backward-compatibility-with-initialization-based-versions)

so, I think the proper fix will need to implement this slightly differently.

chrisbennight:
I ran into issues form this in rcmp - flagged here:  https://github.com/modelcontextprotocol/rust-sdk/issues/1040#issuecomment-5161948549

I think the relevant bit is other servers are giving transport level errors (rcmp and mcp-go flagged in that coment), and at a quick glance this pr looks like it retries on the same http worker, which would be dead I think.

ip2a:
Confirming this fix resolves my case — a deployed `2025-03-26` server that enforces session IDs via middleware, the same class of failure @chrisbennight described in #1040.

**Repro:** a deployed streamable-HTTP MCP server (self-reported `serverInfo`: "Weather Service" v2.x), protocol `2025-03-26`, session enforcement enabled.

A stateless `server/discover` probe gets rejected by the session layer **before method dispatch**, so it never reaches a `-32601`:

```http
POST /mcp
(no Mcp-Session-Id header)

HTTP/1.1 400 Bad Request
mcp-session-id: 717fee78781b4526a207c98334554b8b
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Bad Request: Missing session ID"}}
```

(Note the server both *allocates* a session id and rejects the request for not carrying one — the session middleware runs before method dispatch, so `-32601` is unreachable here.)

**Behavior:**
- `ClientLifecycleMode::Auto` (current rmcp 3.1.0): fails with the `-32600` above, since the fallback only triggers on `-32601`.
- `ClientLifecycleMode::Initialize`: connects and lists tools successfully.
- With this PR's `Err(_) => legacy fallback`: the `-32600` is no longer in the `METHOD_NOT_FOUND`-only path, so Auto would fall back as intended. 

The `NoCompatibleProtocolVersion => return Err` carve-out is the right disambiguator — it preserves the one signal that positively identifies a modern server.

One suggestion to strengthen coverage: the added test covers `INVALID_PARAMS` (-32602). Would be worth adding a `-32600` case too, since session/auth middleware rejecting a sessionless probe is a distinct and fairly common failure shape on streamable-HTTP. Happy to push a test commit if helpful.

Thanks for the clean, minimal fix — this unblocks Auto against real deployed servers.

jamadeo:
Closing at this was a very incomplete fix that was superseded by #1133

PR #1097, opened 2026-07-30
fix(model): decode metadata-bearing input-required results affecting mrtr

### LINKED ISSUE #1098: InputRequiredResult metadata causes input and resume state to be dropped - mrtr requests sep 2322
**Describe the bug**
While testing the conformance suite we found issues with some of the mrtr tests with missing inputRequests and requestState

**To Reproduce**
Deserialize this valid server result through `ServerResult` / `ServerJsonRpcMessage`:

     ```json
     {
       "resultType": "input_required",
       "_meta": {},
       "inputRequests": {
         "confirm": {
           "method": "elicitation/create",
           "params": {
             "mode": "form",
             "message": "Please confirm",
             "requestedSchema": {
               "type": "object",
               "properties": {
                 "ok": { "type": "boolean" }
               },
               "required": ["ok"]
             }
           }
         }
       },
       "requestState": "opaque-server-state"
     }

  Observe that it is parsed as ServerResult::CallToolResult, rather than ServerResult::InputRequiredResult.

**Expected behavior**
A result with resultType: "input_required" should always deserialize as ServerResult::InputRequiredResult, preserving both inputRequests and requestState for the client retry.

**Logs**

**Additional context**
Add any other context about the problem here.

filintod:
The draft permits every Result, including InputRequiredResult, to carry
  _meta and recommends that servers include
  io.modelcontextprotocol/serverInfo on every response. It also requires
  an InputRequiredResult to contain inputRequests or requestState so the
  client knows what to provide or how to resume the request.

  ServerResult uses untagged deserialization and attempts CallToolResult
  before InputRequiredResult. Because CallToolResult accepts _meta and
  defaults missing content to an empty list, it consumed valid
  input_required results carrying metadata. Unknown inputRequests and
  requestState fields were then discarded, leaving clients without the
  information needed to continue.

  Reject resultType input_required when deserializing CallToolResult so
  Serde proceeds to the InputRequiredResult variant. Add regression
  coverage proving that inputRequests, requestState, and serverInfo
  metadata are preserved.

  Spec:
  - https://github.com/modelcontextprotocol/modelcontextprotocol/blob/41d9e938e9a9edf23a69429be180cad172e00f4e/schema/draft/schema.ts#L137-L151
- https://github.com/modelcontextprotocol/modelcontextprotocol/blob/41d9e938e9a9edf23a69429be180cad172e00f4e/schema/draft/schema.ts#L198-L224
 - https://github.com/modelcontextprotocol/modelcontextprotocol/blob/41d9e938e9a9edf23a69429be180cad172e00f4e/schema/draft/schema.ts#L541-L565

Fixes https://github.com/modelcontextprotocol/rust-sdk/issues/1098

## Motivation and Context

Found while checking against the conformance tests and several of mrtr 2322 tests failed with missing inputRequests and/or requestState

A valid MRTR `InputRequiredResult` with result metadata can be misclassified as `CallToolResult` during untagged deserialization. The SDK then drops `inputRequests` and `requestState`, preventing clients from presenting the required elicitation or retrying with the server state.

## How Has This Been Tested?

Re-run the conformance tests after building locally and tests passed

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -3814,6 +3814,8 @@ pub struct CallToolResult {
 // 2. Requires at least one known field to be present, so that `CallToolResult` doesn't
 //    greedily match arbitrary JSON objects when used inside `#[serde(untagged)]` enums
 //    (e.g. `ServerResult`), which would shadow `CustomResult`.
+// 3. Rejects `resultType: "input_required"` so untagged `ServerResult`
+//    decoding selects `InputRequiredResult` and preserves input requests and state.
 impl<'de> Deserialize<'de> for CallToolResult {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
@@ -3832,6 +3834,16 @@ impl<'de> Deserialize<'de> for CallToolResult {
         }
 
         let helper = Helper::deserialize(deserializer)?;
+
+        if helper
+            .result_type
+            .as_ref()
+            .is_some_and(ResultType::is_input_required)
+        {
+            return Err(serde::de::Error::custom(
+                "CallToolResult cannot use resultType \"input_required\"",
+            ));
+        }
 
         if helper.content.is_none()
             && helper.structured_content.is_none()
--- a/crates/rmcp/src/model/mrtr.rs
+++ b/crates/rmcp/src/model/mrtr.rs
@@ -274,6 +274,12 @@ impl<'de> Deserialize<'de> for InputRequiredResult {
             }
         }
 
+        if helper.input_requests.is_none() && helper.request_state.is_none() {
+            return Err(serde::de::Error::custom(
+                "InputRequiredResult requires at least one of inputRequests or requestState",
+            ));
+        }
+
         Ok(InputRequiredResult {
             result_type: ResultType::INPUT_REQUIRED,
             input_requests: helper.input_requests,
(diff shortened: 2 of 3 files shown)


crates/rmcp/src/model.rs
```diff
@@ -3833,6 +3835,16 @@ impl<'de> Deserialize<'de> for CallToolResult {
...
 
+        if helper
+            .result_type
+            .as_ref()
+            .is_some_and(ResultType::is_input_required)
+        {
+            return Err(serde::de::Error::custom(
+                "CallToolResult cannot use resultType \"input_required\"",
+            ));
+        }
```
DaleSeo, line 3846: After this rejection, how should `ServerResult` handle `{"resultType":"input_required","_meta":{}}`? The draft and `InputRequiredResult` documentation require at least one of `inputRequests` or `requestState`, but the deserializer currently accepts neither. That leaves the client with no way to fulfill or resume the request.

filintod: good point, I'll add another guard for at least one of them

filintod: done, added a couple the new guard and tests. one thing is that I'm only checking presence, the spec and schema don't say much about empty value

DaleSeo (approved):

PR #1104, opened 2026-07-31
fix: expose MRTR state to tool handlers

### LINKED ISSUE #1095: `#[tool]` drops `request_state`/`input_responses`, so MRTR handlers cannot use it
MRTR (SEP-2322) gives `tools/call` two new inputs: `request_state`, the opaque continuation the
client echoes back, and `input_responses`, the client's answers to the previous round's
`inputRequests`. Both live on `CallToolRequestParams`.

A handler written with `#[tool]` cannot see either. `ToolCallContext::new` takes the params apart
and forwards only the arguments, so the macro-generated body has no access to the two fields that
make a multi-round call multi-round.

The consequence is that **every MRTR server in Rust must hand-write `ServerHandler::call_tool`**,
giving up the macro's routing, schema derivation, and argument extraction for the whole server —
not just for the tool that needs MRTR.

## Why this is worth fixing

MRTR is not a niche extension. SEP-2322 routes *all* server-initiated interaction through it:
`elicitation/create`, `sampling/createMessage`, and `roots/list` no longer exist as
server-to-client requests. Any tool that needs to ask the user anything is an MRTR tool. The
ergonomic path in the SDK currently excludes the entire category.

## Suggested fix

Extractors, in the style the SDK already uses for `Extension`:

```rust
#[tool(description = "book a trip")]
async fn book_trip(
    &self,
    Parameters(args): Parameters<BookTrip>,
    RequestState(state): RequestState,          // Option<String>
    InputResponses(answers): InputResponses,    // Option<BTreeMap<String, Value>>
) -> Result<CallToolResponse, ErrorData> { … }
```

That would need `ToolCallContext` to carry the two fields through rather than drop them, and a
return type of `CallToolResponse` rather than `CallToolResult` so a handler can answer
`InputRequired`.

A smaller first step that would unblock people: keep the fields on `ToolCallContext` and expose
them through an accessor, so a macro-based handler can reach them even if the extractor sugar
comes later.

DaleSeo:
Fixes #1095.

## Motivation and Context

`ToolCallContext::new` was dropping `request_state` and `input_responses`, so `#[tool]` handlers could not continue an MRTR call after the client completed its input requests. Because of that, servers had to bypass the macro routing path and implement `ServerHandler::call_tool` manually.

This PR preserves the SEP-2322 retry state and input responses when macro-generated tool handlers route calls through `ToolCallContext`. It also adds extractors for both values and lets `#[tool]` handlers return `CallToolResponse`. This allows a single handler to produce both `input_required` responses and completed results.

## How Has This Been Tested?

Added regression tests

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/handler/server/tool.rs
+++ b/crates/rmcp/src/handler/server/tool.rs
@@ -34,9 +34,13 @@ pub fn parse_json_object<T: DeserializeOwned>(input: JsonObject) -> Result<T, cr
 }
 #[non_exhaustive]
 pub struct ToolCallContext<'s, S> {
+    /// The request-specific context for this tool call.
     pub request_context: RequestContext<RoleServer>,
+    /// The server handling this tool call.
     pub service: &'s S,
+    /// The name of the tool being called.
     pub name: Cow<'static, str>,
+    /// The arguments supplied for the tool call.
     pub arguments: Option<JsonObject>,
     /// Client responses to input requests from the previous MRTR round.
     pub input_responses: Option<crate::model::InputResponses>,
--- a/crates/rmcp/src/handler/server/tool.rs
+++ b/crates/rmcp/src/handler/server/tool.rs
@@ -38,6 +38,10 @@ pub struct ToolCallContext<'s, S> {
     pub service: &'s S,
     pub name: Cow<'static, str>,
     pub arguments: Option<JsonObject>,
+    /// Client responses to input requests from the previous MRTR round.
+    pub input_responses: Option<crate::model::InputResponses>,
+    /// Opaque state returned by the server during the previous MRTR round.
+    pub request_state: Option<String>,
 }
 
 impl<'s, S> ToolCallContext<'s, S> {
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/handler/server/tool.rs
```diff
@@ -38,6 +38,10 @@ pub struct ToolCallContext<'s, S> {
     pub service: &'s S,
     pub name: Cow<'static, str>,
     pub arguments: Option<JsonObject>,
+    /// Client responses to input requests from the previous MRTR round.
```
alexhancock, line 45: I'd strip these comments

alexhancock (dismissed):

crates/rmcp/src/handler/server/tool.rs
DaleSeo: @alexhancock Would you be open to documenting the existing fields instead? Since this struct is public, I think rustdoc for all of its fields would be useful and more consistent.

DaleSeo: I've added comments to the existing fields.

alexhancock (approved):

PR #1105, opened 2026-07-31
feat: update LATEST

alexhancock:
Simple update of the `LATEST` alias to `V_2026_07_28`

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -172,7 +172,7 @@ impl ProtocolVersion {
     pub const V_2025_06_18: Self = Self(Cow::Borrowed("2025-06-18"));
     pub const V_2025_03_26: Self = Self(Cow::Borrowed("2025-03-26"));
     pub const V_2024_11_05: Self = Self(Cow::Borrowed("2024-11-05"));
-    pub const LATEST: Self = Self::V_2025_11_25;
+    pub const LATEST: Self = Self::V_2026_07_28;
 
     /// First protocol version that requires SEP-2243 standard HTTP headers.
     pub const STANDARD_HEADERS: Self = Self::V_2026_07_28;
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -472,15 +472,23 @@ pub(crate) fn negotiate_protocol_version(
     server_supported: &[ProtocolVersion],
 ) -> ProtocolVersion {
     if server_supported.contains(client_requested) {
-        client_requested.clone()
-    } else {
-        tracing::warn!(
-            client_requested = %client_requested,
-            server_fallback = %server_fallback,
-            "client requested unsupported protocol version; falling back to server default"
-        );
-        server_fallback
+        return client_requested.clone();
     }
+    let fallback = if server_supported.is_empty() || server_supported.contains(&server_fallback) {
+        server_fallback
+    } else {
+        server_supported
+            .iter()
+            .max_by(|a, b| a.as_str().cmp(b.as_str()))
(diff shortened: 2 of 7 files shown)


crates/rmcp/src/model.rs
```diff
@@ -172,7 +172,7 @@ impl ProtocolVersion {
     pub const V_2025_06_18: Self = Self(Cow::Borrowed("2025-06-18"));
     pub const V_2025_03_26: Self = Self(Cow::Borrowed("2025-03-26"));
     pub const V_2024_11_05: Self = Self(Cow::Borrowed("2024-11-05"));
-    pub const LATEST: Self = Self::V_2025_11_25;
+    pub const LATEST: Self = Self::V_2026_07_28;
```
DaleSeo, line 175: The hardcoded version causes the following tests to fail:

https://github.com/modelcontextprotocol/rust-sdk/blob/830e088d733c7964c806a2305760dd8deb30dff9/crates/rmcp/tests/test_server_initialization.rs#L269-L273

https://github.com/modelcontextprotocol/rust-sdk/blob/830e088d733c7964c806a2305760dd8deb30dff9/crates/rmcp/tests/test_server_initialization.rs#L290-L294

DaleSeo (approved):

crates/rmcp/tests/test_streamable_http_session_store.rs
```diff
@@ -108,7 +109,10 @@ async fn test_session_state_persisted_to_store() -> anyhow::Result<()> {
     let transport = StreamableHttpClientTransport::from_config(
         StreamableHttpClientTransportConfig::with_uri(format!("http://{addr}/mcp")),
     );
-    let client = ().serve(transport).await?;
+    let client = ClientInfo::default()
+        .with_protocol_version(ProtocolVersion::V_2025_11_25)
+        .serve(transport)
+        .await?;
```
DaleSeo, line 115: Once we pin this, these tests will only cover `2025-11-25` behavior, and nothing will verify what happens with the new default. What do you feel about adding a companion test that explicitly checks that no session is created at `LATEST`?

crates/rmcp/src/service/server.rs
DaleSeo, line 464: The function no longer returns `server_fallback` as-is when it isn't in `server_supported`.

crates/rmcp/src/service/server.rs
```diff
@@ -472,15 +472,23 @@ pub(crate) fn negotiate_protocol_version(
...
+        server_supported
+            .iter()
+            .max_by(|a, b| a.as_str().cmp(b.as_str()))
+            .cloned()
+            .unwrap_or(server_fallback)
+    };
+    tracing::warn!(
+        client_requested = %client_requested,
+        server_fallback = %fallback,
+        "client requested unsupported protocol version; falling back to server default"
```
DaleSeo, line 489: When the handler's declared fallback isn't in `supported_protocol_versions`, this field carries the computed replacement.

```suggestion
        "client requested unsupported protocol version; falling back to a supported version"
```

DaleSeo (commented):
@alexhancock Once the default becomes `2026-07-28`, even the `initialize` handshake will be handled statelessly, so no session will be created. That follows the SEP-2567 direction, but it also means that deployments relying on `Mcp-Session-Id` could be affected by the upgrade alone, with nothing in the API to signal the change. I'm a bit concerned about the impact, even though this is technically a non-breaking change.

DaleSeo:
@alexhancock Any updates on this?

alexhancock:
@DaleSeo Sorry I got away from this for a bit. Revising now

PR #1111, opened 2026-08-01
fix: support nested MCP header property paths

nolankhoingo:
## Summary

- accept primitive x-mcp-header annotations reached through plain properties chains
- emit and validate Mcp-Param headers from the corresponding nested argument path
- reject annotations reached through arrays, composition, conditionals, or references

## Why

MCP 2026-07-28 permits statically reachable primitive annotations under plain properties chains. The current top-level-only cache rejects otherwise valid tools.

Specification: https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http

## Validation

cargo test -p rmcp --lib --features client,transport-streamable-http-client,server-side-http transport::common::mcp_headers::tests

All 36 focused header tests pass.

### DIFF
--- a/crates/rmcp/src/transport/common/mcp_headers.rs
+++ b/crates/rmcp/src/transport/common/mcp_headers.rs
@@ -99,80 +99,135 @@ fn is_tchar(c: char) -> bool {
         )
 }
 
-/// Top-level properties carrying an `x-mcp-header` annotation, as `(property, header)` pairs.
-fn param_header_annotations(input_schema: &JsonObject) -> Vec<(String, String)> {
+/// Properties carrying an `x-mcp-header` annotation, as `(property path, header)` pairs.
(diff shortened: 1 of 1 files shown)


crates/rmcp/src/transport/common/mcp_headers.rs
```diff
@@ -99,82 +99,137 @@ fn is_tchar(c: char) -> bool {
...
 }
 
+#[cfg(feature = "client-side-sse")]
+fn contains_annotation(value: &Value) -> bool {
+    match value {
+        Value::Object(object) => {
+            object.contains_key("x-mcp-header") || object.values().any(contains_annotation)
+        }
+        Value::Array(values) => values.iter().any(contains_annotation),
+        _ => false,
```
DaleSeo, line 229: How could this distinguish subschemas from ordinary JSON values so that something like `default: {"x-mcp-header":"literal"}` does not cause a valid tool to be rejected?

PR #1119, opened 2026-08-03
chore: make async-trait optional

### LINKED ISSUE #1112: async-trait: make it an optional dependency
## Problem

`async-trait` is declared as an unconditional dependency in `crates/rmcp/Cargo.toml`, but
every use of it sits behind non-default features:

- `CredentialStore` / `StateStore` (`transport/auth.rs`) — feature `auth`
- `EventStore` / `SessionStore` (`session/store.rs`) — feature `transport-streamable-http-server-session`

Default features (`base64`, `macros`, `server`) pull in neither, yet the proc-macro crate
still compiles for every consumer:

```bash
$ cargo tree -p rmcp -e normal --no-default-features --features base64,macros,server -i async-trait
async-trait (proc-macro)
└── rmcp
```

## Proposal

```toml
# crates/rmcp/Cargo.toml
async-trait = { version = "0.1.89", optional = true }

[features]
auth = [..., "dep:async-trait"]
transport-streamable-http-server-session = [..., "dep:async-trait"]
```
Dev-dependency stays as-is (used directly by the event/session store integration tests).
Non-breaking, no public API change — trims the default build graph only.

DaleSeo:
Closes #1112.

## Motivation and Context

The default `rmcp` feature set does not compile any code that uses `async-trait`, but the unconditional dependency still adds the proc-macro crate to every consumer's build graph. Moving the dependency behind the two features that use it keeps the build graph smaller without changing the public API or runtime behavior.

This PR makes `async-trait` optional. Consumers that do not enable OAuth or streamable HTTP server sessions will no longer compile an unused proc macro. The `auth` and `transport-streamable-http-server-session` features now enable the dependency explicitly.

## How Has This Been Tested?

- Verified `async-trait` is absent with `--no-default-features --features base64,macros,server`.
- Verified `async-trait` is present with `auth` and `transport-streamable-http-server-session` independently.

## Breaking Changes

None. Existing features continue to enable `async-trait` wherever it is required.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -45,7 +45,7 @@ features = [
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-async-trait = "0.1.89"
+async-trait = { version = "0.1.89", optional = true }
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 thiserror = "2"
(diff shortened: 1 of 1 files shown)


alexhancock (approved):
Good catch. A nice improvement

PR #1128, opened 2026-08-04
feat: add request-state key rotation

### LINKED ISSUE #1096: `rs1.` request-state format carries no `kid`, so key rotation cannot be targeted
`RequestStateCodec`'s wire format is `rs1.<base64url(expiry ‖ payload)>.<base64url(tag)>`. 
The `rs1` prefix versions the *format*.
Nothing in the token identifies the *key* that produced the tag.

A server that rotates its signing key therefore has two options, and both are unpleasant:

1. **Rotate and break.** Every `requestState` in flight becomes unverifiable the moment the new
   key is installed. For a multi-round tool call that is every dialogue a user has open.
2. **Trial-open.** Keep the previous key and attempt `open_with` against each key in turn. This
   works — it is what this demo does — but it costs an extra HMAC verification per request for
   every key in the window, and it obliges every server author to rediscover the pattern.

## Why it matters here

MRTR makes the codec's key a **deployment-wide shared secret**: every replica that might
continue a dialogue must hold it. That is precisely the population most likely to need staged
rotation, and least able to coordinate an atomic switch.

JWS solves this with a `kid` header, JWE likewise. `rs1.` has no equivalent field.

## Suggested fix

Add an optional key identifier to the format — `rs1.<kid>.<body>.<tag>`, or a length-prefixed
`kid` inside the authenticated region — and let `RequestStateCodec` hold a small keyring keyed
by it. Opening then selects one key instead of trying all of them, and a server can publish a
new key before it starts signing with it.

If a format change is unacceptable, documenting the trial-open pattern in `RequestStateCodec`'s
type-level docs would at least stop each author from working it out alone.

gocamille:
Adds targeted key selection and rolling-safe signing-key rotation to `RequestStateCodec` for SEP-2322 `requestState` values.

Closes #1096.

## Motivation and Context

The existing `rs1.<body>.<tag>` format identifies the wire format but not the key that produced the HMAC. Rotating a deployment-wide request-state key therefore either invalidates every in-flight multi-round exchange or requires applications to trial-open each token against every retained key.

That is particularly difficult for multi-replica deployments, where every replica that may continue an MRTR exchange must share the same signing keys and cannot safely switch them atomically.

## How

- Added an `rs2.<base64url(kid)>.<base64url(expiry || payload)>.<base64url(tag)>` format. The key id is included in the HMAC input, allowing opening to select one key directly while preventing key-id substitution.
- Added `RequestStateCodec::new_with_keyring(active_kid, keys)` for keyed sealing and targeted verification. `active_kid` selects the signing key, while any configured key can verify an `rs2` value naming that key.
- Key ids are opaque, case-sensitive UTF-8 strings limited to 255 bytes. They are authenticated but visible to clients and are not confidential.
- Preserved `RequestStateCodec::new()` and the existing `rs1` wire format byte-for-byte.
- Added `with_rs1_signing()` and `with_rs1_fallback()` for a rolling-safe migration:

  1. Deploy both keys while continuing to emit `rs1` with the old key.
  2. Promote the new key and emit `rs2`, retaining the old key as an `rs1` fallback.
  3. Wait for existing request states to expire, then remove the old key.

- Legacy fallback verification evaluates every configured `rs1` fallback before returning, so the HMAC count does not reveal which fallback matched. The rustdocs recommend keeping this set small and temporary.
- Added distinct HMAC domains for `rs1` and `rs2`, with length-prefixed key ids and associated data.
- Added explicit errors for unknown key ids, invalid wire key ids, and invalid keyring configuration. Configuration messages are documented as diagnostics rather than values applications should match programmatically.
- The rustdocs require applications to map all token-opening failures to one client-visible error.
- Added wire-format, key-selection, migration, and key-retirement guidance to the codec rustdocs and README.

## How Has This Been Tested?

- `cargo test -p rmcp --features request-state model::request_state::tests --lib` — 34 tests passed.
- `cargo test -p rmcp --features request-state --doc` — 33 doctests passed, including the staged-rotation example.
- Strict all-feature rustdoc generation with warnings denied.
- Known-answer tests confirm the existing `rs1` output is unchanged and independently verify the new `rs2` format.
- Rotation tests cover old-key verification, key promotion, retirement, multiple legacy fallbacks, and bidirectional compatibility during migration.
- Security tests cover key-id, body, and tag tampering; version-domain separation; associated data; TTL enforcement; key redaction; and authentication before JSON deserialization.
- Parser tests cover malformed tokens, invalid and oversized key ids, unknown keys, stable error precedence, and arbitrary mutated inputs without panics.

## Breaking Changes

None. Existing `RequestStateCodec::new()` callers continue to emit and accept the same `rs1` format. Keyrings and `rs2` are opt-in behind the existing `request-state` feature.

New public API:

- `RequestStateCodec::new_with_keyring`
- `RequestStateCodec::with_rs1_signing`
- `RequestStateCodec::with_rs1_fallback`
- `RequestStateError::{UnknownKeyId, InvalidKeyId, InvalidKeyring}`

`RequestStateError` remains `#[non_exhaustive]`, so adding these variants does not require downstream callers to update exhaustive matches.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Note that the codec intentionally does not provide runtime key reload or a default TTL. Applications must retain old verification keys until the maximum request-state lifetime has elapsed before retiring them.

### DIFF
--- a/crates/rmcp/CHANGELOG.md
+++ b/crates/rmcp/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- add `RequestStateCodec` keyrings and authenticated `rs2` key identifiers for
+  rolling-safe key rotation while preserving `new()` and the legacy `rs1`
+  format
+
 ## [3.1.0](https://github.com/modelcontextprotocol/rust-sdk/compare/rmcp-v3.0.1...rmcp-v3.1.0) - 2026-07-31
 
 ### Added
--- a/crates/rmcp/CHANGELOG.md
+++ b/crates/rmcp/CHANGELOG.md
@@ -7,12 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- add `RequestStateCodec` keyrings and authenticated `rs2` key identifiers for
-  rolling-safe key rotation while preserving `new()` and the legacy `rs1`
-  format
-
 ## [3.1.1](https://github.com/modelcontextprotocol/rust-sdk/compare/rmcp-v3.1.0...rmcp-v3.1.1) - 2026-08-05
 
 ### Fixed
--- a/crates/rmcp/src/model/request_state.rs
+++ b/crates/rmcp/src/model/request_state.rs
@@ -465,8 +465,9 @@ impl RequestStateCodec {
     /// - [`RequestStateError::InvalidKeyId`] if an `rs2` key id is empty, too
     ///   long, or not valid UTF-8.
     ///
-    /// Callers should expose a single client-facing failure and reserve the
-    /// detailed variants for internal diagnostics.
+    /// Applications MUST map all token-opening failures to a single
+    /// client-facing error and reserve the detailed variants for internal
+    /// diagnostics.
     pub fn open_with(
         &self,
         sealed: &str,
@@ -610,6 +611,7 @@ impl RequestStateCodec {
             .decode(tag_b64)
             .map_err(|_| RequestStateError::InvalidEncoding)?;
 
+        // `verify_slice` compares tags in constant time and rejects wrong-length tags.
         match &self.keys {
             Keys::Single(key) => Self::mac_v1(key, associated_data, &body)
                 .verify_slice(&tag)
@@ -619,12 +621,15 @@ impl RequestStateCodec {
                 rs1_fallbacks,
                 ..
             } => {
-                let verified = rs1_fallbacks.iter().any(|kid| {
+                // Evaluate every fallback so the HMAC count does not reveal which key matched.
+                let mut verified = false;
+                for kid in rs1_fallbacks {
                     let key = keys.get(kid).expect("validated rs1 fallback key");
-                    Self::mac_v1(key, associated_data, &body)
+                    let matches = Self::mac_v1(key, associated_data, &body)
                         .verify_slice(&tag)
-                        .is_ok()
-                });
+                        .is_ok();
+                    verified |= matches;
+                }
                 if !verified {
                     return Err(RequestStateError::IntegrityCheckFailed);
                 }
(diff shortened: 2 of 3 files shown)


crates/rmcp/src/model/request_state.rs
```diff
@@ -148,35 +173,193 @@ impl<'a> SealOptions<'a> {
...
+    ///
+    /// All configured keys can open matching `rs2` values. Legacy `rs1` values
+    /// require [`with_rs1_fallback`](Self::with_rs1_fallback) or
+    /// [`with_rs1_signing`](Self::with_rs1_signing).
+    /// Key ids are case-sensitive UTF-8 strings of 1 to 255 bytes.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`RequestStateError::InvalidKeyring`] if the keyring is empty,
+    /// contains duplicate or invalid key ids, or does not contain `active_kid`.
```
DaleSeo, line 250: The PR description says the doctests include a staged rotation example, but I couldn't find it.

crates/rmcp/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- add `RequestStateCodec` keyrings and authenticated `rs2` key identifiers for
+  rolling-safe key rotation while preserving `new()` and the legacy `rs1`
+  format
+
```
DaleSeo, line 15: This repository's changelog is maintained automatically by release-plz. Manually updating it will only cause conflicts.

crates/rmcp/src/model/request_state.rs
gocamille: Good catch @DaleSeo the staged example was accidentally removed while streamlining the rustdocs but the PR description wasn’t updated. I’ve restored a runnable three-stage rotation example and verified it with the doctest suite.

crates/rmcp/CHANGELOG.md
gocamille: Thanks @DaleSeo I removed the manual changelog entry and left release-plz to generate it from the `feat:` change 👍

DaleSeo (approved):

ianewsfr-a11y:
Reading your PR on request-state key rotation, because we hit the same problem: per-agent secret + hash-chained events.
Question: how do you handle mid-session invalidation? Keyless read: ergonia.works/api/arena. Disclosure: I'm the human behind Ergonia.

ianewsfr-a11y:
Reading through your merged PR on request-state key rotation, because our marketplace has a per-agent-secret + hash-chained mutation shape that would need the same story eventually. Concrete question, not a pitch: given the rs1 fallback flag, how long do you recommend keeping it on in production before removing, and is there a wire-visible signal for "safe to disable" (all in-flight rs1 exchanges drained)? Keyless read at ergonia.works/api/arena
 if you want to see the marketplace side. Disclosure: I'm the human behind Ergonia. Feel free to close if wrong venue for a cold reach.

PR #1133, opened 2026-08-05
fix(client): fall back from discover for any non-modern error

### LINKED ISSUE #1040: `ClientLifecycleMode::Auto` does not fall back for deployed legacy-server responses

ip2a:
Fixes #1040.

## Problem

`ClientLifecycleMode::Auto` only fell back to the legacy `initialize`
handshake when `server/discover` failed with `-32601` (`METHOD_NOT_FOUND`).
Legacy servers commonly reject an unknown pre-`initialize` request with other
implementation-defined errors (`-32600`, `-32602`, session-middleware errors),
so Auto broke against servers that previously worked.

## Approach

Classification happens inside `discover_startup`, where the full context
(request id, response correlation, transport state) is still available — not
after the fact on `ClientInitializeError`, where that context is gone.

`discover_startup` returns a `DiscoverOutcome`:

- `Modern` — discover succeeded.
- `Legacy(error)` — the probe received a complete, correlated JSON-RPC error
  whose code is not a modern-era rejection. The transport delivered a full
  response and is ready for the next request, so a legacy `initialize` can
  follow on the same connection.
- `Err(error)` — everything else (transport failure, uncorrelated response,
  modern rejection, client-side state). Surfaced, not retried.

`Auto` simply matches the outcome. No methods on `ClientInitializeError`, no
downcast, no transport-specific types leaking into the lifecycle layer.

Two additional fixes that fall out naturally:

- **Response correlation**: `expect_response` now checks the request id on
  both success and error responses. Previously error responses skipped id
  correlation entirely. A new `UncorrelatedErrorResponse` variant surfaces
  responses that cannot be tied to the request.

- **Fallback failure preservation**: when both discover and the legacy
  fallback fail, a `LegacyFallbackFailed` compound error preserves both
  phases instead of discarding the discover error.

A silently legacy server that ignores the probe and hangs `expect_response`
is a separate concern (needs a discover timeout); tracked in #1142.

## Tests

The existing lifecycle tests verify the behavior end-to-end: `-32601`,
`-32600`, and `-32602` discover responses trigger fallback; `-32021` and
`-32020` are surfaced. The client-initialization test was fixed to echo the
real request id (it previously hardcoded `1` against an id provider that
starts at `0`, which only worked because error responses were never
correlated).

## Scope

Non-JSON HTTP responses (e.g. a legacy server returning a plain-text 422) are
not covered by this PR. Recognizing them as a legacy signal requires the
transport layer to carry a structured HTTP-status signal across the transport
boundary, which is an architectural change best discussed separately.

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -49,6 +49,14 @@ pub enum ClientInitializeError {
     #[error("conflict initialized response id: expected {0}, got {1}")]
     ConflictInitResponseId(RequestId, RequestId),
 
+    #[error(
+        "uncorrelated error response: expected id {expected}, error response carried {received}"
+    )]
+    UncorrelatedErrorResponse {
+        expected: RequestId,
+        received: RequestId,
+    },
+
     #[error("connection closed: {0}")]
     ConnectionClosed(String),
 
@@ -74,6 +82,12 @@ pub enum ClientInitializeError {
 
     #[error("Cancelled")]
     Cancelled,
+
+    #[error("discover and legacy initialize both failed")]
+    LegacyFallbackFailed {
+        discover: Box<ClientInitializeError>,
+        fallback: Box<ClientInitializeError>,
+    },
 }
 
 impl ClientInitializeError {
@@ -96,8 +110,13 @@ impl ClientInitializeError {
     pub fn auth_challenge(&self) -> Option<&str> {
         use crate::transport::streamable_http_client::{AuthRequiredError, InsufficientScopeError};
 
-        let Self::TransportError { error, .. } = self else {
-            return None;
+        let error = match self {
+            Self::TransportError { error, .. } => error,
+            // A 401/403 in the fallback phase is still actionable.
+            Self::LegacyFallbackFailed { fallback, .. } => {
+                return fallback.auth_challenge();
+            }
+            _ => return None,
         };
         let mut source: Option<&(dyn std::error::Error + 'static)> = Some(error.error.as_ref());
         while let Some(current) = source {
@@ -117,10 +136,11 @@ impl ClientInitializeError {
     /// This covers both missing or expired local OAuth authorization and an HTTP
     /// authorization challenge from the MCP server.
     pub fn is_authorization_required(&self) -> bool {
-        matches!(
-            self,
-            Self::TransportError { error, .. } if error.is_authorization_required()
-        )
+        match self {
+            Self::TransportError { error, .. } => error.is_authorization_required(),
+            Self::LegacyFallbackFailed { fallback, .. } => fallback.is_authorization_required(),
+            _ => false,
+        }
     }
 }
 
@@ -138,13 +158,20 @@ where
         .ok_or_else(|| ClientInitializeError::ConnectionClosed(context.to_string()))
 }
 
-/// Helper function to expect a response from the stream
+/// Helper function to expect a response from the stream, correlated to
+/// `expected_id`.
+///
+/// Both success and error responses are checked here: a mismatched id on a
+/// success response is `ConflictInitResponseId`; on an error response (whose
+/// `id` is optional per spec) it is `UncorrelatedErrorResponse`. The caller
+/// never sees an uncorrelated response.
 async fn expect_response<T, S>(
     transport: &mut T,
     context: &str,
     service: &S,
     peer: Peer<RoleClient>,
-) -> Result<(ServerResult, RequestId), ClientInitializeError>
+    expected_id: &RequestId,
+) -> Result<ServerResult, ClientInitializeError>
 where
     T: Transport<RoleClient>,
     S: Service<RoleClient>,
@@ -152,13 +179,29 @@ where
     loop {
         let message = expect_next_message(transport, context).await?;
         match message {
-            // Expected message to complete the initialization
             ServerJsonRpcMessage::Response(JsonRpcResponse { id, result, .. }) => {
-                break Ok((result, id));
+                if !expected_id.matches_response_id(&id) {
+                    return Err(ClientInitializeError::ConflictInitResponseId(
+                        expected_id.clone(),
+                        id,
+                    ));
+                }
+                return Ok(result);
             }
-            // Handle JSON-RPC error responses
             ServerJsonRpcMessage::Error(error) => {
-                break Err(ClientInitializeError::JsonRpcError(error.error));
+                return Err(match &error.id {
+                    Some(id) if expected_id.matches_response_id(id) => {
+                        ClientInitializeError::JsonRpcError(error.error)
+                    }
+                    // Spec: error id is optional; a server that cannot read
+                    // the request id omits it. The error is still a response
+                    // to our request, so it remains available to the caller.
+                    None => ClientInitializeError::JsonRpcError(error.error),
+                    Some(id) => ClientInitializeError::UncorrelatedErrorResponse {
+                        expected: expected_id.clone(),
+                        received: id.clone(),
+                    },
+                });
             }
             // Server could send logging messages before handshake
             ServerJsonRpcMessage::Notification(mut notification) => {
@@ -714,7 +757,7 @@ where
             legacy_startup(&service, &mut transport, &id_provider, &peer, client_info).await?;
         }
         ClientLifecycleMode::Discover { preferred_versions } => {
-            discover_startup(
+            match discover_startup(
                 &service,
                 &mut transport,
                 &id_provider,
@@ -722,13 +765,18 @@ where
                 &client_info,
                 preferred_versions,
             )
-            .await?;
+            .await?
+            {
+                DiscoverOutcome::Modern => {}
+                // Discover mode does not fall back; a legacy server is an error.
+                DiscoverOutcome::Legacy(error) => return Err(*error),
+            }
         }
         ClientLifecycleMode::Auto {
             preferred_versions,
             legacy_version,
         } => {
-            let discover_result = discover_startup(
+            match discover_startup(
                 &service,
                 &mut transport,
                 &id_provider,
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/service/client.rs
```diff
@@ -739,23 +739,56 @@ where
...
+                // version mismatch): the peer speaks the modern protocol and
+                // must not trigger a legacy fallback. Surface the error.
+                Err(error) if is_modern_server_error(&error) => return Err(error),
+                // Any other outcome — `-32601`, `-32602`, `-32600`, another
+                // implementation-defined error, or no response at all — means
+                // the server does not implement the modern discovery flow.
+                // Fall back to the legacy `initialize` handshake. The 2026-07-28
+                // backward-compatibility guidance is explicit that this MUST NOT
+                // be keyed to one specific error code.
+                Err(_) => {
```
DaleSeo: `Err(_)` also catches 401 and 403, as well as client-side failures like `NoPreferredProtocolVersion`. Neither indicates that the peer is legacy, but `Auto` now attempts `initialize` and may hide the original actionable error.

crates/rmcp/src/service/client.rs
```diff
@@ -739,23 +739,56 @@ where
...
-                Err(ClientInitializeError::JsonRpcError(error))
-                    if error.code == crate::model::ErrorCode::METHOD_NOT_FOUND =>
-                {
+                // A recognized modern-server failure (version negotiation,
+                // missing capability, header validation, or a negotiated
+                // version mismatch): the peer speaks the modern protocol and
+                // must not trigger a legacy fallback. Surface the error.
+                Err(error) if is_modern_server_error(&error) => return Err(error),
+                // Any other outcome — `-32601`, `-32602`, `-32600`, another
+                // implementation-defined error, or no response at all — means
```
DaleSeo: `discover_startup` waits in `expect_response` until a message comes in or the transport closes. So, if a legacy server is open but silently ignores `server/discover`, it never reaches this branch.

ip2a:
Redesigned per your feedback. Replaced the `Err(_)` + modern-code carve-out
with `ClientInitializeError::indicates_legacy_server`: an exhaustive match
that falls back only on legacy signals (non-modern JSON-RPC error, closed
connection, unexpected response, non-auth transport error) and surfaces the
rest (modern rejections, auth/scope gate, client state). Added
`is_authorization_failure` for the 403 gap; `test_legacy_server_classification`
pins each variant.

Three calls I'd like your input on:

1. `ConnectionClosed` falls back. On stdio the transport is dead so
   `legacy_startup`'s send fails and masks the original error; on HTTP a new
   `initialize` POST can still work. Kept fallback, lean that way but ok to
   surface instead.
2. `ExpectedInitResponse` / `ExpectedInitResult` / `ConflictInitResponseId`
   fall back — rare, transport still alive, seemed worth a try.
3. Non-auth `TransportError` (IO) is lumped with non-JSON 400; both fall
   back. Splitting needs a per-transport downcast, didn't seem worth it.

crates/rmcp/src/service/client.rs
ip2a: Fixed in the new push — no more `Err(_)`. `indicates_legacy_server`
classifies each variant, so 401/403/missing-OAuth (new
`is_authorization_failure`) and `NoPreferredProtocolVersion` / `Cancelled`
all surface instead of falling back. Match is exhaustive, so a new variant
has to be classified explicitly.

crates/rmcp/src/service/client.rs
ip2a: Right, didn't fix that here. `expect_response` blocks until the transport
closes or a message arrives, so a server that ignores `server/discover`
never reaches the fallback. Needs a discover probe timeout, which changes
`discover_startup` / `expect_response` — I'll open a separate issue so it
stays out of this PR.

crates/rmcp/src/service/client.rs
```diff
@@ -122,6 +122,58 @@ impl ClientInitializeError {
...
+    /// variant forces the author to decide whether it indicates a legacy server.
+    pub fn indicates_legacy_server(&self) -> bool {
+        match self {
+            // The server returned an MCP-level error. It is legacy unless the
+            // error is a modern-era rejection; version negotiation is already
+            // handled inside discover_startup, so UNSUPPORTED_PROTOCOL_VERSION
+            // never reaches here.
+            Self::JsonRpcError(data) => !is_modern_rejection_code(data.code),
+            // The server closed the connection without an MCP response.
+            Self::ConnectionClosed(_) => true,
```
DaleSeo: `ConnectionClosed` means `Transport::receive()` returned `None`, but fallback immediately reuses that same transport and can mask the original error when `initialize` also fails.

crates/rmcp/src/service/client.rs
```diff
@@ -122,6 +122,58 @@ impl ClientInitializeError {
...
+            // A transport failure: surface an authorization/scope gate, since
+            // initialize would hit the same gate; otherwise the server did not
+            // produce a valid response, so attempt the fallback.
+            Self::TransportError { .. } => !self.is_authorization_failure(),
```
DaleSeo: How can `Auto` avoid downgrading a modern peer after an unrelated transport failure while still falling back for the HTTP responses described in #1040?

crates/rmcp/src/service/client.rs
```diff
@@ -122,6 +122,58 @@ impl ClientInitializeError {
...
+            // The server answered discover but no version overlaps — modern.
+            Self::NoCompatibleProtocolVersion { .. } => false,
+            // Client-side state, not a signal about the server.
+            Self::NoPreferredProtocolVersion | Self::Cancelled => false,
+            // The server answered with an unexpected shape or request id.
+            Self::ExpectedInitResponse(_)
+            | Self::ExpectedInitResult(_)
+            | Self::ConflictInitResponseId(_, _) => true,
```
DaleSeo: A mismatched response ID points to a correlation failure, not necessarily a legacy lifecycle issue. If we retry on the same transport, we might consume the actual discover response as the initialize response.

ip2a:
Rewrote this to classify inside discover_startup instead of on the error
type. It now returns DiscoverOutcome::Modern or DiscoverOutcome::Legacy,
and Auto just matches on it — no indicates_legacy_server, no downcast,
no transport types in the lifecycle layer.

The three open review comments are addressed by the rewrite:
- ConnectionClosed: doesn't enter the fallback path at all (it's not a
  JsonRpcError, so discover_startup returns Err and Auto surfaces it)
- mismatched response id: expect_response now correlates ids on error
  responses too; UncorrelatedErrorResponse is surfaced, not retried
- non-JSON HTTP responses: out of scope for this PR. Needs the transport
  to carry a structured signal, which I'd rather discuss separately

crates/rmcp/src/service/client.rs
ip2a: ConnectionClosed is not a JsonRpcError, so discover_startup returns Err and Auto surfaces it
directly — it never reaches the fallback path.

crates/rmcp/src/service/client.rs
ip2a: Fair question. This PR only falls back on a correlated JsonRpcError, so transport errors are always surfaced — there's no risk of downgrading a modern peer. The non-JSON HTTP response case (a legacy server returning
plain text) needs the transport layer to carry a structured signal, and I think that's worth a separate discussion rather than squeezing it in here.

crates/rmcp/src/service/client.rs
ip2a: Yeah, that was a real risk. expect_response now correlates ids on error responses too — a mismatched or absent id surfaces as UncorrelatedErrorResponse, which Auto treats as Err and never retries.
So initialize can't pick up a stray discover response.

crates/rmcp/src/service/client.rs
```diff
@@ -138,27 +152,46 @@ where
...
+                }
+                return Ok(result);
             }
-            // Handle JSON-RPC error responses
             ServerJsonRpcMessage::Error(error) => {
-                break Err(ClientInitializeError::JsonRpcError(error.error));
+                return Err(match &error.id {
+                    Some(id) if expected_id.matches_response_id(id) => {
+                        ClientInitializeError::JsonRpcError(error.error)
+                    }
```
DaleSeo, line 195: Response-scoped errors without an ID should remain available to the legacy classifier.

```suggestion
                    }
                    None => ClientInitializeError::JsonRpcError(error.error),
```

crates/rmcp/src/service/client.rs
```diff
@@ -74,6 +82,12 @@ pub enum ClientInitializeError {
 
     #[error("Cancelled")]
     Cancelled,
+
+    #[error("discover and legacy initialize both failed")]
+    LegacyFallbackFailed {
```
DaleSeo, line 87: In `ClientInitializeError::auth_challenge` and `ClientInitializeError::is_authorization_required`, which drives reactive OAuth, the new wrapper hides fallback 401/403 challenges and causes authorization-required fallback errors to return `false`. Recurse into the fallback error so callers keep the documented classification behavior.

crates/rmcp/src/service/client.rs
ip2a: Done — absent-id errors now go to JsonRpcError, only a mismatched Some(id) surfaces as UncorrelatedErrorResponse.

crates/rmcp/src/service/client.rs
ip2a: Fixed — both methods now recurse into the fallback phase of LegacyFallbackFailed.

DaleSeo (approved):
Thanks for addressing all my feedback, @ip2a!

PR #1141, opened 2026-08-05
fix(client): fall back on legacy JSON-RPC errors

j45856021-dev:
## Summary

- Fall back from `server/discover` to legacy `initialize` for JSON-RPC errors that do not positively identify a modern server.
- Keep authentication, transport, lifecycle configuration, and recognized modern-protocol errors actionable instead of downgrading them.
- Add a regression test for the deployed `-32000` unsupported `MCP-Protocol-Version` response, plus guards for modern capability and header errors.
- Clarify `ClientLifecycleMode::Auto` behavior in the README.

This addresses the JSON-RPC interoperability cases in #1040 and is a narrower alternative to #1133 after the review concern that `Err(_)` also catches 401/403 and client-side failures.

## Motivation and Context

The 2026-07-28 compatibility guidance says a client should fall back when a legacy server returns an error that is not a recognized modern JSON-RPC error. Some deployed legacy servers reject the modern protocol header before dispatch with an implementation-defined response such as:

```text
-32000 Bad Request: The MCP-Protocol-Version header value 2026-07-28 is not supported.
```

Current `Auto` handling only falls back for `METHOD_NOT_FOUND`, so these servers fail even though the configured legacy protocol works. This change limits the broader fallback to `JsonRpcError`; transport and authorization failures still surface unchanged.

## How Has This Been Tested?

- `cargo fmt --all --check`
- `cargo test -p rmcp --test test_client_lifecycle_modes --features client`
- `cargo test -p rmcp --test test_discover_http_client_startup --features client,reqwest,transport-streamable-http-client-reqwest,transport-streamable-http-server`
- `cargo clippy -p rmcp --test test_client_lifecycle_modes --features client -- -D warnings`

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

- [x] My code follows the repository style guidelines

### DIFF
--- a/crates/rmcp/src/service/client.rs
+++ b/crates/rmcp/src/service/client.rs
@@ -740,7 +740,7 @@ where
             match discover_result {
                 Ok(()) => {}
                 Err(ClientInitializeError::JsonRpcError(error))
-                    if error.code == crate::model::ErrorCode::METHOD_NOT_FOUND =>
+                    if !is_modern_server_json_rpc_error(&error) =>
                 {
                     let mut legacy_info = client_info;
                     if let Some(version) = legacy_version {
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/service/client.rs
```diff
@@ -740,7 +740,7 @@ where
             match discover_result {
                 Ok(()) => {}
                 Err(ClientInitializeError::JsonRpcError(error))
-                    if error.code == crate::model::ErrorCode::METHOD_NOT_FOUND =>
+                    if !is_modern_server_json_rpc_error(&error) =>
```
DaleSeo, line 743: How should this retain the HTTP status when deciding to fall back?

jamadeo:
Thanks! this identified and tested an important part of #1040. The JSON-RPC fallback behavior has since been implemented through #1133 and the expanded era-classification logic on main, which also retains HTTP status and handles version negotiation and other probe outcomes. #1211 covers the remaining non-JSON transport-level 4xx cases, so this PR is now superseded.

PR #1145, opened 2026-08-07
fix(auth): preserve issuer trailing slash during discovery

### LINKED ISSUE #1144: OAuth discovery loses trailing slash from path-based issuer before validation
## Description

Authorization server metadata discovery rejects a valid path-based issuer when the issuer ends with a trailing slash.

For example, given the authorization server issuer:

```
https://auth.example.com/tenant/
```

RFC 8414 path-insertion discovery uses:

```
https://auth.example.com/.well-known/oauth-authorization-server/tenant
```

If that endpoint returns metadata containing the original issuer:

```json
{
  "issuer": "https://auth.example.com/tenant/",
  "authorization_endpoint": "https://auth.example.com/tenant/authorize",
  "token_endpoint": "https://auth.example.com/tenant/token"
}
```

rmcp rejects it with an issuer mismatch equivalent to:

```
expected: https://auth.example.com/tenant
received: https://auth.example.com/tenant/
```

## Root cause

`generate_discovery_urls()` correctly removes the issuer path's trailing slash when constructing the RFC 8414 discovery URL. However, the original issuer is not retained.

After fetching the document, `expected_issuer_for_authorization_metadata_url()` reconstructs the expected issuer from that discovery URL. This reverse transformation is lossy: both these issuers produce the same discovery URL:

- `https://auth.example.com/tenant`
- `https://auth.example.com/tenant/`

The reconstructed expected issuer therefore becomes `https://auth.example.com/tenant`, even when the original issuer was `https://auth.example.com/tenant/`.

`issuer_identifiers_match()` intentionally treats only root-path trailing slashes as equivalent, so the path-based values fail validation. The strict comparison itself should remain unchanged because these are distinct issuer identifiers; the problem is that the expected identifier was reconstructed from a lossy representation.

## Expected behavior

Metadata whose `issuer` exactly matches the original authorization server issuer should be accepted, including a trailing slash on a non-root path.

## Suggested fix

Carry the expected issuer together with each generated discovery URL, deriving it from the original authorization server URL. Validate metadata against that retained value instead of reconstructing it from the discovery URL.

This preserves strict issuer validation while avoiding the lossy reverse transformation.

## Regression context

Issuer validation was introduced in #996 as the fix for #983. The validation is correct, but the current implementation does not preserve enough information to validate this trailing-slash case correctly.

satouriko:
## Motivation and Context

Authorization server metadata discovery currently reconstructs the expected issuer from the generated discovery URL. For a path-based issuer ending in a trailing slash, RFC 8414 path insertion removes that slash, so the reverse transformation is lossy and valid metadata is rejected with an issuer mismatch.

This change preserves the exact authorization server identifier advertised by Protected Resource Metadata and uses it for validation. Discovery URL generation and strict issuer comparison remain unchanged.

Fixes #1144

## How Has This Been Tested?

- Verified the new regression test fails on `origin/main` with `AuthorizationServerMismatch`
- `cargo test -p rmcp --lib --features auth protected_resource_metadata_preserves_non_root_issuer_trailing_slash -- --nocapture`
- `cargo test -p rmcp --lib --features auth authorization_metadata -- --nocapture`
- `cargo clippy -p rmcp --all-targets --all-features -- -D warnings`
- `cargo test -p rmcp --all-features`

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix intentionally does not normalize non-root trailing slashes. `https://auth.example.com/tenant` and `https://auth.example.com/tenant/` remain distinct issuer identifiers; validation now compares metadata against the original identifier rather than a reconstructed one.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -1439,7 +1439,7 @@ impl AuthorizationManager {
             });
         }
 
-        if let Some(metadata) = self.try_discover_oauth_server(&self.base_url).await? {
+        if let Some(metadata) = self.try_discover_oauth_server(&self.base_url, None).await? {
             return Ok(AuthorizationMetadataResolution {
                 metadata,
                 source: AuthorizationMetadataSource::AuthorizationServerMetadata,
@@ -2298,9 +2298,13 @@ impl AuthorizationManager {
     async fn try_discover_oauth_server(
         &self,
         base_url: &Url,
+        expected_issuer: Option<&str>,
     ) -> Result<Option<AuthorizationMetadata>, AuthError> {
         for discovery_url in Self::generate_discovery_urls(base_url) {
-            if let Some(metadata) = self.fetch_authorization_metadata(&discovery_url).await? {
+            if let Some(metadata) = self
+                .fetch_authorization_metadata(&discovery_url, expected_issuer)
+                .await?
+            {
                 return Ok(Some(metadata));
             }
         }
@@ -2310,6 +2314,7 @@ impl AuthorizationManager {
     async fn fetch_authorization_metadata(
         &self,
         discovery_url: &Url,
+        expected_issuer: Option<&str>,
     ) -> Result<Option<AuthorizationMetadata>, AuthError> {
         debug!("discovery url: {:?}", discovery_url);
         let response = self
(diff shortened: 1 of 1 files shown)


DaleSeo (approved):

DaleSeo (approved):
Thanks for the fix, @satouriko!

PR #1150, opened 2026-08-07
fix(model): preserve elicitation property order

### LINKED ISSUE #1109: Elicitation schema `properties` uses `BTreeMap`, discarding the server's declared field order
### Summary

`ElicitationSchema.properties` is typed as `BTreeMap<String, PrimitiveSchemaDefinition>`, so rmcp
re-orders elicitation form fields alphabetically on both deserialize and serialize. The order the
server declared on the wire is unrecoverable through the typed API.

Elicitation schemas are the one place in MCP where a JSON object's key order carries presentation
meaning: it is the order a client renders the form fields in. A server that declares

```json
{"type":"object","properties":{"firstName":{...},"lastName":{...},"email":{...}}}
```

is seen by an rmcp client as `email`, `firstName`, `lastName`.

### Affected code (3.1.0)

- `crates/rmcp/src/model/elicitation_schema.rs:1126` — `ElicitationSchema.properties: BTreeMap<…>`
- `crates/rmcp/src/model/elicitation_schema.rs:1262` — same field on `ElicitationSchemaBuilder`

Because the field is shared, this affects **all three** paths that carry an elicitation schema:

1. legacy server-initiated `elicitation/create` requests (≤ `2025-06-18`),
2. SEP-2322 task steering — `TaskPayload::InputRequired { input_requests }` (`model/task.rs:126`),
3. `2026-07-28` non-task MRTR — `InputRequiredResult.inputRequests` (`model/mrtr.rs:235`).

So it applies equally to servers and clients, and it survived the transition to MRTR.

### Why consumers cannot work around it

`serde_json`'s `preserve_order` feature does **not** help. We enable it, and Cargo's feature
unification means rmcp's own `serde_json::Value::Object` is `IndexMap`-backed in our builds and
does preserve order — but `properties` never passes through `Value`. It deserializes straight into
`BTreeMap`, so the ordering is discarded before any consumer can observe it.

The only workaround is to tap raw JSON ahead of rmcp's decode, which means decorating every
transport (stdio `AsyncRead`, Streamable HTTP, SSE, and any in-process bridge) and carrying a
request-id correlation store alongside. We implemented exactly that and are trying to delete it —
it is a lot of accidental machinery for what is a one-type change upstream.

### Suggested fix

Use an order-preserving map for `properties` — `indexmap::IndexMap<String, PrimitiveSchemaDefinition>`.
`indexmap` is already in the tree transitively via `serde_json`, and its serde impl is a drop-in for
`BTreeMap` (`IndexMap` deserializes in encounter order and serializes in insertion order).

If a hard dependency is unwelcome, gating it behind a feature that swaps the map type would also
work, though an unconditional change seems preferable given the field's semantics — nothing relies
on `properties` being sorted, and `BTreeMap`'s ordering guarantee is not meaningful for a JSON
Schema property bag.

`InputRequests = BTreeMap<String, InputRequest>` (`model/mrtr.rs:91`) and `InputResponses`
(`model/mrtr.rs:100`) have the same shape. Those keys are correlation identifiers looked up by name,
so the impact is much weaker — but if servers ever intend `inputRequests` to be presented in
declaration order, the same argument applies. Happy to scope a PR to `properties` alone.

### Relationship to #1097

Same defect class as [#1097](https://github.com/modelcontextprotocol/rust-sdk/pull/1097) (which we
hit independently, and which is why we are on 3.1.0): the typed model silently discards information
that was present on the wire, and round-tripping through rmcp is not faithful to the input. This one
loses ordering rather than fields, so it degrades rendering rather than breaking protocol flow —
lower severity, but the same underlying shape.

I'm glad to open a PR if the `IndexMap` direction is agreeable.

nightcityblade:
## What Problem This Solves

`ElicitationSchema.properties` uses a `BTreeMap`, so deserializing a server-declared form sorts its fields alphabetically. Clients could not recover the presentation order from the typed model.

## Why This Change Was Made

The public `properties` field, constructor, and builder remain `BTreeMap`-based to avoid a breaking API change. Deserialization now uses an internal `IndexMap` wire representation to capture encountered property names in a new serde-skipped `property_order: Option<Vec<String>>` field. `IndexMap` is therefore not exposed in the public API.

## User Impact

Clients can render deserialized elicitation fields in server-declared order by consulting `property_order`, while existing code using `properties` remains source-compatible within 3.x.

## Evidence

- `cargo test -p rmcp --all-features --lib test_legacy_enum_schema_roundtrip_preserves_enum_names` — passed
- `cargo +nightly fmt --all -- --check` — passed

Fixes #1109

### DIFF
--- a/crates/rmcp/src/model/elicitation_schema.rs
+++ b/crates/rmcp/src/model/elicitation_schema.rs
@@ -18,7 +18,8 @@
 
 use std::{borrow::Cow, collections::BTreeMap, marker::PhantomData};
 
-use serde::{Deserialize, Serialize};
+use indexmap::IndexMap;
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::{const_string, model::ConstString};
 
@@ -1109,9 +1110,10 @@ impl EnumSchema {
 ///     .optional_bool("newsletter", false)
 ///     .build();
 /// ```
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize)]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
-#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "schemars", schemars(!into))]
+#[serde(rename_all = "camelCase", into = "ElicitationSchemaWire")]
 #[non_exhaustive]
 pub struct ElicitationSchema {
     /// Always "object" for elicitation schemas
@@ -1125,6 +1127,11 @@ pub struct ElicitationSchema {
     /// Property definitions (must be primitive types)
     pub properties: BTreeMap<String, PrimitiveSchemaDefinition>,
 
+    /// Property names in wire order. Schemas constructed from a `BTreeMap`
+    /// use the map's sorted key order.
+    #[serde(skip)]
+    pub property_order: Option<Vec<String>>,
+
     /// List of required property names
     #[serde(skip_serializing_if = "Option::is_none")]
     pub required: Option<Vec<String>>,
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/model/elicitation_schema.rs
```diff
@@ -1136,7 +1137,7 @@ pub struct ElicitationSchema {
 
 impl ElicitationSchema {
     /// Create a new elicitation schema with the given properties
-    pub fn new(properties: BTreeMap<String, PrimitiveSchemaDefinition>) -> Self {
+    pub fn new(properties: IndexMap<String, PrimitiveSchemaDefinition>) -> Self {
```
DaleSeo: The no-`local` integration test still calls `ElicitationSchema::new(BTreeMap::new())`, so this parameter change prevents that target from compiling.

nightcityblade: Thanks for catching this. I updated the no-local integration test to construct an IndexMap and ran the exact no-local CI test command; it passes locally. The current head is f85271d.

DaleSeo:
@nightcityblade Thanks for addressing my comment. The no-local test fix looks good.

Sorry to bring this up late, but I noticed the `!` in the commit message and realized this is a breaking change. We just cut v3, so merging this would make release-plz publish v4. It would also make `indexmap` a public dependency.

Could we fix the ordering within 3.x instead? I think we could keep `properties: BTreeMap` and store the wire order as explicit metadata, like this:

```rust
pub struct ElicitationSchema {
    // ...
    pub property_order: Option<Vec<String>>,
}
```

Let me know if this would work as a fix.

nightcityblade:
Thanks for the suggestion — the metadata approach works without changing the public `properties` type.

I updated the PR so that:

- `ElicitationSchema::properties`, its constructor, and the builder remain `BTreeMap`-based.
- Deserialization captures the encountered keys in a new serde-skipped `property_order` field.
- `IndexMap` is used only inside the private wire representation, so it is not exposed in the public API.
- The branch is now a single non-breaking commit, removing the commit-lint failure caused by the earlier `!` marker.

Validation passed with `cargo test -p rmcp --all-features --lib test_legacy_enum_schema_roundtrip_preserves_enum_names` and rustfmt checking. Could you take another look?

nightcityblade:
Follow-up on the CI failures: the current head (`c240e6f`) keeps derived equality intact and initializes `property_order` consistently in the constructor and builder, so existing serialization round trips remain equal. The handwritten `Deserialize` signature also now matches the previous derived public signature.

Local validation passed:

- `cargo test -p rmcp --all-features --test test_elicitation` — 56 passed
- `cargo test -p rmcp --all-features --lib model::elicitation_schema::tests` — 27 passed
- nightly rustfmt check
- `cargo public-api` against the current base with default features and with all non-`local` features — no removed or changed items; only the new `property_order` field is added

The refreshed fork workflows are currently awaiting GitHub Actions approval.

crates/rmcp/src/model/elicitation_schema.rs
```diff
@@ -1109,7 +1110,7 @@ impl EnumSchema {
 ///     .optional_bool("newsletter", false)
 ///     .build();
 /// ```
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize)]
```
DaleSeo, line 1113: `Serialize` is still derived for `ElicitationSchema`, so `properties: BTreeMap<...>` gets serialized alphabetically. That means the round-trip and proxy/relay parts of the issue are still unresolved. When `property_order` is available, serialization should follow it and then append any properties that are not listed.

nightcityblade: Thanks, you're right — deserialization captured the order, but derived serialization still sorted the `BTreeMap`.

The current head (`297deb9`) now converts through the private wire representation on serialization, rebuilding its `IndexMap` from `property_order` (and appending any newly added properties). The regression test asserts that a deserialize/serialize round trip preserves the exact property order.

Local validation passed:

- `cargo test -p rmcp --all-features --lib model::elicitation_schema::tests` — 27 passed
- `cargo test -p rmcp --all-features --test test_elicitation` — 56 passed
- `cargo clippy -p rmcp --all-targets --all-features -- -D warnings`

DaleSeo (approved):

PR #1156, opened 2026-08-08
fix(model): deprecate ServerInfo and ClientInfo aliases

### LINKED ISSUE #1082: Deprecate misleading `ServerInfo` and `ClientInfo` aliases
## Problem

`rmcp::model` currently defines:

```rust
pub type ServerInfo = InitializeResult;
pub type ClientInfo = InitializeRequestParams;
```

These are SDK convenience aliases for the full initialization payloads. However, the 2026-07-28 protocol also uses `serverInfo` and `clientInfo` for identity metadata, where the value is `Implementation`.

Because of that, `ServerInfo` can easily be mistaken for the server identity, even though it includes the protocol version, capabilities, implementation identity, instructions, and metadata. The same issue applies to `ClientInfo`. This became especially noticeable when we aligned the discovery and result metadata in #1065 and #1078.

The naming collision is visible in several core paths:

```rust
// Subscription teardown added by #1078
let server_info = self.get_info(); // ServerInfo (InitializeResult)
let server_implementation = server_info.server_info; // Implementation

// Converting handler information into a discovery result
pub fn from_server_info(server_info: ServerInfo) {
    let ServerInfo {
        server_info, // Implementation
        // ...
    } = server_info;
}

// Building metadata for client discovery
let client_info = service.get_info(); // ClientInfo (InitializeRequestParams)
let implementation = client_info.client_info.clone(); // Implementation
```

In each case, the outer `*_info` value represents a complete initialization payload, while the inner field is only the implementation identity. The code is valid, but the repeated names obscure the distinction between the SDK aliases and the protocol metadata concepts.

## Proposed migration

1. Deprecate `ServerInfo` in favor of `InitializeResult`, and `ClientInfo` in favor of `InitializeRequestParams`. Add notes to explain the metadata name collision.

2. Update the handler trait signatures, generated macro code, documentation, examples, and internal references to use the canonical schema type names.

3. Define the existing `Default` implementations using the canonical underlying types.

4. Keep the deprecated aliases for the rest of the 3.x release series, and remove them in the next major release.

The resulting public API would look roughly like this during the deprecation period:

```rust
#[deprecated(
    note = "use InitializeResult; ServerInfo can be confused with serverInfo metadata"
)]
pub type ServerInfo = InitializeResult;

#[deprecated(
    note = "use InitializeRequestParams; ClientInfo can be confused with clientInfo metadata"
)]
pub type ClientInfo = InitializeRequestParams;

pub trait ServerHandler {
    fn get_info(&self) -> InitializeResult;
}

pub trait ClientHandler {
    fn get_info(&self) -> InitializeRequestParams;
}
```

Because these are type aliases, existing downstream implementations that return `ServerInfo` or `ClientInfo` will remain type-compatible during the deprecation period. They will only receive warnings. This proposal does not change serialization or the wire protocol.

## Alternatives

A protocol-independent ServerConfig and ClientConfig abstraction might eventually be a better way to describe local endpoint configuration, especially since stateless discovery now uses this information too. That would require a larger API design discussion, so we can consider it separately. For now, deprecating the misleading aliases is a smaller first step that directly addresses the confusion around identity metadata.

teddytennant:
## Summary

`ServerInfo` and `ClientInfo` are aliases for the full initialize payloads, but
the MCP wire names `serverInfo` / `clientInfo` mean only the `Implementation`
identity. That collision makes code like `server_info.server_info` easy to
misread.

- Mark both type aliases `#[deprecated]` pointing at `InitializeResult` and
  `InitializeRequestParams`
- Move internal call sites and `Default` impls onto the real type names
- Keep the aliases as shims so existing downstream code still compiles

## Test plan

- `cargo check -p rmcp --features server,client`
- `cargo test -p rmcp --features server,client --lib` (204 passed)

Fixes #1082

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,10 @@ Use [`ClientServiceExt::serve_with_lifecycle`](crates/rmcp/src/service/client.rs
 select another lifecycle explicitly:
 
 ```rust, ignore
-use rmcp::{ClientInfo, ClientLifecycleMode, ClientServiceExt, ProtocolVersion};
+use rmcp::{ClientLifecycleMode, ClientServiceExt, InitializeRequestParams, ProtocolVersion};
 
 // Start directly with server/discover and include client metadata on every request.
-let client = ClientInfo::default()
+let client = InitializeRequestParams::default()
     .serve_with_lifecycle(
         transport,
         ClientLifecycleMode::Discover {
@@ -113,7 +113,7 @@ let client = ClientInfo::default()
 
 // Or probe the discover lifecycle and fall back when a legacy server reports
 // that server/discover is not implemented or does not respond within 10 seconds.
-let client = ClientInfo::default()
+let client = InitializeRequestParams::default()
     .serve_with_lifecycle(
         transport,
         ClientLifecycleMode::Auto {
@@ -369,8 +369,8 @@ use serde_json::json;
 struct MyServer;
 
 impl ServerHandler for MyServer {
-    fn get_info(&self) -> ServerInfo {
-        ServerInfo::new(
+    fn get_info(&self) -> InitializeResult {
+        InitializeResult::new(
             ServerCapabilities::builder()
                 .enable_resources()
                 .build(),
@@ -582,8 +582,8 @@ impl MyServer {
 
 #[prompt_handler]
 impl ServerHandler for MyServer {
-    fn get_info(&self) -> ServerInfo {
-        ServerInfo::new(ServerCapabilities::builder().enable_prompts().build())
+    fn get_info(&self) -> InitializeResult {
+        InitializeResult::new(ServerCapabilities::builder().enable_prompts().build())
     }
 }
 ```
@@ -934,8 +934,8 @@ Enable the logging capability, handle level changes from the client, and send lo
 use rmcp::{ServerHandler, model::*, service::RequestContext};
 
 impl ServerHandler for MyServer {
-    fn get_info(&self) -> ServerInfo {
-        ServerInfo::new(
+    fn get_info(&self) -> InitializeResult {
+        InitializeResult::new(
             ServerCapabilities::builder()
                 .enable_logging()
                 .build(),
@@ -1007,8 +1007,8 @@ Enable the completions capability and implement the `complete()` handler. Use `r
 use rmcp::{ErrorData as McpError, ServerHandler, model::*, service::RequestContext, RoleServer};
 
 impl ServerHandler for MyServer {
-    fn get_info(&self) -> ServerInfo {
-        ServerInfo::new(
+    fn get_info(&self) -> InitializeResult {
+        InitializeResult::new(
             ServerCapabilities::builder()
                 .enable_completions()
                 .enable_prompts()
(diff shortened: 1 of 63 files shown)


teddytennant:
Clippy is fixed. The deprecation was firing from the integration tests and from the code the `tool_handler` macro generates, so I finished the migration: the macro output, all tests, examples and docs now use `InitializeResult` and `InitializeRequestParams`. `cargo clippy --all-targets --all-features -- -D warnings` and the test suite are both clean locally.

The Public API Check needs a call from you. It flags six changed items, all of them `get_info` going from `ServerInfo` to `InitializeResult`. Since `ServerInfo` is just `pub type ServerInfo = InitializeResult`, those are the same type and nothing downstream breaks. The tool compares the written names, so it can't tell the difference.

Updating the trait signatures is step 2 of #1082, so I'd rather not revert it just to make the check quiet. Happy to do whatever you prefer here.

teddytennant:
Went ahead and fixed the Public API Check instead of leaving it hanging on you.

The items it flagged are back on the aliases: `ServerHandler::get_info`, `ClientHandler::get_info`, `DiscoverResult::from_server_info` and `impl ClientHandler for ClientInfo`. rustdoc records a type alias by the name it was written with, so respelling those as `InitializeResult` / `InitializeRequestParams` reads as a changed public item even though it resolves to the same type. The check is doing its job; my change was just invisible to it.

Everything else stays as it was: the `#[deprecated]` attributes, plus the `tool_handler` macro output, tests, examples and docs on the canonical names, so new code still gets pointed the right way. Moving the published signatures over is a rendered-API change with no compile impact, so I'd rather it ride along with the next major — happy to open a follow-up issue to track it.

Verified locally with the same commands the job runs, against `02c62ae`, for both feature sets:

```
cargo public-api --package rmcp -ss diff --deny changed --deny removed --force 02c62ae..HEAD
cargo public-api --package rmcp --features "<all except local>" -ss diff --deny changed --deny removed --force 02c62ae..HEAD
```

Both come back with no removed and no changed items. The only thing left in the diff is the two added `Default` impls, which the minor deny flags allow:

```
Added items to the public API
=============================
+impl core::default::Default for rmcp::model::InitializeRequestParams
+pub fn rmcp::model::InitializeRequestParams::default() -> Self
+impl core::default::Default for rmcp::model::InitializeResult
+pub fn rmcp::model::InitializeResult::default() -> Self
```

`cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --all-features` (771 tests) are both clean.

The new workflow runs are sitting in `action_required` and need an approval before they'll start.

DaleSeo:
Thanks for putting this together, @teddytennant. I haven't aligned with the other maintainers on the best direction for this naming issue, so I'll hold off on reviewing this for now and come back next week.

teddytennant:
Any luck aligning on the naming? Happy to rebase this whenever there's a direction, I've been holding off since the diff touches type names across 62 files and would just conflict again while it waits.

teddytennant:
Rebased on main. The type-name conflicts in the new version tests are resolved.

teddytennant:
Rebased again. Conflicts were in the new negotiate_initialize tests.

DaleSeo:
@teddytennant Sorry for the delay with this PR. The maintainers are still discussing better aliases... 😭

teddytennant:
Clippy and the no-local tests were still hitting `ClientInfo` in two tests that landed on main after the rebase. They're on `InitializeRequestParams` now.

DaleSeo (approved):

DaleSeo:
Thanks for your patience, @teddytennant! We decided to merge this PR and follow up with new aliases.

PR #1165, opened 2026-08-11
fix: preserve SSE resume cache and correct event replay

freedom-winds:
## Summary

Fix two issues in the Streamable HTTP server's SSE resume behavior:

- The event identified by `Last-Event-ID` was replayed again.
- Request-scoped SSE caches were deleted immediately after the final response, preventing clients from recovering responses after a disconnect.

## Changes

- Adjust the SSE resume offset to replay only events after `Last-Event-ID`.
- Reject event IDs older than the retained cache window.
- Mark completed request channels as complete while retaining their cached events.
- Allow `completed_cache_ttl` to evict completed request caches.
- Add regression coverage for retaining completed request channel state.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/session/local.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/session/local.rs
@@ -355,6 +355,17 @@ impl CachedTx {
         }
         Ok(())
     }
+
+    async fn replay_cached(&mut self) -> Result<(), SessionError> {
+        for message in &self.cache {
+            let send_result = self.tx.send(message.clone()).await;
+            if send_result.is_err() {
+                let event_id: EventId = message.event_id.as_deref().unwrap_or_default().parse()?;
+                return Err(SessionError::ChannelClosed(Some(event_id.index as u64)));
+            }
+        }
+        Ok(())
+    }
 }
 
 struct HttpRequestWise {
@@ -705,17 +716,16 @@ impl LocalSessionWorker {
     async fn establish_common_channel(
         &mut self,
     ) -> Result<StreamableHttpMessageReceiver, SessionError> {
-        let last_event_index = self.event_store.is_none().then_some(0);
-        self.resume_or_shadow_common(last_event_index).await
+        self.resume_or_shadow_common(None).await
     }
 
     /// Resume the common channel, or create a shadow stream if the primary is
     /// still active.
     ///
     /// When the primary common channel is dead (receiver dropped), replace it
-    /// so this stream becomes the new primary notification channel. Cached
-    /// messages are replayed from `last_event_index` so the client receives
-    /// any events it missed (including server-initiated requests).
+    /// so this stream becomes the new primary notification channel. Resume
+    /// requests replay events after `last_event_index`; fresh GETs without a
+    /// Last-Event-ID replay the locally retained cache.
     ///
     /// When the primary is still active, create a "shadow" stream — an idle SSE
     /// connection kept alive by keep-alive pings. This prevents multiple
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -638,8 +645,14 @@ impl LocalSessionWorker {
             OutboundChannel::RequestWise { id, close } => {
                 if let Some(request_wise) = self.tx_router.get_mut(&id) {
                     request_wise.tx.send(message).await?;
-                    if close && let Some(channel) = self.tx_router.remove(&id) {
-                        for resource in channel.resources {
+                    if close {
+                        let resources: Vec<_> = request_wise.resources.drain().collect();
+                        request_wise.completed_at = Some(Instant::now());
```
DaleSeo, line 660: The worker checks for eviction before it waits, so it checks this newly completed entry right away. That means the entry can outlive its TTL. If `Resume` is the next event, it gets served before the next eviction pass.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -335,7 +335,14 @@ impl CachedTx {
...
             .parse::<EventId>()?;
-        let sync_index = index.saturating_sub(front_event_id.index);
+        if index < front_event_id.index {
+            return Err(SessionError::InvalidEventId);
+        }
+        // Last-Event-ID is the last event the client received, so resume only
+        // with events that follow it.
+        let sync_index = index
+            .saturating_sub(front_event_id.index)
+            .saturating_add(1);
```
DaleSeo: If the retained cache starts at event  `N` and the client last received `N - 1`, every missed event is still available, but this check rejects the resume.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
freedom-winds: Addressed — resume() now evicts expired completed channels before looking up the cache, and I added a regression test for an expired completed cache.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
freedom-winds: Addressed — resume validation now checks the next event after Last-Event-ID, so N-1 correctly resumes from a cache starting at N. Added a regression test for this boundary case.

crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -335,7 +335,13 @@ impl CachedTx {
             .as_deref()
             .unwrap_or_default()
             .parse::<EventId>()?;
-        let sync_index = index.saturating_sub(front_event_id.index);
+        // Last-Event-ID is the last event the client received. The resume is
+        // valid as long as the next event has not already fallen out of cache.
+        let next_index = index.saturating_add(1);
```
DaleSeo, line 340: A standalone GET emits a synthetic priming event with ID `0`. But `CachedTx::new_common` also assigns `0` to the first cached common event. How should a client that only received the priming event resume without `sync(0)` skipping that unseen notification?

crates/rmcp/src/transport/streamable_http_server/session/local.rs
```diff
@@ -1371,4 +1398,91 @@ mod sep2260_routing_tests {
...
+        worker
+            .handle_server_message(ServerJsonRpcMessage::error(
+                crate::model::ErrorData::internal_error("failed", None),
+                Some(request_id),
+            ))
+            .await
+            .unwrap();
+
+        assert!(worker.tx_router[&http_request_id].completed_at.is_some());
+        assert!(worker.tx_router[&http_request_id].resources.is_empty());
```
DaleSeo, line 1422: Could this regression exercise `resume` and assert that the cached final response is delivered?

PR #1166, opened 2026-08-11
fix: harden signing key handling

### LINKED ISSUE #1161: request-state and JWT signing key material lacks minimum-length enforcement and zeroization
## Vulnerability
Two related hardening gaps around signing-key handling for security-sensitive paths:

1. `RequestStateCodec::new` (`crates/rmcp/src/model/request_state.rs:177`) accepts `impl Into<Vec<u8>>` with no length check. The doc comment recommends "a high-entropy key of at least 32 bytes" but nothing enforces it — a caller can construct a codec with a key as short as `b"key"` (see the crate's own test at `request_state.rs:395`, `:438`, `:449`, `:466`, which use 3-byte keys) and it compiles and works normally in production code, not just tests.
2. Key material is never zeroized on drop. `RequestStateCodec.key: Box<[u8]>` (`model/request_state.rs:164`) and the `signing_key: Vec<u8>` used for `private_key_jwt` client assertions (`transport/auth.rs:931`, passed into `EncodingKey::from_rsa_pem`/`from_ec_pem` at `auth.rs:3205-3206`) are plain heap buffers with no `Drop` impl clearing them, and the crate does not depend on `zeroize`.

## Severity
Low (P3) — defense-in-depth. Neither issue is directly triggerable by a network peer; both require a downstream integrator to misuse the API (pick a weak key, or rely on the process not being memory-inspected/core-dumped) rather than an attacker supplying malicious input.

## Location
- `crates/rmcp/src/model/request_state.rs:177` (`RequestStateCodec::new`, no key-length validation)
- `crates/rmcp/src/model/request_state.rs:164` (`key: Box<[u8]>` field, no zeroize)
- `crates/rmcp/src/transport/auth.rs:931,3205-3206` (`signing_key: Vec<u8>` for `private_key_jwt`, no zeroize)

## Attack Scenario
A downstream server operator constructs `RequestStateCodec::new(b"short")` (nothing in the API prevents it). The resulting HMAC-SHA256 key has far less entropy than the documented 32-byte minimum, making offline brute-force/dictionary recovery of the key from observed `seal()` outputs meaningfully more tractable than intended, which would let an attacker forge `requestState` blobs (SEP-2322) and bypass the integrity guarantee the codec exists to provide. Separately, because neither the HMAC key nor the JWT signing key is zeroized on drop, either can linger in freed heap memory and surface in a core dump, swap file, or memory-disclosure bug in a way a `zeroize`-backed type would prevent.

## Remediation
- Have `RequestStateCodec::new` return a `Result` (or otherwise reject) keys shorter than a defined minimum (e.g. 16 or 32 bytes), instead of only documenting the requirement.
- Wrap `RequestStateCodec.key` and the `private_key_jwt` `signing_key` in a `zeroize::Zeroizing<Vec<u8>>` (or implement `ZeroizeOnDrop`) so key material is cleared when dropped.

DaleSeo:
Fixes #1161.

## Motivation and Context

`RequestStateCodec` protects opaque MRTR requestState values with an HMAC signing key. However, its existing constructor accepted keys of any length, even though its documentation recommended at least 32 bytes of high-entropy key material. This made it easy for integrations to accidentally configure weak key material.

The codec also kept the key bytes in memory without zeroizing them when dropped. In addition, the derived Debug implementation for `ClientCredentialsConfig` could expose OAuth client secrets and private signing keys.

This change makes the secure path explicit. The `try_new` constructor now requires a key of at least 32 bytes. The legacy new constructor is deprecated as of version 3.1.4, while `new_unchecked` remains available for keys that are validated separately. Stored request-state keys are now zeroized, and credential Debug output redacts secret fields.

## How Has This Been Tested?

Added tests

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/conformance/src/bin/server.rs
+++ b/conformance/src/bin/server.rs
@@ -52,6 +52,10 @@ fn custom_header_tool() -> Tool {
 /// Signing key for SEP-2322 `requestState` sealing. A fixed key is fine for a
 /// conformance harness; real servers must load a secret out of clients' reach.
 const REQUEST_STATE_KEY: &[u8] = b"rust-sdk-conformance-request-state-key!!";
+const _: () = assert!(
+    REQUEST_STATE_KEY.len() >= RequestStateCodec::MIN_KEY_LENGTH,
+    "REQUEST_STATE_KEY is shorter than RequestStateCodec::MIN_KEY_LENGTH",
+);
 
 #[derive(Clone)]
 struct ConformanceServer {
@@ -70,8 +74,7 @@ impl ConformanceServer {
             subscriptions: Arc::new(Mutex::new(HashMap::new())),
             next_subscription: Arc::new(AtomicU64::new(0)),
             log_level: Arc::new(Mutex::new(LoggingLevel::Debug)),
-            request_state_codec: RequestStateCodec::try_new(REQUEST_STATE_KEY)
-                .expect("conformance request-state key meets the minimum length"),
+            request_state_codec: RequestStateCodec::new_unchecked(REQUEST_STATE_KEY),
             tasks: TaskManager::new(),
         }
     }
--- a/conformance/src/bin/server.rs
+++ b/conformance/src/bin/server.rs
@@ -70,7 +70,8 @@ impl ConformanceServer {
             subscriptions: Arc::new(Mutex::new(HashMap::new())),
             next_subscription: Arc::new(AtomicU64::new(0)),
             log_level: Arc::new(Mutex::new(LoggingLevel::Debug)),
-            request_state_codec: RequestStateCodec::new(REQUEST_STATE_KEY),
+            request_state_codec: RequestStateCodec::try_new(REQUEST_STATE_KEY)
+                .expect("conformance request-state key meets the minimum length"),
             tasks: TaskManager::new(),
         }
     }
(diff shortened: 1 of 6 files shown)


conformance/src/bin/server.rs
```diff
@@ -70,7 +70,8 @@ impl ConformanceServer {
             subscriptions: Arc::new(Mutex::new(HashMap::new())),
             next_subscription: Arc::new(AtomicU64::new(0)),
             log_level: Arc::new(Mutex::new(LoggingLevel::Debug)),
-            request_state_codec: RequestStateCodec::new(REQUEST_STATE_KEY),
+            request_state_codec: RequestStateCodec::try_new(REQUEST_STATE_KEY)
+                .expect("conformance request-state key meets the minimum length"),
```
alexhancock: We don't  need the runtime check here for `.expect` when it's a constant defined in the file, right. Can't we just check it at compile time or leave the check out given we know where it comes from?

alexhancock (dismissed):

conformance/src/bin/server.rs
DaleSeo: Makes sense, @alexhancock! I changed the fixed-key call sites to use a const assertion and `new_unchecked`.

alexhancock (approved):

PR #1167, opened 2026-08-12
fix(auth): retain state until issuer validation

stevenlee-oai:
## Summary

- validate the RFC 9207 authorization response issuer before deleting the stored authorization state
- keep valid callbacks one-time by deleting state immediately after issuer validation succeeds
- add regression coverage for missing and mismatched required `iss`

## Why

The returned OAuth `state` is only the lookup key. An unknown state loads no `StoredAuthorizationState` and deletes nothing. But an authorization server that received the authorization request can send a callback with the correct state and a forged or missing required `iss`. Previously RMCP loaded and deleted the real stored state before issuer validation rejected that callback. That consumed the PKCE verifier, so the later legitimate callback could not complete.

The new order is load state, validate `iss`, delete state, then exchange the code. Invalid issuer callbacks do not reach the token endpoint or consume state; a valid callback still consumes the one-time state before token exchange.

## Tests

- `cargo fmt --check`
- `cargo test -p rmcp --lib --features auth invalid_issuer_does_not_consume_authorization_state`
- `cargo test -p rmcp --lib --features auth`

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -2018,10 +2018,12 @@ impl AuthorizationManager {
                 AuthError::InternalError("Authorization state not found".to_string())
             })?;
 
-        // Delete state after retrieval (one-time use)
-        self.state_store.delete(csrf_token).await?;
-
         Self::validate_authorization_response_issuer(&stored_state, received_issuer)?;
+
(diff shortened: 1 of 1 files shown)


DaleSeo (approved):
Thanks, @stevenlee-oai!

PR #1176, opened 2026-08-14
fix(rmcp): preserve elicitation requestedSchema $schema dialect

### LINKED ISSUE #1168: Elicitation `requestedSchema` silently drops `$schema` and unknown property keywords
## Describe the bug

rmcp deserializes an elicitation `requestedSchema` into typed structs and silently discards fields those structs do not model.

I found two related cases:

1. The top-level `$schema` field, which is present in the MCP protocol schemas starting with 2025-11-25:

```ts
requestedSchema: {
  $schema?: string;
  type: "object";
  properties: { [key: string]: PrimitiveSchemaDefinition };
  required?: string[];
};
```

See [`[ElicitRequestFormParams](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-11-25/schema.ts)`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-11-25/schema.ts).

This is particularly significant when a schema explicitly declares a non-default dialect. [[SEP-1613](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1613)](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1613) allows embedded schemas to declare an alternative dialect through `$schema`, and [[rust-sdk issue #525](https://github.com/modelcontextprotocol/rust-sdk/issues/525)](https://github.com/modelcontextprotocol/rust-sdk/issues/525) explicitly includes elicitation `requestedSchema` in that requirement.

[[PR #549](https://github.com/modelcontextprotocol/rust-sdk/pull/549)](https://github.com/modelcontextprotocol/rust-sdk/pull/549) changed generated tool schemas to JSON Schema 2020-12, but it did not add `$schema` support to `ElicitationSchema`. It looks like that part of issue #525 was missed.

2. Unknown keywords inside individual property schemas.

`PrimitiveSchemaDefinition` is a fixed, untagged enum of typed schema variants. A property containing an additional keyword is accepted, but that keyword is silently discarded and cannot be recovered through the decoded model.

These cases are slightly different: `$schema` is a missing protocol field that can be modeled directly, while preserving unrecognized property keywords would require a catch-all or another lossless representation.

## To reproduce

Using rmcp 3.1.2, deserialize this request into `rmcp::model::InputRequest` and serialize it again:

```rust
let decoded: rmcp::model::InputRequest =
    serde_json::from_value(wire.clone()).unwrap();

println!("IN : {}", serde_json::to_string(&wire).unwrap());
println!("OUT: {}", serde_json::to_string(&decoded).unwrap());
```

Input:

```json
{
  "method": "elicitation/create",
  "params": {
    "message": "attach the report",
    "mode": "form",
    "requestedSchema": {
      "$schema": "https://json-schema.org/draft/2020-12/schema",
      "properties": {
        "report": {
          "description": "the quarterly report",
          "format": "uri",
          "type": "string",
          "x-mcp-file": {
            "accept": ["application/pdf"],
            "transferModes": ["upload"]
          }
        }
      },
      "required": ["report"],
      "type": "object"
    }
  }
}
```

Output:

```json
{
  "method": "elicitation/create",
  "params": {
    "message": "attach the report",
    "mode": "form",
    "requestedSchema": {
      "properties": {
        "report": {
          "description": "the quarterly report",
          "format": "uri",
          "type": "string"
        }
      },
      "required": ["report"],
      "type": "object"
    }
  }
}
```

`description`, `format`, `type`, and `required` survive. `$schema` and `x-mcp-file` do not.

I originally reproduced this on 3.1.0 and have confirmed the same behavior on 3.1.2. The current `main` branch still lacks fields for `$schema` and unrecognized property keywords.

## Expected behavior

At minimum, the spec-defined `$schema` field should remain available after deserialization.

Separately, preserving unmodeled property keywords in an `extra` map or another lossless raw representation would make inbound decoding forward-compatible with JSON Schema annotations and newer MCP extension keywords.

For example, draft [[SEP-2631](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2631)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2631) carries forward the proposed `x-mcp-file` inline schema annotation. An older rmcp version does not need to understand the keyword’s semantics, but silently discarding it prevents consumers that do understand the extension from accessing it without intercepting the raw JSON before rmcp decodes the request.

The current MCP specification does not prescribe a particular SDK representation for unknown keywords, so I see this as a forward-compatibility improvement rather than the same normative omission as `$schema`.

## Affected paths

The same `ElicitationSchema` and `InputRequest` models are used by:

- Direct `elicitation/create` requests, including legacy form requests and the current form mode
- Task input steering through `TaskPayload::InputRequired`
- MRTR through `InputRequiredResult.inputRequests`

I confirmed the loss through all three paths on rmcp 3.1.2.

Inbound clients and gateways lose the fields during decoding. Outbound servers also cannot express `$schema` or extension keywords through the typed `ElicitationSchema` API.

Tool schemas do not have this problem. `Tool::input_schema` and `output_schema` retain their underlying `JsonObject`, so JSON Schema keywords survive a round trip. That behavior is discussed in [[#874](https://github.com/modelcontextprotocol/rust-sdk/issues/874)](https://github.com/modelcontextprotocol/rust-sdk/issues/874) and exercised by the keyword-preservation work in [[#1003](https://github.com/modelcontextprotocol/rust-sdk/issues/1003)](https://github.com/modelcontextprotocol/rust-sdk/issues/1003) and [[#1018](https://github.com/modelcontextprotocol/rust-sdk/pull/1018)](https://github.com/modelcontextprotocol/rust-sdk/pull/1018).

## Practical impact

I’m building a gateway that brokers MCP file transfer while keeping file bytes out of model context. When an upstream server pauses for elicitation, the gateway relays that request downstream. On retry, it needs to know which response fields may contain files.

[[SEP-2356](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2356) originally proposed the `x-mcp-file` declaration. It was closed in favor of draft [[SEP-2631](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2631)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2631), which carries that declaration forward and builds its transfer flow around it. rmcp drops the declaration before the gateway can inspect it.

I cannot safely rely on a schema copy echoed by the client during retry. The [[MRTR requirements](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr)](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr) require servers to treat `requestState` as attacker-controlled and protect its integrity when it affects authorization, resource access, or business logic.

latent-9:
## What

`ElicitationSchema` (de)serializes through an internal `ElicitationSchemaWire`
bridge. The bridge had no field for the top-level `$schema` keyword, so a
dialect declared on an incoming `requestedSchema` was accepted but silently
dropped when the schema was re-serialized.

The 2025-11-25 protocol revision allows a `requestedSchema` to carry a
`$schema` dialect identifier, and `ElicitationSchema::from_type` already emits
one (schemars draft-07), so the value was being lost on any round-trip.

## Change

- Add a `schema: Option<Cow<'static, str>>` field (serde `rename = "$schema"`)
  to both `ElicitationSchema` and the `ElicitationSchemaWire` bridge.
- Thread it through the two `From` conversions.
- Add a `with_schema` setter, matching the existing `with_title` and
  `with_description` builders. The struct is `#[non_exhaustive]`, so an
  external caller needs a setter to populate the field.

The field uses `skip_serializing_if = "Option::is_none"`, so a schema without a
dialect serializes exactly as before. Adding a field to a `#[non_exhaustive]`
struct is semver compatible.

## Tests

Three unit tests in `elicitation_schema.rs`:

- a `$schema` value survives a deserialize then serialize round-trip,
- a schema without a dialect emits no `$schema` key,
- the `with_schema` setter serializes the dialect.

## Scope

This handles the `$schema` half of #1168. The other half (preserving unknown
property-level keywords) needs a data-model decision on where arbitrary
keywords should live, so I left it out of this change and kept the issue open
for it.

Refs #1168

### DIFF
--- a/crates/rmcp/src/model/elicitation_schema.rs
+++ b/crates/rmcp/src/model/elicitation_schema.rs
@@ -1116,6 +1116,14 @@ impl EnumSchema {
 #[serde(rename_all = "camelCase", into = "ElicitationSchemaWire")]
 #[non_exhaustive]
 pub struct ElicitationSchema {
+    /// Optional JSON Schema dialect identifier (the `$schema` keyword).
+    ///
+    /// The 2025-11-25 protocol revision allows a `requestedSchema` to declare its
+    /// dialect. It is preserved verbatim so a declared dialect survives a
+    /// decode/re-encode round-trip instead of being silently dropped.
+    #[serde(rename = "$schema", skip_serializing_if = "Option::is_none")]
+    pub schema: Option<Cow<'static, str>>,
+
     /// Always "object" for elicitation schemas
     #[serde(rename = "type")]
     pub type_: ObjectTypeConst,
@@ -1144,6 +1152,8 @@ pub struct ElicitationSchema {
 #[derive(Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct ElicitationSchemaWire {
+    #[serde(rename = "$schema", skip_serializing_if = "Option::is_none")]
+    schema: Option<Cow<'static, str>>,
     #[serde(rename = "type")]
     type_: ObjectTypeConst,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -1158,6 +1168,7 @@ struct ElicitationSchemaWire {
 impl From<ElicitationSchemaWire> for ElicitationSchema {
     fn from(schema: ElicitationSchemaWire) -> Self {
         Self {
+            schema: schema.schema,
             type_: schema.type_,
             title: schema.title,
             property_order: Some(schema.properties.keys().cloned().collect()),
@@ -1183,6 +1194,7 @@ impl From<ElicitationSchema> for ElicitationSchemaWire {
         properties.extend(remaining);
 
         Self {
+            schema: schema.schema,
             type_: schema.type_,
             title: schema.title,
             properties,
@@ -1206,6 +1218,7 @@ impl ElicitationSchema {
     pub fn new(properties: BTreeMap<String, PrimitiveSchemaDefinition>) -> Self {
         let property_order = Some(properties.keys().cloned().collect());
         Self {
+            schema: None,
             type_: ObjectTypeConst,
             title: None,
             properties,
@@ -1298,6 +1311,12 @@ impl ElicitationSchema {
     /// Set the description
     pub fn with_description(mut self, description: impl Into<Cow<'static, str>>) -> Self {
         self.description = Some(description.into());
+        self
+    }
+
+    /// Set the JSON Schema dialect identifier (the `$schema` keyword).
+    pub fn with_schema(mut self, schema: impl Into<Cow<'static, str>>) -> Self {
+        self.schema = Some(schema.into());
         self
     }
 
@@ -1703,6 +1722,7 @@ impl ElicitationSchemaBuilder {
 
         let property_order = Some(self.properties.keys().cloned().collect());
         Ok(ElicitationSchema {
+            schema: None,
             type_: ObjectTypeConst,
             title: self.title,
             properties: self.properties,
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/model/elicitation_schema.rs
```diff
@@ -1116,6 +1116,14 @@ impl EnumSchema {
 #[serde(rename_all = "camelCase", into = "ElicitationSchemaWire")]
 #[non_exhaustive]
 pub struct ElicitationSchema {
+    /// Optional JSON Schema dialect identifier (the `$schema` keyword).
+    ///
+    /// The 2025-11-25 protocol revision allows a `requestedSchema` to declare its
+    /// dialect. It is preserved verbatim so a declared dialect survives a
+    /// decode/re-encode round-trip instead of being silently dropped.
+    #[serde(rename = "$schema", skip_serializing_if = "Option::is_none")]
```
DaleSeo: The MCP references define `$schema` as an optional string, but this generates `"type": ["string", "null"]`.
```suggestion
    #[serde(rename = "$schema", default, skip_serializing_if = "Option::is_none")]
    #[cfg_attr(feature = "schemars", schemars(with = "String"))]
```

DaleSeo (dismissed):
Thanks for the fix, @latent-9. I have a small suggestion to make the generated schema more precise.

DaleSeo (approved):

PR #1177, opened 2026-08-14
fix(auth): ignore query parameters in protected resource matching

stevenlee-oai:
## Why

The [AWS MCP Server](https://docs.aws.amazon.com/agent-toolkit/latest/userguide/getting-started-aws-mcp-server.html) exposes both public documentation tools and authenticated AWS API tools. Requiring OAuth during every initialization would block public tools, but some clients only start OAuth after an initialization-time `401`. AWS therefore uses `?oauth=initialize` to request an early `WWW-Authenticate` challenge without changing the token audience:

```text
MCP endpoint: https://mcp.example.com/mcp?oauth=initialize
Metadata resource: https://mcp.example.com/mcp
```

MCP permits [authorization-dependent tool visibility](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#capabilities) and documents [401 discovery](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/authorization-server-discovery#protected-resource-metadata-discovery-requirements) and [step-up authorization](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization#step-up-authorization-flow), but does not prescribe first-time OAuth negotiation for endpoints mixing public and protected tools.

## Standards and compatibility

[RFC 8707 §2](https://datatracker.ietf.org/doc/html/rfc8707#section-2) says resource URIs SHOULD NOT contain queries, but explicitly permits them; it prohibits fragments and does not define query equivalence. [RFC 9728 §3.3](https://datatracker.ietf.org/doc/html/rfc9728#section-3.3) requires exact metadata matching, which literally rejects this AWS pattern. The [TypeScript SDK](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core-internal/src/shared/authUtils.ts#L17-L56) resolves that interoperability tension by comparing origin and path while ignoring queries.

This PR adopts the same policy: preserve scheme, host, port, path-boundary, and fragment checks; ignore queries for equal or permitted parent paths; and use the metadata's canonical resource in OAuth requests. Consequently, queries no longer distinguish same-origin, path-compatible resources.

## Validation

```bash
cargo +1.97.1 test -p rmcp --features auth,transport-io
cargo +nightly-2025-09-01 fmt --all -- --check
git diff --check
```

554 tests pass, including 369 library tests. Coverage includes the AWS-style OAuth flow, absent/differing queries, parent paths, deceptive path prefixes, and cross-origin resources; the new regression fails before the fix.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -2559,14 +2559,12 @@ impl AuthorizationManager {
         let expected_path = expected.path();
         let actual_path = actual.path();
 
-        // URL query part supported, even if it is discouraged in RFC 8707
-        if expected_path == actual_path && expected.query() == actual.query() {
+        // Query parameters may carry transport hints rather than resource identity.
+        if expected_path == actual_path {
             return true;
         }
 
         expected_path.starts_with(actual_path)
-            && expected.query().is_none()
-            && actual.query().is_none()
             && (actual_path.ends_with('/')
                 || expected_path.as_bytes().get(actual_path.len()) == Some(&b'/'))
     }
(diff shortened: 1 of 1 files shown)


mrorigo:
LGTM. Strong +1 from the client/proxy side ([Warmplane](https://github.com/Warmplane/warmplane)). 
We frequently see upstream servers and reverse proxies use query parameters for transport negotiation, session routing, or initialization hints (`?oauth=initialize`, `?transport=sse`). Strict query matching in RFC 9728 validation caused unnecessary rejections when servers declared base canonical resources in metadata.

Two details done right here:
1. **TS SDK Parity:** Eliminates subtle cross-language divergence between `@modelcontextprotocol/typescript-sdk` and `rmcp`.
2. **Canonical Resource Audience:** Using the metadata's canonical URI (rather than the transport URL with queries) for the RFC 8707 `resource` parameter prevents ephemeral transport queries from leaking into token audience claims.

Clean fix and solid test coverage around path boundary edge cases.

alexhancock (approved):

PR #1186, opened 2026-08-18
fix: allow concurrent streamable http requests

nickcoai:
## Summary

Allow ordinary streamable http POSTs to run concurrently. Previously, one slow `application/json` response blocked unrelated requests.

`StreamableHttpClientTransportConfig::max_concurrent_requests` defaults to 16. Set it to `1` to keep ordinary POSTs serial; `0` is treated as `1`. Open sse response streams do not count against this limit. Initialization and protocol-version changes remain ordered with ordinary POSTs.

Response cleanup matches ids against all pending requests, not just open response streams. Exact ids take precedence over the existing numeric-string fallback.

## Cancellation and recovery

Cancellation and replies to server requests use a separate bounded queue with one extra POST slot. Configure their timeout with `control_request_timeout`, which defaults to five seconds. The timeout starts when the POST starts, excluding time in the queue. Queued and active ordinary POSTs can be cancelled without waiting for a free slot.

For an open legacy response stream, cancellation stops reading but keeps the stream alive until the cancellation send finishes or is dropped. This lets custom http adapters handle cancellation using their existing stream state. Ordinary legacy cancellation notifications still reach the adapter; modern http cancellation remains local.

When recovery is enabled, ordinary POSTs that report `SessionExpired` in the same session share one recovery attempt. Wait up to `session_recovery_timeout` for old POSTs, then stop any that remain. Those interrupted POSTs return `SessionRecoveryTimeout` and are not retried: the server may have processed them. The replacement handshake has a separate timeout of the same length; both default to five seconds.

Retry only ordinary POSTs that returned `SessionExpired`, at most once each. Control POSTs are not retried. A request that already completed on the common response stream is not retried. Old queued controls cannot change replacement-session state, and a cancelled or dropped version-change request does not block later requests.

The shared worker's control queue is opt-in; server scheduling is unchanged. The generation is captured when a send future is created; it does not identify the session that started an inbound handler. Worker ownership is covered by the refactor in #1205. Late-handler session tracking is covered by #1206. Both drafts are based on this PR. Callers still decide which tools may run together and which need approval. Oauth refresh coordination remains a separate follow-up.

## Validation

Passed locally:

- Client-only concurrency target: 20 tests, including configurable cancellation/reply timeouts, distinct numeric/string response ids, early responses, recovery, cancellation, and custom-adapter compatibility.
- Worker unit tests: 3 tests for abandoned cancellation with a separate registration owner, request-id reuse, and exact cancellation-id matching.
- `cargo +1.96 test --offline --all-features`
- Selected transport targets with the complete non-local feature set from `justfile`: 36 tests covering concurrency, subscriptions, expired sessions, disconnects, connection reuse, and shutdown.
- `cargo +1.96 clippy --offline --all-targets --all-features -- -D warnings`
- `git diff --check`

Two downstream custom-adapter integration tests also passed without adapter source changes.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -623,16 +623,13 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
     fn clear_stream_response_pending(
         pending_stream_response_ids: &mut HashSet<RequestId>,
         message: &ServerJsonRpcMessage,
-    ) {
-        let Some(response_id) = Self::server_response_id(message) else {
-            return;
-        };
-        if pending_stream_response_ids.remove(response_id) {
-            return;
+    ) -> Option<RequestId> {
+        let response_id = Self::server_response_id(message)?;
+        if let Some(id) = pending_stream_response_ids.take(response_id) {
+            return Some(id);
         }
-        if let Some(id) = response_id.numeric_string_value() {
-            pending_stream_response_ids.remove(&RequestId::Number(id));
-        }
+        let id = RequestId::Number(response_id.numeric_string_value()?);
+        pending_stream_response_ids.take(&id)
     }
 
     async fn drain_queued_stream_messages(
@@ -643,7 +640,8 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
         loop {
             match sse_worker_rx.try_recv() {
                 Ok(message) => {
-                    Self::clear_stream_response_pending(pending_stream_response_ids, &message);
+                    let _ =
+                        Self::clear_stream_response_pending(pending_stream_response_ids, &message);
                     context.send_to_handler(message).await?;
                 }
                 Err(tokio::sync::mpsc::error::TryRecvError::Empty) => return Ok(()),
@@ -1676,18 +1674,13 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
                     let _ = responder.send(send_result);
                 }
                 Event::ServerMessage(mut json_rpc_message) => {
-                    if let Some(response_id) = Self::server_response_id(&json_rpc_message)
-                        && let Some(registration) = crate::service::remove_pending_request(
-                            &mut request_stream_cancellations,
-                            response_id,
-                        )
-                    {
-                        drop(registration);
-                    }
-                    Self::clear_stream_response_pending(
+                    // Match against all pending requests, not just open response streams.
+                    if let Some(request_id) = Self::clear_stream_response_pending(
                         &mut pending_stream_response_ids,
                         &json_rpc_message,
-                    );
+                    ) {
+                        drop(request_stream_cancellations.remove(&request_id));
+                    }
                     cache_tools_from_response(
                         &mut tool_header_cache,
                         &mut json_rpc_message,
@@ -2477,11 +2470,13 @@ mod tests {
             NumberOrString::String("1".into()),
         );
 
-        StreamableHttpClientWorker::<reqwest::Client>::clear_stream_response_pending(
-            &mut pending,
-            &response,
-        );
+        let matched_id =
+            StreamableHttpClientWorker::<reqwest::Client>::clear_stream_response_pending(
+                &mut pending,
+                &response,
+            );
 
+        assert_eq!(matched_id, Some(NumberOrString::Number(1)));
         assert!(pending.is_empty());
     }
 
@@ -2492,14 +2487,16 @@ mod tests {
         let mut pending = HashSet::from([NumberOrString::Number(1), string_id.clone()]);
         let response = ServerJsonRpcMessage::response(
             ServerResult::ListToolsResult(ListToolsResult::default()),
-            string_id,
+            string_id.clone(),
         );
 
-        StreamableHttpClientWorker::<reqwest::Client>::clear_stream_response_pending(
-            &mut pending,
-            &response,
-        );
+        let matched_id =
+            StreamableHttpClientWorker::<reqwest::Client>::clear_stream_response_pending(
+                &mut pending,
+                &response,
+            );
 
+        assert_eq!(matched_id, Some(string_id));
         assert_eq!(pending, HashSet::from([NumberOrString::Number(1)]));
     }
 }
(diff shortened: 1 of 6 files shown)


crates/rmcp/src/transport/worker.rs
```diff
@@ -190,15 +336,52 @@ impl<W: Worker> Transport<W::Role> for WorkerTransport<W> {
...
+        let registration = if W::supports_request_cancellation() {
+            match &item {
+                JsonRpcMessage::Request(request) => Some(RequestCancellationRegistration::new(
+                    request.id.clone(),
+                    self.ct.child_token(),
+                    self.request_cancellations.clone(),
+                )),
+                JsonRpcMessage::Notification(notification) => {
+                    cancellation_target =
+                        self.cancel_request_from_notification(&notification.notification);
```
DaleSeo, line 361: A request from the old session may still be running when recovery moves to a new generation. If its handler finishes afterward, this code stamps the response with the new generation, and the client posts it to the replacement session.

crates/rmcp/src/transport/worker.rs
```diff
@@ -142,11 +252,32 @@ impl<W: Worker> WorkerTransport<W> {
...
+            control_send_service: control_to_transport_tx,
+            request_cancellations,
+            control_generation,
             join_handle: Some(join_handle),
             ct: transport_task_ct.clone(),
             _drop_guard: transport_task_ct.drop_guard(),
         }
     }
+
+    fn cancel_request_from_notification(
```
oxcabe, line 268: nb: this function (alongside `is_control_message`, `supports_request_cancellation`, and the control-generation counter) would make `WorkerTransport` less transport agnostic. That is, implying that the cancellation policy of one transport would now live inside the shared mechanisms.

What I'd suggest here is to keep `WorkerTransport` generic and put the interception in a _streamable_ HTTP wrapper that implements `Transport` and delegates, e.g:

```rust
pub struct StreamableHttpClientTransport<C: StreamableHttpClient> {
    inner: WorkerTransport<StreamableHttpClientWorker<C>>,
    request_cancellations: RequestCancellations,
    control_tx: mpsc::Sender<WorkerSendRequest<...>>,
    control_generation: Arc<AtomicU64>,
}
```

Now `send()` on the wrapper should perform the registration, the cancellation lookup, and the control-lane routing, then delegate to inner.

As a note, this change would require that `StreamableHttpClientTransport` becomes a NewType.

crates/rmcp/src/transport/worker.rs
```diff
@@ -53,17 +64,110 @@ pub trait Worker: Sized + Send + 'static {
...
+    /// Keep this request alive while its work is active. Cloning the token does
+    /// not keep its cancellation registration alive.
+    pub fn cancellation_token(&self) -> Option<CancellationToken> {
+        self.cancellation
+            .as_deref()
+            .map(RequestCancellationRegistration::token)
+    }
+
+    /// Keep the same cancellation registration alive after the POST completes.
+    #[cfg(feature = "transport-streamable-http-client")]
```
oxcabe, line 154: nb: also a transport-specific feature as part of the generic worker model, similar situation to the `cancel_request_from_notification` comment down below.

crates/rmcp/src/transport/worker.rs
```diff
@@ -159,10 +290,25 @@ pub struct SendRequest<W: Worker> {
...
+    pub fn control_generation(&self) -> u64 {
+        self.control_generation.load(Ordering::SeqCst)
+    }
+
+    /// Advance the generation so the worker can reject older control sends.
+    pub fn advance_control_generation(&self) -> u64 {
+        self.control_generation
+            .fetch_add(1, Ordering::SeqCst)
+            .wrapping_add(1)
+    }
```
oxcabe: nb: similar situation with the rest of comments from this file, i.e. this might be too specific to the streamable HTTP implementation.

It's not that big of an issue if this stays here, but in that case I'd encourage expanding on the doc comment, explaining what the implications of "control generation" would be for any generic worker.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -26,13 +30,17 @@ use crate::{
...
+            RequestCancellationRegistration, Worker, WorkerQuitReason, WorkerSendRequest,
+            WorkerTransport,
+        },
     },
 };
 
 type BoxedSseStream = BoxStream<'static, Result<Sse, SseError>>;
 type SseTaskResult<E> = (Option<RequestId>, Result<(), StreamableHttpError<E>>);
 const SESSION_CLEANUP_TIMEOUT: Duration = Duration::from_secs(5);
+const CONTROL_POST_TIMEOUT: Duration = Duration::from_secs(5);
```
oxcabe: nb: this should be made configurable, as a reply POST that runs longer than
five seconds would inevitably fail with `ControlRequestTimeout`.

My suggestion would be to approach it in a similar way to how `session_recovery_timeout` is defined in line 1976, promoting it into a config field:

```
/// Maximum time that a cancellation or reply POST can run after it starts
/// (default: five seconds).
pub control_request_timeout: Duration,
```

oxcabe (approved):
LGTM 👍🏼 added some non-blocking comments

crates/rmcp/src/transport/worker.rs
nickcoai: You're right. The generation is captured when the reply's send is created, so it doesn't identify the session that started the handler. I checked the base implementation; it has the same late-handler routing issue.

Will send a follow up!

nickcoai:
Agreed that the transport-specific interception belongs closer to the http transport. Moving it into a wrapper also changes the existing public type alias and needs to preserve direct worker construction, so I'd keep that as a separate refactor. The registration accessor and its ownership should move together.

For this PR, I'll expand the generation docs: it is captured when send() creates its future, before queue admission. Advancing it affects subsequent sends; it doesn't drain queues, cancel work, or reject stale messages automatically. Those decisions remain the worker's responsibility.

crates/rmcp/src/transport/streamable_http_client.rs
nickcoai: Agreed. I'll add control_request_timeout and a matching builder, keeping the five-second default.

DaleSeo (approved):
Thanks @nickcoai @oxcabe!

PR #1194, opened 2026-08-19
fix(transport): cancel pending modern HTTP requests

### LINKED ISSUE #1193: streamable-http-client: RequestHandle::cancel hangs before response stream starts
**Describe the bug**

With protocol version `2026-07-28` over Streamable HTTP, `RequestHandle::cancel()` can hang while the original request POST is waiting for its first response event.

The client worker awaits `post_message_with_max_sse_event_size(...)` inline. While that future is pending, the worker cannot process the cancellation message that should close the request's HTTP/SSE response path. A long-running tool that emits no events therefore cannot be cancelled through RMCP's public client API.

This reproduces on current `main` (`6f8dcde`, `rmcp` 3.1.3).

**To Reproduce**

1. Start a stateless Streamable HTTP server using protocol version `2026-07-28`.
2. Make its `tools/call` handler wait for `RequestContext::ct` without returning a response or progress event.
3. Connect with `StreamableHttpClientTransport` and `ClientLifecycleMode::Discover`.
4. Start the call with `send_cancellable_request(...)`.
5. After the handler starts, call `RequestHandle::cancel(...)` inside a five-second timeout.

The cancel call times out:

```text
RequestHandle::cancel should not wait for the response stream: Elapsed(())
```

The regression test in the proposed PR exercises this exact flow using RMCP on both sides.

**Expected behavior**

`RequestHandle::cancel()` should return promptly. RMCP should abort the pending request POST, closing that request's response path so the server cancels the handler and fires `RequestContext::ct`.

This is the required behavior for the modern protocol:

- [MCP 2026-07-28 transport overview](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports#cancellation): Streamable HTTP cancellation closes the request's response stream.
- [MCP 2026-07-28 Streamable HTTP](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#cancellation): the server must treat that close as cancellation and stop work as soon as practical.

**Logs**

No transport error is logged. The cancellation future remains pending until the test timeout.

**Additional context**

The modern client path already converts a cancellation message into cancellation of a per-request `CancellationToken`, so this appears to be an ordering bug rather than an intentional limitation. The token is currently registered only after the POST produces an SSE response, which is too late to cancel a POST still waiting for its first response event.

Issues #857 and PR #967 fixed the matching server-side disconnect handling. This issue is the client-side path needed to trigger that behavior through `RequestHandle::cancel()`.

lucarlig:
Allow modern Streamable HTTP requests to be cancelled before their POST returns the first response event. The client worker stays responsive while request POSTs are in flight, and the same per-request cancellation token continues to own any returned SSE stream.

## Motivation and Context

Fixes #1193.

`RequestHandle::cancel()` currently waits forever when a `2026-07-28` request POST is still waiting for its response stream. The worker awaits the POST inline, so it cannot process the cancellation that should close that same request.

This is required by the modern MCP transport rules:

- [Transport overview: cancellation](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports#cancellation)
- [Streamable HTTP: cancellation](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#cancellation)

## How Has This Been Tested?

The new integration test uses an RMCP client and server over a real Axum HTTP listener. Before the fix, it fails after five seconds because `RequestHandle::cancel()` remains pending. With the fix, cancellation returns and the server's `RequestContext::ct` fires.

- `cargo test --all-features` — 783 passed, 24 ignored
- Full RMCP suite with all non-local features — 1,093 passed, 13 ignored
- Focused disconnect suite — 3 passed

## Breaking Changes

None. This fixes modern Streamable HTTP cancellation behavior without changing the public API. Legacy lifecycle, session, and re-initialization paths are unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The implementation uses RMCP's existing Tokio `JoinSet` and `CancellationToken`; it adds no dependency. A modern request's token is registered before its POST starts. Cancelling the request drops the pending HTTP future, and if the POST returns an SSE stream first, that same token controls the stream as before. Response-versus-cancellation races keep the existing behavior of ignoring a response after cancellation.

Issues #857 and PR #967 added the matching server-side disconnect handling. This change makes RMCP's public client cancellation API trigger that path even before response headers or the first SSE event arrive.

### DIFF
--- a/crates/rmcp/Cargo.toml
+++ b/crates/rmcp/Cargo.toml
@@ -454,7 +454,9 @@ path = "tests/test_streamable_http_connection_reuse.rs"
 [[test]]
 name = "test_streamable_http_disconnect_cancel"
 required-features = [
+  "client",
   "server",
+  "transport-streamable-http-client-reqwest",
   "transport-streamable-http-server",
   "reqwest",
 ]
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -32,6 +32,14 @@ use crate::{
 
 type BoxedSseStream = BoxStream<'static, Result<Sse, SseError>>;
 type SseTaskResult<E> = (Option<RequestId>, Result<(), StreamableHttpError<E>>);
+
+struct ModernPostTaskResult<E: std::error::Error + Send + Sync + 'static> {
+    request_id: Option<RequestId>,
+    responder: tokio::sync::oneshot::Sender<Result<(), StreamableHttpError<E>>>,
+    response: Option<Result<StreamableHttpPostResponse, StreamableHttpError<E>>>,
+    stream_ct: CancellationToken,
+}
+
 const SESSION_CLEANUP_TIMEOUT: Duration = Duration::from_secs(5);
 
 fn build_request_headers(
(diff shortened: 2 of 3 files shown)


lucarlig:
Closing this PR to move this cancellation fix into our downstream dataplane alignment and RMCP integration plan. This work will be tracked via a separate PR instead.

PR #1192, opened 2026-08-19
fix(http): enforce Origin validation semantics

lucarlig:
Adds opt-in strict Origin validation while preserving the existing default and non-empty allowlist behavior.

## Motivation and Context

When Origin validation is enabled, malformed and non-UTF-8 `Origin` headers currently return HTTP 400 instead of the specification-required HTTP 403.

This change adds `enforce_origin_validation()` so servers can reject every present Origin when the allowlist is empty without changing the current default. Requests without an `Origin` header continue to pass. The conformance server opts into strict validation explicitly.

Closes #1191 and includes the malformed-Origin fix from #1188.

## How Has This Been Tested?

- `cargo fmt --all -- --check`
- Focused custom-header integration tests: 29 passed
- Conformance-server test: 1 passed
- Server conformance suites for protocol versions 2025-11-25 and 2026-07-28

Tests cover allowed, disallowed, absent, malformed, and non-UTF-8 Origins; enforced empty allowlists; explicit disabling; and legacy compatibility paths.

## Breaking Changes

None. Strict empty-allowlist validation is opt-in. Existing default and non-empty allowlist behavior are preserved.

When validation is enabled, malformed and non-UTF-8 Origins now return HTTP 403 instead of HTTP 400.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

Existing Origin port-matching behavior is intentionally unchanged because the MCP specification does not define allowlist matching semantics.

PR #1196, opened 2026-08-20
rmcp: return forbidden for invalid origin

### LINKED ISSUE #1188: Streamable HTTP returns 400 instead of 403 for malformed Origin
## Problem

When Streamable HTTP Origin validation is enabled, a present malformed or non-UTF-8 `Origin` header is rejected with HTTP 400. The MCP 2026-07-28 Streamable HTTP transport specification requires a server to respond with HTTP 403 when a present `Origin` is invalid.

A syntactically valid but non-allowlisted origin already returns 403, so malformed values currently have observably different behavior.

## Reproduction

On current `main` / rmcp 3.1.3:

1. Configure `StreamableHttpService` with `with_allowed_origins(["https://app.example.com"])`.
2. Send a request containing `Origin: not-an-origin`.
3. Observe HTTP 400.

Expected: HTTP 403.

The same mismatch occurs when the header cannot be decoded as text.

Implementation: https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/rmcp/src/transport/streamable_http_server/tower.rs

Specification: https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http

## Proposed change

Return the existing forbidden response for Origin encoding and parsing failures, and add integration coverage for both malformed and non-UTF-8 values. Missing `Origin` behavior and explicitly disabled validation would remain unchanged.

This follows up on #822 / #823. I can submit the focused patch after confirmation, per the discuss-first contribution guidance.

ECD5A:
## Summary

- return `403 Forbidden` for malformed or non-UTF-8 `Origin` headers when Origin validation is enabled
- add regression coverage for both invalid-header cases

Fixes #1188

## Testing

- `cargo test -p rmcp --test test_custom_headers origin_validation --features 'client,server,transport-streamable-http-client-reqwest,transport-streamable-http-server' -- --nocapture`
- `git diff --check`
- targeted `rustfmt` on the two changed files

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -893,13 +893,13 @@ fn validate_origin_header(
         .inspect_err(|_| {
             tracing::warn!(origin = ?origin_header, "rejected request with non-UTF-8 Origin header");
         })
-        .map_err(|_| bad_request_response("Bad Request: Invalid Origin header encoding"))?;
+        .map_err(|_| forbidden_response("Forbidden: Invalid Origin header encoding"))?;
     let origin = parse_origin_value(origin_str).ok_or_else(|| {
         tracing::warn!(
             origin = origin_str,
             "rejected request with malformed Origin header",
         );
-        bad_request_response("Bad Request: Invalid Origin header")
+        forbidden_response("Forbidden: Invalid Origin header")
     })?;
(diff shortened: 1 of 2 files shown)


ECD5A:
Closing in favor of #1192, which already covers this status-code fix as part of the broader Origin validation change. I missed the overlap before opening this — sorry for the noise.

PR #1200, opened 2026-08-21
feat: allow externally supplied tool schemas

lucarlig:
Summary:
- let Streamable HTTP clients seed tool schemas learned by another discovery component
- let Streamable HTTP servers resolve schemas from request parts and extensions
- keep tools/list cache refresh and get_tool fallback behavior unchanged when the new APIs are unused

This enables gateways to validate and forward SEP-2243 parameter headers without issuing tools/list on every upstream connection, while keeping tenant-specific schemas isolated by request context.

Validation:
- cargo test -p rmcp --lib configured_tool_schema_builds_param_headers_without_tools_list --features client,transport-streamable-http-client-reqwest
- cargo test -p rmcp --test test_streamable_http_standard_headers --features server,client,transport-streamable-http-server,reqwest


lucarlig:
Closing this draft

PR #1205, opened 2026-08-23
refactor!: keep streamable http state in its transport

nickcoai:
## Summary

Follow up on [oxcabe's worker ownership review](https://github.com/modelcontextprotocol/rust-sdk/pull/1186#discussion_r3838250185) in #1186.

This draft is based on #1186, so its diff contains only the refactor and its tests.

- Move http request cancellation, the control queue, and session generation into `StreamableHttpClientTransport`.
- Keep the ordinary queue and worker task in `WorkerTransport`.
- Leave scheduling, retries, timeouts, cancellation behavior, and wire messages unchanged.

## Validation

Passed locally:

- 21 client-only concurrency tests, including worker construction and conversion.
- 37 selected transport tests using the non-local features from `justfile`.
- 13 http unit tests and six transport doctests.
- All-features `rmcp` suite: 506 unit tests, enabled integration tests, and 48 doctests passed (10 ignored), with the two exclusions below.
- Strict all-targets, all-features clippy with `-D warnings`.
- Two unchanged downstream custom-adapter integration tests.
- Nightly formatting and diff checks.

Two existing process tests remain excluded because the sandbox blocks process-state inspection: `test_tokio_child_process_drop` and `test_tokio_child_process_graceful_shutdown`. Neither test changed.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -623,16 +623,13 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
     fn clear_stream_response_pending(
         pending_stream_response_ids: &mut HashSet<RequestId>,
         message: &ServerJsonRpcMessage,
-    ) {
-        let Some(response_id) = Self::server_response_id(message) else {
-            return;
-        };
-        if pending_stream_response_ids.remove(response_id) {
-            return;
+    ) -> Option<RequestId> {
+        let response_id = Self::server_response_id(message)?;
+        if let Some(id) = pending_stream_response_ids.take(response_id) {
+            return Some(id);
         }
-        if let Some(id) = response_id.numeric_string_value() {
-            pending_stream_response_ids.remove(&RequestId::Number(id));
-        }
+        let id = RequestId::Number(response_id.numeric_string_value()?);
+        pending_stream_response_ids.take(&id)
     }
 
     async fn drain_queued_stream_messages(
@@ -643,7 +640,8 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
         loop {
             match sse_worker_rx.try_recv() {
                 Ok(message) => {
-                    Self::clear_stream_response_pending(pending_stream_response_ids, &message);
+                    let _ =
+                        Self::clear_stream_response_pending(pending_stream_response_ids, &message);
                     context.send_to_handler(message).await?;
                 }
                 Err(tokio::sync::mpsc::error::TryRecvError::Empty) => return Ok(()),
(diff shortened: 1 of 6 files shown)


crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -1866,13 +1970,56 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
 /// );
 /// ```
 ///
+/// This is a separate type, not a [`WorkerTransport`] alias.
+/// [`StreamableHttpClientWorker`] no longer implements [`Worker`].
```
oxcabe, line 1974: nb: I believe nobody would need this information as docs. My suggestion is to remove it and keeping the type docs timeless.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -485,15 +486,105 @@ struct SessionCleanupInfo<C> {
...
+        }
+    }
+}
+
+struct HttpSendRequest<C: StreamableHttpClient> {
+    message: ClientJsonRpcMessage,
+    responder: tokio::sync::oneshot::Sender<Result<(), StreamableHttpError<C::Error>>>,
+    cancellation: Option<Arc<RequestCancellationRegistration>>,
+    // Captured before the send future is polled.
+    control_generation: u64,
```
oxcabe, line 557: nb: consider making `control_generation` an `Option<u64>`, which then could be used to make the "ordinary sends carry no generation" rule unrepresentable.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -485,15 +486,105 @@ struct SessionCleanupInfo<C> {
...
+    client: C,
+    config: StreamableHttpClientTransportConfig,
+    control_rx: tokio::sync::mpsc::Receiver<HttpSendRequest<C>>,
+    control_generation: Arc<AtomicU64>,
+}
+
+type RequestCancellations = Arc<Mutex<HashMap<RequestId, Weak<RequestCancellationRegistration>>>>;
+
+/// Cancellation state for a request and its response stream.
+struct RequestCancellationRegistration {
```
oxcabe, line 507: nb: some extra docs for this


```suggestion
/// Cancellation state for a request and its response stream.
/// - `cancellation` fires when the request is cancelled; causing readers to
///   stop consuming the response stream.
/// - `lifetime` fires when the last owner of the registration drops. For legacy
///   http, the open response stream stays alive until then, so a custom adapter
///   can handle the cancellation POST while its stream state still exists.
/// - Owners are the pending map entry, the in-flight `HttpSendRequest`, and a
///   cancellation send future that targets this request.
struct RequestCancellationRegistration {
```

oxcabe (approved):
LGTM, thanks for addressing my review from https://github.com/modelcontextprotocol/rust-sdk/pull/1186 !

alexhancock:
Closing for now per discord chat with Nick, but can always revisit if we want parts of this

PR #1206, opened 2026-08-23
fix: reject replies from retired streamable http sessions

nickcoai:
## Summary

Fix the late-handler reply issue from [Dale's review](https://github.com/modelcontextprotocol/rust-sdk/pull/1186#discussion_r3837437415).

This PR is based on `main`, independently of the worker refactor in #1205. The late-handler routing issue predates #1186. Whichever follow-up lands second will need a rebase to combine their transport changes.

- Record the local transport and session generation when an inbound request is handed to the service.
- Carry that origin through automatic responses and errors. Do not send an old handler's reply to a replacement session, including during shutdown draining.
- Check handler identity when completing a request. If a new session reuses its request id, the old handler cannot remove or cancel the new one.
- Keep the origin local. There are no wire-format changes or new public transport methods.

## Scope

This protects automatic service replies, including request-association errors. Manually constructed raw transport replies have no captured origin.

Transport wrappers must preserve request extensions. Wrappers that defer the inner send within the same future retain the reply context; detached tasks must construct the inner send before spawning. Serializing requests or deferring send construction to a detached task loses the private context.

The tests cover late success and error replies, reused request ids, shutdown draining, and recovery during a pending reply send.

## Validation

Validated at `ad88b6f`, based on `main` at `3b5ca4d`:

- 245 unit tests and 24 concurrency tests with default features plus `client` and `transport-streamable-http-client`.
- All four new late-handler regressions fail on `main` at `3b5ca4d` and pass with this fix.
- Production feature checks for the client without a worker, the client with the generic worker, and the local-mode HTTP client.
- Nightly formatting and diff checks.

Full all-features tests, clippy, and downstream adapter tests were not rerun after this rebase.

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -1323,6 +1323,105 @@ where
     tokio::task::spawn_local(future)
 }
 
+/// Local routing metadata retained separately from a handler's wire response.
+#[derive(Debug, Clone, Default)]
+struct ReplyContext {
+    #[cfg(feature = "transport-worker")]
+    origin: Option<crate::transport::worker::ResponseOrigin>,
+}
+
+impl ReplyContext {
+    fn from_request(request: &mut impl GetExtensions) -> Self {
+        #[cfg(feature = "transport-worker")]
+        {
+            Self {
+                origin: request
+                    .extensions_mut()
+                    .remove::<crate::transport::worker::ResponseOrigin>(),
+            }
+        }
+        #[cfg(not(feature = "transport-worker"))]
+        {
+            let _ = request;
+            Self::default()
+        }
+    }
+
+    fn is_current(&self) -> bool {
+        #[cfg(feature = "transport-worker")]
+        {
+            self.origin
+                .as_ref()
+                .is_none_or(|origin| origin.is_current())
+        }
+        #[cfg(not(feature = "transport-worker"))]
+        {
+            true
+        }
+    }
+
+    fn send<R: ServiceRole, T: Transport<R>>(
+        self,
+        transport: &mut T,
+        message: TxJsonRpcMessage<R>,
+    ) -> impl Future<Output = Result<(), T::Error>> + Send + 'static {
+        let send = self.is_current().then(|| {
+            #[cfg(feature = "transport-worker")]
+            {
+                crate::transport::worker::with_response_origin(self.origin.clone(), || {
+                    transport.send(message)
+                })
+            }
+            #[cfg(not(feature = "transport-worker"))]
+            {
+                transport.send(message)
+            }
+        });
+        async move {
+            let Some(send) = send else {
+                return Ok(());
+            };
+            let result = send.await;
+            // Recovery may retire the origin between the initial check and the
+            // worker's dispatch check. Dropping that reply must not stop a drain
+            // from sending later, current-session replies.
+            if self.is_current() { result } else { Ok(()) }
+        }
+    }
+}
+
+#[derive(Debug)]
+struct HandlerResponse<R: ServiceRole> {
+    message: TxJsonRpcMessage<R>,
+    owner: Arc<CancellationToken>,
+    context: ReplyContext,
+}
+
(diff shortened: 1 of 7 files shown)


crates/rmcp/src/service.rs
```diff
@@ -1323,6 +1323,105 @@ where
...
+
+impl ReplyContext {
+    fn from_request(request: &impl GetExtensions) -> Self {
+        #[cfg(feature = "transport-worker")]
+        {
+            Self {
+                origin: request
+                    .extensions()
+                    .get::<crate::transport::worker::ResponseOrigin>()
+                    .cloned(),
```
DaleSeo: I think we should take the origin out at capture instead. `ReplyContext::from_request` only clones it, so the token rides `mem::swap` into the public `RequestContext::extensions`.

```suggestion
                     .extensions_mut()
                     .remove::<crate::transport::worker::ResponseOrigin>(),
```

crates/rmcp/src/service.rs
```diff
@@ -1323,6 +1323,105 @@ where
...
+            _ => return false,
+        };
+        if handlers
+            .get(id)
+            .is_some_and(|owner| Arc::ptr_eq(owner, &self.owner))
+        {
+            handlers.remove(id);
+            true
+        } else {
+            false
```
DaleSeo, line 1420: The old `Event::ToSink` arm logged before discarding a response.

```suggestion
            tracing::debug!(%id, "dropping response: request was cancelled or its id was reused");
            false
```

crates/rmcp/src/transport/worker.rs
```diff
@@ -320,10 +367,27 @@ impl<W: Worker> WorkerContext<W> {
             .wrapping_add(1)
     }
 
+    /// Check both the captured generation and, for replies, the originating transport.
+    #[cfg(feature = "transport-streamable-http-client")]
+    pub(crate) fn is_current_control(&self, request: &WorkerSendRequest<W>) -> bool {
```
DaleSeo, line 372: Is this supposed to be visible only within the crate?

crates/rmcp/src/transport/worker.rs
```diff
@@ -135,6 +177,8 @@ pub struct WorkerSendRequest<W: Worker> {
     pub responder: tokio::sync::oneshot::Sender<Result<(), W::Error>>,
     cancellation: Option<Arc<RequestCancellationRegistration>>,
     control_generation: u64,
+    #[cfg(feature = "transport-streamable-http-client")]
```
DaleSeo, line 180: Should the origin only be stored when `transport-streamable-http-client` is enabled?

DaleSeo:
Thanks for addressing my feedback, @nickcoai. I have a couple of questions before we move forward.

crates/rmcp/src/service.rs
```diff
@@ -1777,6 +1877,240 @@ where
...
+                owner: current.clone(),
+                context: ReplyContext::default(),
+            };
+            assert!(response.finish(&mut handlers));
+            assert!(current.is_cancelled());
+            assert!(handlers.is_empty());
+        }
+    }
+
+    #[cfg(all(feature = "transport-worker", not(feature = "local")))]
```
DaleSeo, line 1920: Do we not want to run these tests with `--all-features`?

crates/rmcp/src/service.rs
```diff
@@ -1323,6 +1323,105 @@ where
...
+        });
+        async move {
+            let Some(send) = send else {
+                return Ok(());
+            };
+            let result = send.await;
+            // Recovery may retire the origin between the initial check and the
+            // worker's dispatch check. Dropping that reply must not stop a drain
+            // from sending later, current-session replies.
+            if self.is_current() { result } else { Ok(()) }
```
DaleSeo, line 1388: so the drain can tell "dropped because stale" from "the transport just broke".

```suggestion
            match result {
                Err(error) if !self.is_current() => {
                    tracing::debug!(%error, "discarding retired reply's send failure");
                    Ok(())
                }
                result => result,
            }
```

PR #687, opened 2026-02-25
feat: mcp sdk conformance

alexhancock:
# Spec Conformance Testing

* adds conformance server and client for testing against https://github.com/modelcontextprotocol/conformance
* adds results from initial run of https://github.com/modelcontextprotocol/conformance/tree/main/.claude/skills/mcp-sdk-tier-audit skill (via goose)
* various small changes applied during the testing loop

PR #1208, opened 2026-08-23
fix: do not treat malformed JSON 200 as Accepted for requests

SebTardif:
## Summary

Fail a streamable HTTP request POST when the server returns HTTP 200 with `application/json` that is not a JSON-RPC message, instead of treating that body as Accepted and hanging `call_tool`.

## Problem

`post_message` maps any `application/json` body that fails to parse as `ServerJsonRpcMessage` to `StreamableHttpPostResponse::Accepted`. That fallback is correct for notification POSTs: nothing is waiting for a reply. The same branch also runs for requests.

The worker then records the request id as pending and waits on the SSE/GET stream. `call_tool` never returns when the server answers `200` + `Content-Type: application/json` with `{}`, an empty body, or other non-JSON-RPC JSON.

Public repro:

1. `StreamableHttpClientTransport::from_uri`
2. `().serve(transport)`
3. `peer.call_tool(...)`
4. Mock: initialize succeeds; `tools/call` returns `200` + `application/json` + `{}`

## Change

Keep the Accepted fallback only when the outbound message is a notification, response, or error. For a request, return `UnexpectedServerResponse`.

The same check is applied in the reqwest client and the Unix socket client. Notification empty-200 handling from Ref #849 is unchanged. This does not implement the unexpected-content-type fallback in open Ref #1063.

The parse-as-Accepted path was added in [#687](https://github.com/modelcontextprotocol/rust-sdk/pull/687) (2026-02-26) for notification POSTs that lack an `id`.

## Validation

Red (production restored from `origin/main`, new tests only):

```
cargo test -p rmcp --test test_streamable_http_malformed_json_request \
  --features 'client,transport-streamable-http-client,transport-streamable-http-client-reqwest'
```

Failed with `call_tool must return instead of hanging on malformed JSON 200: Elapsed(())` and `expected UnexpectedServerResponse, got: Ok(Accepted)`.

Green (after the fix): same command, 5 passed in 0.01s.

Also passed:

```
cargo test -p rmcp --test test_streamable_http_4xx_error_body \
  --test test_streamable_http_empty_2xx_notification \
  --features 'client,transport-streamable-http-client,transport-streamable-http-client-reqwest'
cargo test -p rmcp --test test_unix_socket_transport \
  --features 'client,server,transport-streamable-http-client-unix-socket'
cargo +nightly fmt --all -- --check
```

## Related

Ref #849 (accept empty 200 for notification POSTs)

Ref #1063 (notification POST response shape; still open). That PR tests that a request still rejects `text/plain`. It does not close JSON parse failure on a request POST.

### DIFF
--- a/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
@@ -303,17 +303,28 @@ impl StreamableHttpClient for reqwest::Client {
                 Ok(StreamableHttpPostResponse::Sse(event_stream, session_id))
             }
             Some(ct) if ct.as_bytes().starts_with(JSON_MIME_TYPE.as_bytes()) => {
-                // Try to parse as a valid JSON-RPC message. If the body is
-                // malformed (e.g. a 200 response to a notification that lacks
-                // an `id` field), treat it as accepted rather than failing.
+                // A notification/response/error POST does not await a JSON-RPC
+                // reply. Treat an unusable JSON body as Accepted. A request
+                // still needs a reply; the same body is an error so the worker
+                // does not wait on SSE forever.
                 match response.json::<ServerJsonRpcMessage>().await {
-                    Ok(message) => Ok(StreamableHttpPostResponse::Json(message, session_id)),
-                    Err(e) => {
+                    Ok(parsed) => Ok(StreamableHttpPostResponse::Json(parsed, session_id)),
+                    Err(e)
+                        if matches!(
+                            message,
+                            ClientJsonRpcMessage::Notification(_)
+                                | ClientJsonRpcMessage::Response(_)
+                                | ClientJsonRpcMessage::Error(_)
+                        ) =>
+                    {
                         tracing::warn!(
                             "could not parse JSON response as ServerJsonRpcMessage, treating as accepted: {e}"
                         );
                         Ok(StreamableHttpPostResponse::Accepted)
                     }
+                    Err(e) => Err(StreamableHttpError::UnexpectedServerResponse(Cow::Owned(
+                        format!("could not parse JSON response as ServerJsonRpcMessage: {e}"),
+                    ))),
                 }
             }
             _ => {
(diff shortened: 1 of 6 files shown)


crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
```diff
@@ -303,17 +303,28 @@ impl StreamableHttpClient for reqwest::Client {
...
+                        ) =>
+                    {
                         tracing::warn!(
                             "could not parse JSON response as ServerJsonRpcMessage, treating as accepted: {e}"
                         );
                         Ok(StreamableHttpPostResponse::Accepted)
                     }
+                    Err(e) => Err(StreamableHttpError::UnexpectedServerResponse(Cow::Owned(
+                        format!("could not parse JSON response as ServerJsonRpcMessage: {e}"),
+                    ))),
```
DaleSeo, line 328: Whoever hits this error is probably seeing for the first time that the server sent something unusable. But the message only includes the serde diagnostic, `expected value at line 1 column 1`, and drops what the server actually sent. The 4xx path in this same file includes the body as `HTTP {status}: {body}`. On the Unix socket side, we already have the body bytes, so including them would cost almost nothing. Did you consider including the body, or at least a prefix of it, or was leaving it out intentional?

SebTardif: @DaleSeo

> Did you consider including the body, or at least a prefix of it, or was leaving it out intentional?

The serde-only string was too thin. The request-POST error now includes a 256-byte lossy prefix of the body (`<empty>` when there is none), on both HTTP and Unix.

I did not copy the 4xx `HTTP {status}: {body}` form here. A 200 JSON body can be large, and this error is returned to `call_tool` callers. A prefix is enough to tell empty, `{}`, and HTML apart.

HTTP now reads `bytes()` and parses with `from_slice`, same as Unix, so the preview is available after the parse fails.

DaleSeo (approved):

Issue #1215, opened 2026-08-25
Track OpenSSF Scorecard security follow-ups


jamadeo:
## Context

OpenSSF Scorecard identified several follow-up opportunities while validating #1214.

## Follow-up work

- [x] Pin all GitHub Actions to immutable full commit SHAs, retaining version comments for readability — #1216
- [x] Pin the commitlint packages installed dynamically in CI — #1217
- [x] Establish restrictive top-level `GITHUB_TOKEN` permission defaults — #1218
- [x] Remove unnecessary `contents: write` from the coverage job in `ci.yml` — #1219
- [x] Tighten `release-plz.yml` permissions by moving write grants to job scope — #1220
- [x] Pin the temporary release-plz fork to an immutable revision — #1221
- [ ] Add fuzzing coverage for protocol parsing/deserialization and consider continuous OSS-Fuzz integration.

The linked PRs are intentionally focused so permission and dependency changes can be reviewed independently.

PR #1221, opened 2026-08-25
ci: pin release-plz fork revision

jamadeo:
## Summary

Pins the temporary release-plz fork dependency to an exact commit instead of resolving the mutable `feat/semver-check-features` branch at release time.

Pinned revision: `c2ea743bacf9bbd56e344df68219e60b0ad35d04`

Follow-up to the dependency hardening tracked in #1215.


jamadeo:
cc @DaleSeo in case this specifically needs to be a dynamic ref?

alexhancock (approved):

PR #1226, opened 2026-08-27
fix: map HeaderMismatch to HTTP 400

### LINKED ISSUE #1225: Handler-generated HeaderMismatch responses use HTTP 200 instead of 400
**Describe the bug**

When a modern Streamable HTTP request reaches a `ServerHandler` and the handler returns `ErrorData::header_mismatch(...)`, RMCP serializes the expected JSON-RPC `HeaderMismatch` error (`-32020`) but sends it with HTTP 200.

The modern response path derives the HTTP status through `jsonrpc_http_status`, which maps several protocol errors to HTTP 400 but currently falls through to HTTP 200 for `ErrorCode::HEADER_MISMATCH`. Transport-generated header validation errors do not show the bug because `header_mismatch_jsonrpc_response` constructs an HTTP 400 response directly.

**To Reproduce**

1. Implement a `ServerHandler` request method that returns `Err(ErrorData::header_mismatch("header does not match request parameter", None))`.
2. Send a valid modern Streamable HTTP JSON-RPC request using protocol version `2026-07-28`, so the request reaches the handler.
3. Observe an HTTP 200 response whose JSON-RPC body contains error code `-32020`.

**Expected behavior**

Handler-generated `HeaderMismatch` errors on the modern per-request path should use HTTP 400 while preserving JSON-RPC error code `-32020`, as required by the Streamable HTTP server-validation rules:

https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#server-validation

**Logs**

```text
actual:   HTTP 200, JSON-RPC error -32020
expected: HTTP 400, JSON-RPC error -32020
```

**Additional context**

The missing mapping is in:

https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/rmcp/src/transport/streamable_http_server/tower.rs

Related work such as #1083 covers mismatches detected by the transport before handler dispatch. This issue is specifically about `HeaderMismatch` returned by application handlers after dispatch.

lucarlig:
Map handler-generated JSON-RPC `HeaderMismatch` errors (`-32020`) to HTTP 400 on the modern per-request Streamable HTTP path. Add integration coverage for valid `2026-07-28` requests reaching a handler, with JSON responses and legacy session support each enabled and disabled. Verify HTTP status, JSON content type, request ID, error code, and the handler-generated message.

Fixes #1225.

## Motivation and Context

The MCP 2026-07-28 Streamable HTTP server-validation rules require header-validation failures to return HTTP 400 with JSON-RPC error code `-32020`.

RMCP already returns that combination when its transport detects a mismatch before dispatch. However, when application validation returns `ErrorData::header_mismatch(...)` from a `ServerHandler`, the response flows through `jsonrpc_http_status`, where `HEADER_MISMATCH` previously fell through to HTTP 200.

This change adds `HEADER_MISMATCH` to the modern bad-request mapping. Legacy requests continue to bypass this mapper and retain their existing HTTP 200 JSON-RPC behavior.

Specification:

https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#server-validation

## How Has This Been Tested?

Rebased onto main at `e27c5d1`. All four regression configurations fail with HTTP 200 instead of 400 when the mapping is removed, and pass with the fix restored.

Ran:

```text
cargo test -p rmcp --test test_streamable_http_json_response --test test_streamable_http_standard_headers --test test_streamable_http_protocol_version --test test_stateless_protocol_version --features server,client,transport-streamable-http-server,reqwest
cargo clippy -p rmcp --all-features --all-targets -- -D warnings
cargo fmt --all -- --check
git diff --check
```

Results: 49 focused tests passed. The 10-test JSON response suite also passed again after restoring the mapping. Clippy, formatting, and whitespace checks passed. Stable rustfmt emitted the repository's existing warnings for nightly-only configuration options.

## Breaking Changes

None. This corrects the HTTP status for an existing modern JSON-RPC error response. The JSON-RPC error code and legacy response behavior are unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Related transport-level validation paths already construct HTTP 400 responses directly. The new test specifically covers a `HeaderMismatch` returned after handler dispatch.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -618,12 +618,13 @@ fn jsonrpc_http_status(message: &ServerJsonRpcMessage) -> http::StatusCode {
     let ServerJsonRpcMessage::Error(error) = message else {
         return http::StatusCode::OK;
     };
-    // Modern per-request HTTP treats invalid params as a malformed request.
+    // Modern per-request HTTP treats protocol validation failures as malformed requests.
     // Legacy requests bypass this mapper and retain HTTP 200 JSON-RPC errors.
     match error.error.code {
         ErrorCode::UNSUPPORTED_PROTOCOL_VERSION
         | ErrorCode::MISSING_REQUIRED_CLIENT_CAPABILITY
-        | ErrorCode::INVALID_PARAMS => http::StatusCode::BAD_REQUEST,
+        | ErrorCode::INVALID_PARAMS
+        | ErrorCode::HEADER_MISMATCH => http::StatusCode::BAD_REQUEST,
         ErrorCode::METHOD_NOT_FOUND => http::StatusCode::NOT_FOUND,
         _ => http::StatusCode::OK,
     }
--- a/crates/rmcp/src/transport/streamable_http_server/tower.rs
+++ b/crates/rmcp/src/transport/streamable_http_server/tower.rs
@@ -639,12 +639,13 @@ fn jsonrpc_http_status(message: &ServerJsonRpcMessage) -> http::StatusCode {
     let ServerJsonRpcMessage::Error(error) = message else {
         return http::StatusCode::OK;
     };
-    // Modern per-request HTTP treats invalid params as a malformed request.
+    // Modern per-request HTTP treats protocol validation failures as malformed requests.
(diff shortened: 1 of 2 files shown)


lucarlig:
https://github.com/modelcontextprotocol/rust-sdk/pull/1259

PR #1231, opened 2026-08-31
fix(sse): saturate exponential reconnect backoff to avoid overflow panic

### LINKED ISSUE #1198: fix(sse): saturate exponential reconnect backoff
### Problem

ExponentialBackoff::retry computes the reconnect multiplier as 2^current_times using u32::pow. With max_times unset, current_times can reach 32; in debug builds this overflows and panics, while release arithmetic can wrap and produce an incorrect delay. The resulting behavior makes a long-lived SSE client vulnerable to a retry-loop panic or unexpectedly short backoff.

### Proposed direction

Saturate the multiplier at u32::MAX once current_times reaches the bit width, then use saturating multiplication for the base duration. This keeps the existing unbounded policy while making the delay monotonic and panic-free.

### Verification

A focused regression at current_times = 32 reproduces the overflow on the pre-fix implementation and passes with the saturating calculation. The rmcp test suite and clippy checks pass on the proposed patch.

Would maintainers like this fix submitted as a PR?

ump45nose:
## Summary

fix(sse): saturate exponential reconnect backoff to avoid overflow panic

## Verification

cargo test -p rmcp --features client-side-sse --lib client_side_sse (16 passed); cargo clippy -p rmcp --features client-side-sse --lib (no new warnings); cargo fmt --check (clean)

Related to https://github.com/modelcontextprotocol/rust-sdk/issues/1198

> AI assistance disclosure: AI was used to discover this opportunity and draft the change or text. The submission was checked against the prepared artifact and recorded verification evidence.

### DIFF
--- a/crates/rmcp/src/transport/common/client_side_sse.rs
+++ b/crates/rmcp/src/transport/common/client_side_sse.rs
@@ -205,10 +205,16 @@ impl Default for FixedInterval {
 pub struct ExponentialBackoff {
     pub max_times: Option<usize>,
     pub base_duration: Duration,
+    /// Upper bound on a single reconnect delay. The unbounded doubling policy can otherwise
+    /// produce delays of decades (once the multiplier saturates), which would pin the stream in
+    /// `tokio::time::sleep` forever — neither reconnecting nor terminating. Capping keeps the
+    /// backoff monotonic and panic-free while guaranteeing the client actually retries.
+    pub max_delay: Option<Duration>,
 }
 
 impl ExponentialBackoff {
     pub const DEFAULT_DURATION: Duration = Duration::from_millis(1000);
+    pub const DEFAULT_MAX_DELAY: Duration = Duration::from_secs(30);
 }
 
 impl Default for ExponentialBackoff {
@@ -216,6 +222,7 @@ impl Default for ExponentialBackoff {
         Self {
             max_times: None,
             base_duration: Self::DEFAULT_DURATION,
+            max_delay: Some(Self::DEFAULT_MAX_DELAY),
         }
     }
 }
@@ -229,10 +236,14 @@ impl SseRetryPolicy for ExponentialBackoff {
         }
         // `current_times` is unbounded when `max_times` is unset, so the exponent can reach
         // the bit width. Saturate the multiplier at `u32::MAX` and use saturating multiplication
-        // for the base duration so a long-lived SSE client gets a monotonic, panic-free delay
-        // instead of an overflow panic (debug) or a wrapped-to-zero backoff (release).
+        // for the base duration so the delay stays monotonic and panic-free instead of an
+        // overflow panic (debug) or a wrapped-to-zero backoff (release).
         let multiplier = 2u32.saturating_pow(current_times as u32);
-        Some(self.base_duration.saturating_mul(multiplier))
+        let delay = self.base_duration.saturating_mul(multiplier);
+        Some(match self.max_delay {
+            Some(max_delay) => delay.min(max_delay),
+            None => delay,
+        })
     }
 }
 
@@ -786,7 +797,11 @@ mod tests {
         // With `max_times` unset, `current_times` can reach the bit width. The old
         // `2u32.pow(current_times)` panicked in debug builds and wrapped in release;
         // the saturating implementation must return a monotonic, non-zero delay instead.
-        let policy = ExponentialBackoff::default();
+        let policy = ExponentialBackoff {
+            max_times: None,
+            base_duration: Duration::from_millis(1),
+            max_delay: None,
+        };
         let mut previous = Duration::ZERO;
         for current_times in [31usize, 32, 63, 64, 100] {
             let delay = policy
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/common/client_side_sse.rs
```diff
@@ -227,7 +227,12 @@ impl SseRetryPolicy for ExponentialBackoff {
         {
             return None;
         }
-        Some(self.base_duration * (2u32.pow(current_times as u32)))
+        // `current_times` is unbounded when `max_times` is unset, so the exponent can reach
+        // the bit width. Saturate the multiplier at `u32::MAX` and use saturating multiplication
+        // for the base duration so a long-lived SSE client gets a monotonic, panic-free delay
+        // instead of an overflow panic (debug) or a wrapped-to-zero backoff (release).
+        let multiplier = 2u32.saturating_pow(current_times as u32);
+        Some(self.base_duration.saturating_mul(multiplier))
```
DaleSeo: Once the multiplier reaches its limit, the delay can be decades long. The caller passes it directly to `tokio::time::sleep`, so the stream neither reconnects nor terminates.

ump45nose:
Thanks for the catch — you're right that a saturating multiplier alone leaves the delay unbounded, and at the saturated value `tokio::time::sleep` would park the stream for decades without reconnecting or terminating.

Pushed a follow-up that adds an optional `ExponentialBackoff::max_delay` (default `DEFAULT_MAX_DELAY` = 30s). The computed delay is now clamped to that ceiling, so the backoff stays monotonic and panic-free while guaranteeing the client actually retries. `max_delay: None` restores the previous unbounded (but now saturating) behavior for callers that want it.

Added a regression test (`exponential_backoff_caps_delay_at_max_delay`) asserting the delay grows monotonically, never exceeds the cap, and pins at `max_delay` past the saturation point.

> AI assistance disclosure: AI was used to discover this opportunity and draft the change or text. The submission was checked against the prepared artifact and recorded verification evidence.

crates/rmcp/src/transport/common/client_side_sse.rs
```diff
@@ -205,17 +205,24 @@ impl Default for FixedInterval {
...
     pub const DEFAULT_DURATION: Duration = Duration::from_millis(1000);
+    pub const DEFAULT_MAX_DELAY: Duration = Duration::from_secs(30);
 }
 
 impl Default for ExponentialBackoff {
     fn default() -> Self {
         Self {
             max_times: None,
             base_duration: Self::DEFAULT_DURATION,
+            max_delay: Some(Self::DEFAULT_MAX_DELAY),
```
DaleSeo: Because the struct is `#[non_exhaustive]`, downstream callers can only create one with `default()` and then mutate its fields. That means anyone who sets `max_times` will inherit the new 30-second limit.

Do you see this shorter window as a bug fix that could go into a patch release, or would leaving `max_delay` set to `None` be the safer default?

ump45nose:
Good question — thanks for thinking through the `#[non_exhaustive]` surface.

I'd frame it as: the 30s cap is a safety improvement, not the actual bug fix. The bug in #1198 is the overflow/wrap — that's fully resolved by the saturating multiplier + `saturating_mul`, independent of `max_delay`. The cap exists to guarantee a saturated policy still reconnects, but it does change observable timing for anyone relying on `ExponentialBackoff::default()`.

Given that, `None` as the default is the safer choice for this patch: it preserves the exact pre-existing behavior for downstream callers (unbounded, now panic-free) instead of silently imposing a 30s ceiling on anyone who constructs via `default()` and only tweaks `max_times`. That keeps it a clean bug-fix PR suitable for a patch release.

I'd suggest making `max_delay: None` the `Default` and letting opt-in callers set `Some(...)` explicitly (the internal `StreamableHttpClientTransportConfig` default can still set `Some(DEFAULT_MAX_DELAY)` where the SDK controls the surface). Would you prefer that, or keeping `Some(30s)` as the struct default with a documented behavior change?

> AI assistance disclosure: AI was used to discover this opportunity and draft the change or text. The submission was checked against the prepared artifact and recorded verification evidence.

crates/rmcp/src/transport/common/client_side_sse.rs
```diff
@@ -205,17 +205,24 @@ impl Default for FixedInterval {
 pub struct ExponentialBackoff {
     pub max_times: Option<usize>,
     pub base_duration: Duration,
+    /// Upper bound on a single reconnect delay. The unbounded doubling policy can otherwise
+    /// produce delays of decades (once the multiplier saturates), which would pin the stream in
+    /// `tokio::time::sleep` forever — neither reconnecting nor terminating. Capping keeps the
+    /// backoff monotonic and panic-free while guaranteeing the client actually retries.
```
DaleSeo: `SseAutoReconnectStream` checks the retry policy before each reconnect but `max_delay` only limits the policy's own output.


```suggestion
    /// Ceiling for the computed delay; `None` lets it grow until the multiplier saturates.
    /// A server-sent `retry:` field can still push the effective wait above this.
     pub max_delay: Option<Duration>,
```

DaleSeo:
@ump45nose Agreed, let’s use `max_delay: None` as the default. The overflow is the bug, while adding a ceiling would be a policy change. `StreamableHttpClientTransportConfig::default()` builds its policy from `ExponentialBackoff::default()`, so setting `Some(DEFAULT_MAX_DELAY)` there would still affect the same callers. Let's leave both as `None` for now and handle the ceiling in a follow-up, where we can call out the timing change separately.

ump45nose:
Implemented `max_delay: None` as the `ExponentialBackoff::default()`, exactly as agreed — the overflow is fixed by the saturating multiplier + `saturating_mul`, and the unbounded policy is preserved with no timing change for downstream callers. `DEFAULT_MAX_DELAY` is removed; callers that want a bounded reconnect delay opt in with `Some(...)`.

Verification (local, `cargo 1.98.0`):
- `cargo test -p rmcp --lib --features client-side-sse transport::common::client_side_sse::` → 17 passed
- `cargo clippy -p rmcp --lib --features client-side-sse` → only pre-existing warnings, none from this change

> AI assistance disclosure: AI was used to discover this opportunity and draft the change or text. The submission was checked against the prepared artifact and recorded verification evidence.

DaleSeo (approved):

PR #1232, opened 2026-08-31
feat(auth): coordinate OAuth refreshes through credential stores

mzeng-openai:
Add optional credential-store coordination for OAuth token refresh.

## Motivation and Context

OAuth clients sharing a rotating refresh token can read the same stored credentials and refresh concurrently. A caller can then try to use a refresh token that another caller has already rotated.

```mermaid
sequenceDiagram
    autonumber
    participant A as Process A
    participant B as Process B
    participant S as Shared credential store
    participant O as OAuth server
    A->>S: Load credentials
    S-->>A: Refresh token R0
    B->>S: Load credentials
    S-->>B: Refresh token R0
    A->>O: Refresh with R0
    O-->>A: New access token and refresh token R1
    Note over O: R0 has been consumed
    A->>S: Save credentials containing R1
    S-->>A: Save completed
    B->>O: Refresh with cached R0
    O-->>B: invalid_grant
```

Preventing this interleaving requires one guard across the credential reload, token exchange, and completed save.

`CredentialStore::acquire_refresh_guard` lets a store supply an owned guard. `AuthorizationManager::refresh_token` acquires it before loading credentials and holds it through the token exchange and completed save, so a waiting caller loads the latest saved refresh token. Guarded refreshes also check the stored client ID and use the stored granted scopes when the provider omits scopes.

`AuthError::CredentialStoreError` keeps backend and lock failures distinct from reauthorization, including refreshes triggered by an HTTP 401.

## Planned use in Codex

Codex plans to use this hook to coordinate OAuth token refreshes across processes that share credentials. One process will refresh and save the new credentials while others wait, so the next process uses the latest saved refresh token. This integration will follow separately in Codex.

## How Has This Been Tested?

Seven added regression cases cover guard ordering through completed save, concurrent token rotation, client mismatch, guard/load/save failures, and credential-store error propagation after a 401.

Local validation against upstream `main` at `3a2ebbc92034f6711e4eac9e93f5de0423be8dfe`:

- `cargo +nightly-2025-09-18 fmt --all -- --check` passed.
- `RUSTUP_TOOLCHAIN=stable CARGO_NET_OFFLINE=true just test` passed with and without `local`: 536 RMCP unit tests per configuration, all seven added regression cases in both runs, and the full integration and documentation suites. The installed stable toolchain is Rust 1.96.1.
- `RUSTUP_TOOLCHAIN=stable CARGO_NET_OFFLINE=true just check` passed with Clippy warnings denied.
- `git diff --check` passed.

The suite was run with permission to create its local Unix-socket fixtures. All three socket tests, three JavaScript interoperability tests, and two Python interoperability tests passed unchanged.

## Breaking Changes

None. Existing stores inherit a no-op guard hook. Stores opt in by implementing the hook; their `load` and `save` methods must not reacquire the same lock. Other credential writers are coordinated only if they use the same lock.

Cancellation behavior is unchanged. Concurrent callers can each refresh using the preceding caller's saved token; refreshes are not coalesced.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -288,11 +288,11 @@ pub trait CredentialStore: Send + Sync {
 
     async fn clear(&self) -> Result<(), AuthError>;
 
-    /// Optionally serialize refreshes that share these credentials.
+    /// Optionally coordinate refreshes that share these credentials.
     ///
     /// The manager acquires this guard before loading credentials and retains it
     /// through the token request and save. `load` and `save` must not reacquire
-    /// the same lock. Writers that bypass the guard are not serialized with it.
+    /// the same lock. Writers that bypass the guard are not coordinated with it.
     /// The default does not coordinate refreshes.
     async fn acquire_refresh_guard(&self) -> Result<Option<CredentialRefreshGuard>, AuthError> {
         Ok(None)
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -255,11 +255,31 @@ impl StoredCredentials {
     }
 }
 
+/// An owned guard held across a credential refresh and its save.
+///
+/// Stores can wrap a file lock, an owned mutex guard, or another coordination
+/// primitive. Dropping this value releases the guard.
+#[must_use = "dropping the guard releases refresh coordination"]
+pub struct CredentialRefreshGuard {
+    _guard: Box<dyn Send>,
+}
+
+impl CredentialRefreshGuard {
+    /// Wrap a guard whose lifetime coordinates access to the stored credentials.
+    pub fn new(guard: impl Send + 'static) -> Self {
+        Self {
+            _guard: Box::new(guard),
+        }
+    }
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -255,18 +255,48 @@ impl StoredCredentials {
...
+/// so they remain distinct from errors requiring reauthorization.
 #[async_trait]
 pub trait CredentialStore: Send + Sync {
     async fn load(&self) -> Result<Option<StoredCredentials>, AuthError>;
 
     async fn save(&self, credentials: StoredCredentials) -> Result<(), AuthError>;
 
     async fn clear(&self) -> Result<(), AuthError>;
+
+    /// Optionally serialize refreshes that share these credentials.
```
alexhancock: Maybe use the word "coordinate" instead of "serialize"? I was confused on first reading that we were talking about serializing something for storage.

alexhancock (dismissed):

alexhancock (approved):

PR #1236, opened 2026-09-01
fix(auth): unify refresh checks and error handling

DaleSeo:
## Motivation and Context

PR #1232 added `CredentialStore::acquire_refresh_guard`. Three refresh behaviors ended up depending on whether a store returns a guard.

-  The stored client ID is checked only for guarded stores.
-  The stored granted scopes are treated as authoritative only for guarded stores.
-  `call_reacting_to_challenges` only propagates `CredentialStoreError` and converts every other refresh error into `AuthRequired`.

The presence of a guard tells us whether a store can coordinate. It does not tell us whether its contents are trustworthy. For example, a store might return `Ok(None)` in single-process mode because it does not need a lock. In that case, we would silently skip the client ID check. The trait docs also describe this hook as coordination only, so implementors have no reason to know that validation depends on it.

This PR makes the following changes:

-  It applies the client ID check to every store, so mismatches are caught before the refresh token leaves the process.
-  It routes refresh scopes through the existing `resolve_granted_scopes` helper. This follows the same SEP-2350 and RFC 6749 section 5.1 rule already used by the authorization-code path. The `current_scopes` fallback is used only when no scopes were recorded.
-  It maps only `AuthorizationRequired` to the challenge in `call_reacting_to_challenges` and propagates every other error. This removes the special handling for `CredentialStoreError`.

## How Has This Been Tested?

Added regression tests

## Breaking Changes

None

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/tests/test_live_oauth_refresh.rs
+++ b/crates/rmcp/tests/test_live_oauth_refresh.rs
@@ -1,0 +1,233 @@
+//! Live authorization-server checks for the refresh path.
+//!
+//! These are `#[ignore]`d, so they never run in CI. Provision the authorization
+//! server with `./scripts/keycloak-oauth-fixture.sh`, then run them with
+//! `cargo test -p rmcp --all-features --test test_live_oauth_refresh -- --ignored`.
+//! `KC_BASE` overrides the server location for both the script and these tests.
+#![cfg(feature = "auth")]
+
+use std::sync::Arc;
+
+use rmcp::transport::auth::{
+    AuthError, AuthorizationManager, AuthorizationMetadata, CredentialRefreshGuard,
+    CredentialStore, InMemoryCredentialStore, OAuthClientConfig, OAuthTokenResponse,
+    StoredCredentials,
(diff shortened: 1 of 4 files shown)


crates/rmcp/tests/test_live_oauth_refresh.rs
```diff
@@ -0,0 +1,233 @@
+//! Live authorization-server checks for the refresh path.
```
jamadeo, line 1: These are nice tests - maybe as a follow up we should consider adding to CI? Maybe not worth it with the setup, but something to think about.

jamadeo (approved):

crates/rmcp/tests/test_live_oauth_refresh.rs
alexhancock: Yes agree some sort of live/smoke testing in CI would be good, and a next step for this project in ci

alexhancock (approved):

PR #1241, opened 2026-09-02
fix: treat empty cacheScope as omitted

### LINKED ISSUE #1242: Empty cacheScope on tools/list drops every tool
**Describe the bug**

**To Reproduce**

```json
{
  "ttlMs": 0,
  "cacheScope": ""
}
```

**Expected behavior**

**Additional context**

ismail-rt:
Fixes #1242.

## Motivation and Context

SEP-2549 allows `cacheScope` to be `"public"`, `"private"`, or omitted. Some hosted servers send `""`. The deserializer treats that as an unknown variant, so `tools/list` (and `resources/read`) fail to parse as the typed result.

`ServerResult` is untagged. When `ListToolsResult` rejects the payload, it falls through to `CustomResult`. The tools are on the wire and then disappear.

Negative `ttlMs` is already clamped to `0`. Empty `cacheScope` now maps to `None`. Unknown or whitespace values still error.

## How Has This Been Tested?

- `empty_cache_scope_is_treated_as_omitted` in `crates/rmcp/tests/test_cache_hints.rs`
- `unknown_cache_scope_still_errors` (`"shared"` and `" "` still fail)
- `empty_cache_scope_list_tools_result_does_not_fall_through` in `crates/rmcp/tests/test_deserialization.rs`

## Breaking Changes

None. Empty `cacheScope` previously failed to decode. It is now treated as omitted.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -1572,6 +1572,28 @@ where
     Ok(value.map(|ttl_ms| ttl_ms.max(0) as u64))
 }
 
+/// Normalize a `cacheScope` value during deserialization.
+///
+/// SEP-2549 permits `"public"` or `"private"`. Omission is also valid. Some
+/// hosted servers emit an empty string; treat that exact sentinel as omitted
+/// instead of failing the entire list/read result. Unknown or whitespace
+/// values still error.
+fn deserialize_cache_scope<'de, D>(deserializer: D) -> Result<Option<CacheScope>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let value = Option::<String>::deserialize(deserializer)?;
+    match value.as_deref() {
+        None | Some("") => Ok(None),
+        Some("public") => Ok(Some(CacheScope::Public)),
+        Some("private") => Ok(Some(CacheScope::Private)),
+        Some(other) => Err(serde::de::Error::unknown_variant(
+            other,
+            &["public", "private"],
+        )),
+    }
+}
+
 macro_rules! paginated_result {
     ($t:ident {
         $i_item: ident: $t_item: ty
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/model.rs
DaleSeo, line 1196: If a server is loose enough to emit `""` on `tools/list`, is there a reason to expect it to be stricter on discover?

crates/rmcp/src/model.rs
```diff
@@ -1572,6 +1572,28 @@ where
...
+    let value = Option::<String>::deserialize(deserializer)?;
+    match value.as_deref() {
+        None | Some("") => Ok(None),
+        Some("public") => Ok(Some(CacheScope::Public)),
+        Some("private") => Ok(Some(CacheScope::Private)),
+        Some(other) => Err(serde::de::Error::unknown_variant(
+            other,
+            &["public", "private"],
+        )),
+    }
```
DaleSeo, line 1594: We can handle only that sentinel and delegate the rest to the derived impl.

```suggestion
    use serde::de::IntoDeserializer;
    match Option::<std::borrow::Cow<'de, str>>::deserialize(deserializer)?.as_deref() {
         None | Some("") => Ok(None),
         Some(other) => CacheScope::deserialize(other.into_deserializer()).map(Some),
     }
```

prins1bap-ui:
Additional production evidence for this compatibility fix: ChatGPT's curated Upwork app now reproduces the same failure boundary as the Codex reports in openai/codex#41437. Fresh OAuth succeeds and the curated tool schemas are visible, but the first authenticated Upwork operation returns no data and the host disables the Upwork tool surface. This was reproduced with account discovery, direct organization read, and a direct marketplace job read. A clean uninstall/reinstall and app-permission changes did not alter it. OpenAI support cases 14667036/14667065 and Upwork MCP support ticket #55494650 now have the protocol evidence. Given that this PR is mergeable and CI/CodeQL/conformance are green, the narrow exact-empty-string normalization here appears to be a useful client-side interoperability safeguard even if Upwork also fixes its server response. No credentials or private account identifiers included.

PR #1243, opened 2026-09-03
fix(macros): accept const paths and concat! in tool/prompt descriptions

### LINKED ISSUE #1175: `#[tool(description = …)]` accepts only a bare literal, while `#[doc]` and `#[schemars]` on the same items accept `include_str!` and `const` paths
## Summary

`description` in `#[tool]` accepts only a bare string literal. Two attributes that sit on the very same items are more permissive: `#[doc]` on the function accepts `include_str!` and that value does reach the published description, and `#[schemars(extend(…))]` on the argument struct accepts `const` paths.

So a server can already source a tool's published text from outside the attribute — through `#[doc]`, or through a literal handed in by a wrapper macro — just not through `description`, which is the field that names it. Accepting a `const` path and `concat!` there would remove the inconsistency.

This is an enhancement request, not a blocker: the workarounds below all work.

## Version

rmcp 3.1.2, rustc 1.96.1, edition 2024.

## What was measured

| form | result |
|---|---|
| `#[tool(description = "…")]` | works |
| `#[tool(description = my_crate::caps::TEXT)]` | `error: Unexpected type \`path\`` |
| `#[tool(description = concat!("a", "b"))]` | `error: Unexpected type \`macro\`` |
| `/// doc comment` + `#[tool]` | works — becomes the description |
| `#[doc = include_str!("text.md")]` + `#[tool]` | **works** — becomes the description |
| `#[schemars(extend("maximum" = MOST_STARTS_AT_ONCE))]` on the args struct | **works** — a `const` path, in the same attribute stack |
| doc comment *and* `description` together | `description` wins (no surprise here) |
| `macro_rules!` wrapping the whole `#[tool_router] impl` | works |
| `macro_rules!` wrapping only the fn inside the impl | compiles, registers nothing — filed separately |

The `schemars` row is the one that makes this feel like an oversight rather than a decision: a `const` is already acceptable one attribute away, for a value that is likewise baked into the published schema.

## Why it matters on a larger surface

On a server with many tools the published text is an interface, and it usefully has two halves: a short one broadcast in `tools/list` before the first call, and a longer one fetched on demand. Ours measures 82 953 B of `description` across 66 tools, against a 13 590 B fetchable handbook — so keeping the split is worth real bytes, and it is far easier to keep the two halves honest when both can be declared together and the short one referenced from the attribute:

```rust
// what one would like to write
const GET_OR_CREATE_PROJECT: Capability = Capability {
    broadcast: "…what prevents a wrong call…",
    taught:    "…the full explanation, fetched on demand…",
};

#[tool(description = GET_OR_CREATE_PROJECT.broadcast)]
```

Today that requires either moving the text into a file and using `#[doc = include_str!(…)]`, or keeping a second copy plus a test comparing it to the literal.

## Suggested change

Accept in `description` what neighbouring attributes already accept: a path to a `const &'static str`, and `concat!`. The generated code stores a `&'static str` either way.

If the literal is required because the value is needed during macro expansion, a line saying so in the docs — and an error message reading "only a string literal is accepted here" rather than naming the token type — would save the next person the experiment.

## What this report does not establish

It does not claim the current behaviour is wrong, only that it is inconsistent with `#[doc]` and `#[schemars]` on the same items. `name` was not tested against a non-literal; only `description` was.

atirna:
Fixes #1175

## Motivation and Context

`description` in `#[tool]` and `#[prompt]` only accepted a bare string literal, while neighbouring attributes on the very same items are more permissive: `#[doc]` on the function accepts `include_str!` and that value reaches the published description, and `#[schemars(extend(...))]` on the argument struct accepts const paths.

So the published text could already be sourced from outside the attribute, just not through the field that names it. On a server with many tools, keeping the short broadcast description next to the longer fetchable one is much easier when both can be declared together and the short one referenced from the attribute:

```rust
const GET_PROJECT_DESCRIPTION: &str = "...";

#[tool(description = GET_PROJECT_DESCRIPTION)]
```

## How Has This Been Tested?

- before, on current `main`: a small crate with `#[tool(description = SOME_CONST)]` fails with `error: Unexpected type `path``, and `#[tool(description = concat!("a", "b"))]` fails with `error: Unexpected type `macro``
- after: both compile, and the generated `<fn>_tool_attr().description` equals the const value / `"ab"`; a bare string literal, `#[doc]`-derived and `include_str!` descriptions keep working; `description = 42` is still rejected with a normal type error at the attribute
- added regression tests in `test_tool_macros.rs` (const path, `concat!`) and `test_prompt_macros.rs` (const path); `cargo test -p rmcp-macros`, the focused integration tests, and the full `cargo test -p rmcp --features "client server transport-io which-command"` suite pass locally, plus `cargo clippy --all-targets --all-features -- -D warnings` and `cargo +nightly fmt --all -- --check`

Implementation notes: the field is parsed through darling's `PreservedStrExpr`, whose `FromMeta` keeps a string literal a literal instead of parsing its contents as code, so existing `description = "..."` usage is unchanged; the generated code stores a `&'static str` either way, exactly as before.

## Breaking Changes

None. Existing string-literal descriptions compile to the same code; only inputs that previously failed to compile are now accepted.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/crates/rmcp-macros/src/prompt.rs
+++ b/crates/rmcp-macros/src/prompt.rs
@@ -1,5 +1,5 @@
 use darling::{FromMeta, ast::NestedMeta};
-use proc_macro2::{Span, TokenStream};
+use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
 use syn::{Expr, Ident, ImplItemFn, ReturnType};
 
@@ -12,8 +12,10 @@ pub struct PromptAttribute {
     pub name: Option<String>,
     /// Human readable title of prompt
     pub title: Option<String>,
-    /// Optional description of what the prompt does
-    pub description: Option<String>,
+    /// Optional description of what the prompt does. A string literal, an
+    /// expression that evaluates to a `&'static str` (such as a `const` path
+    /// or `concat!`), or a `#[doc]` attribute on the function.
+    pub description: Option<darling::util::PreservedStrExpr>,
     /// Arguments that can be passed to the prompt
     pub arguments: Option<Expr>,
     /// Optional icons for the prompt
@@ -104,11 +106,8 @@ pub fn prompt(attr: TokenStream, input: TokenStream) -> syn::Result<TokenStream>
     };
 
     let name = attribute.name.unwrap_or_else(|| fn_ident.to_string());
-    let description = if let Some(s) = attribute.description {
-        Some(Expr::Lit(syn::ExprLit {
-            attrs: Vec::new(),
-            lit: syn::Lit::Str(syn::LitStr::new(&s, Span::call_site())),
-        }))
+    let description = if let Some(description) = attribute.description {
+        Some(description.into())
     } else {
         fn_item.attrs.iter().try_fold(None, extract_doc_line)?
     };
--- a/crates/rmcp-macros/src/tool.rs
+++ b/crates/rmcp-macros/src/tool.rs
@@ -1,7 +1,7 @@
 use darling::{FromMeta, ast::NestedMeta};
-use proc_macro2::{Span, TokenStream};
+use proc_macro2::TokenStream;
 use quote::{ToTokens, format_ident, quote};
-use syn::{Expr, Ident, ImplItemFn, LitStr, ReturnType, parse_quote};
+use syn::{Expr, Ident, ImplItemFn, ReturnType, parse_quote};
 
 use crate::common::extract_doc_line;
 
@@ -65,7 +65,10 @@ pub struct ToolAttribute {
     pub name: Option<String>,
     /// Human readable title of tool
     pub title: Option<String>,
-    pub description: Option<String>,
+    /// The description of the tool. A string literal, an expression that
+    /// evaluates to a `&'static str` (such as a `const` path or `concat!`),
+    /// or a `#[doc]` attribute on the function.
+    pub description: Option<darling::util::PreservedStrExpr>,
     /// A JSON Schema object defining the expected parameters for the tool
     pub input_schema: Option<Expr>,
     /// An optional JSON Schema object defining the structure of the tool's output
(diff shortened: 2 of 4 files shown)


DaleSeo (dismissed):
Thanks for the fix, @atirna! Can we update the rustdoc usage tables for the tool proc macro? These are the public entry points users see first.

https://github.com/atirna/rust-sdk/blob/2d8a8eb610b2381b11618cf7d5166a60612d632b/crates/rmcp-macros/src/lib.rs#L24

https://github.com/atirna/rust-sdk/blob/2d8a8eb610b2381b11618cf7d5166a60612d632b/crates/rmcp-macros/src/lib.rs#L211

atirna:
done, both usage tables updated. `description` now reads `Expr` with a note that it takes a string literal or an expression evaluating to a `&'static str` (const path, `concat!`), matching the wording already on the attribute struct fields.

also made the tool table say "will be used if not provided" like the prompt one already did, it was just missing that half of the sentence.

heads up on verification: i couldnt run `cargo doc` locally, this box has an unaccepted Xcode license so the build scripts fail to link. its a doc-comment only change so CI should cover it.

atirna:
update on the verification note above, i got a working toolchain since. `cargo doc -p rmcp-macros --no-deps` and the macro tests both pass on this head:

```
test result: ok. 13 passed; 0 failed; 0 ignored
```

DaleSeo (dismissed):
Thanks for addressing my feedback, @atirna!

DaleSeo:
@atirna Could you rebase your branch onto the latest main to trigger a fresh CI run? Then, the SemVer check should pass.

atirna:
Rebased onto latest main (head 31d0621). Focused checks still pass on the rebased tree: macro unit tests, tool/prompt macro integration tests, clippy, and rustdoc generation.

DaleSeo (approved):

Issue #1180, opened 2026-08-17
Unbounded response bodies on the streamable-HTTP client POST path (JSON success body and any error body) — memory-exhaustion denial-of-service


tbourrillon:
### Summary

The streamable-HTTP client bounds SSE bodies but not the two non-SSE response bodies it reads on the POST path. `bounded_sse_stream(..., max_sse_event_size)` is applied to every `text/event-stream` response, so the streaming shape is protected — but in the same function, a success response with `Content-Type: application/json` is read with `response.json()`, and *any* non-success response is read with `response.text()` and then interpolated whole into an error message. Neither is bounded, and `reqwest` applies no response-size limit of its own, so a remote server can make the client buffer an arbitrary amount of memory from one HTTP response.

This is the HTTP counterpart of #1030 (unbounded line buffer in `AsyncRwTransport::receive`, stdio). Filing separately because it is a different transport, a different file, and a different fix — and because the peer is materially more untrusted here: a stdio peer is a child process the client chose to spawn, whereas a streamable-HTTP peer is a remote server.

**Severity:** Medium (denial-of-service, untrusted input, no data exposure).

### Affected versions

Confirmed in `rmcp` 3.1.2 (current latest, published 2026-08-07), with the `client-transport-streamable-http-reqwest` feature.

### Code reference

`src/transport/common/reqwest/streamable_http_client.rs` (`crates/rmcp/...` in the repo), inside `post_message_with_max_sse_event_size`:

```rust
// :278-297 — any non-success status
if !status.is_success() {
    let body = response
        .text()                                   // <-- unbounded
        .await
        .unwrap_or_else(|_| "<failed to read response body>".to_owned());
    ...
    return Err(StreamableHttpError::UnexpectedServerResponse(Cow::Owned(
        format!("HTTP {status}: {body}"),          // <-- and copied whole into an error
    )));
}

match content_type.as_deref() {
    Some(ct) if ct.as_bytes().starts_with(EVENT_STREAM_MIME_TYPE.as_bytes()) => {
        let event_stream = bounded_sse_stream(response.bytes_stream(), max_sse_event_size);
        Ok(...)                                    // <-- bounded, correctly
    }
    Some(ct) if ct.as_bytes().starts_with(JSON_MIME_TYPE.as_bytes()) => {
        match response.json::<ServerJsonRpcMessage>().await {   // :309, unbounded
```

[...]

PR #1246, opened 2026-09-05
fix(transport): bound buffered Streamable HTTP responses

GTanger:
## Summary

Refs #1030. Addresses the Streamable HTTP client JSON/error response scope described in #1180 and redirected to #1030 by the maintainer. The stdio work in #1049 remains separate; this PR should not close the combined tracker on its own.

The built-in clients already bound individual SSE events, but buffer JSON and HTTP error bodies before parsing them. This change adds configurable pre-decode bounds without replacing the transports or adding dependencies.

## Changes

- Add `StreamableHttpResponseLimits` and transport config setters. Defaults: **64 MiB per complete JSON response, regardless of HTTP status**, **64 KiB per HTTP diagnostic prefix**, and the unchanged independent **16 MiB SSE-event limit**.
- Share bounded and prefix readers between reqwest and Unix socket backends. JSON is rejected with `ResponseBodyTooLarge { limit }` if Content-Length exceeds its limit or a chunk would exceed it. Count bytes even when the length is missing or understated; do not drain oversized JSON to EOF.
- For non-JSON HTTP error pages, `read_truncated_body` retains at most the diagnostic limit and stops immediately when that prefix is full, without polling for another chunk or EOF. Large legacy 4xx pages can still reach the existing `server/discover` -> `initialize` fallback. The fallback eligibility rules are unchanged.
- `application/json` error responses use the same complete-body JSON limit as successful responses. Parse the full bounded body before considering discovery fallback or truncating diagnostic text. Never parse a diagnostic prefix as a complete JSON-RPC message. Malformed/non-error JSON gets only a diagnostic prefix; oversized JSON still fails without fallback/retry.
- Align the Unix socket backend with reqwest by forwarding complete HTTP JSON-RPC errors as `StreamableHttpPostResponse::Json`, before synthetic discovery fallback. This also fixes the previous backend discrepancy for these errors.
- Forward limits through all six POST sending sites, including startup, discovery fallback, initialized notifications, ordinary/control requests, and session recovery. `AuthClient` preserves them across token refresh.
- Keep the old POST entry points and provide a default implementation of the new trait method that delegates to the old SSE-limit override. Built-in backends apply the new defaults through old entry points too.

## Compatibility and limits

Custom HTTP backends remain source-compatible, but must override `post_message_with_response_limits` to enforce JSON/diagnostic limits; the default cannot bound a body already decoded by custom code. The API docs explain this distinction and the JSON-versus-diagnostic behavior.

Previously accepted JSON bodies above the new default now fail; callers can raise each limit independently. A zero JSON limit accepts only empty bodies. A zero diagnostic limit omits diagnostic text rather than rejecting a non-JSON error response. The defaults are proposed for review.

Existing 202/204, authentication challenge, session-expired, within-limit malformed-success-JSON, and backend-specific success-body-I/O-error behavior is preserved. An error-body I/O failure before the diagnostic limit still uses the existing read-failure placeholder (subject to the diagnostic prefix limit). This does not subsume #1208's separate malformed-JSON proposal.

Limits count bytes yielded by the HTTP backend (decompressed bytes when decompression is enabled). They bound accumulated body length, not backend/decompressor allocations, vector capacity, UTF-8 conversion, or JSON parser allocations, and are not a total RSS bound.

## Validation

Linux, Rust 1.96.1 (the repository's pinned 1.96 toolchain), with the same fixed validation lockfile kept out of this PR:

- **686 passing tests, 0 failures/ignored**: 503 rmcp library tests plus 183 integration tests in 19 targeted binaries. This includes 71 cases added by the PR: body readers 12, OAuth forwarding 1, reqwest 39, Unix socket 8, lifecycle/custom-client compatibility 10, and old-public-API diagnostic regression 1.
- Red/green review regression: the exact same 58 focused integration tests and lockfile on the original PR head `296d1faea6ec1fd3b538eb629e93bd9910725778` compile and produce 31 passes / 27 assertion failures; all 58 pass with the revision. This includes the two real loopback legacy startup regressions.
- Real loopback startup completes `server/discover` -> `initialize` -> `notifications/initialized` within two seconds after a >64 KiB diagnostic error page, with either a fixed body or an unfinished chunked stream. Oversized JSON on HTTP 200 and 400 still fails within two seconds, with exactly one POST and no fallback/retry.
- Identical JSON-RPC errors on HTTP 200/400 share exact-limit and N+1 behavior on both backends. Coverage also includes a >16 MiB base64 tool result accepted by the new default, old POST APIs, zero limits, EOF-free diagnostic prefixes, partial-JSON negatives, auth/session precedence, independent SSE limits, and lifecycle limit forwarding.
- Strict Clippy passes for the affected client library and four focused integration binaries. Both `cargo +nightly-2026-06-20 fmt --all --check` and the repository hook's `cargo +nightly fmt --all --check` pass, as does `git diff --check`.
- Four no-default-feature builds pass: transport without a built-in backend, Unix-only, reqwest+auth, and both backends+auth+local. Existing feature-specific dead-code warnings remain.

The initial submission's broader server-enabled `--all-targets -D warnings` Clippy check reported the same 11 diagnostics in unchanged `streamable_http_server/tower.rs` on both baseline and patch; that broad check was not rerun for this revision. No lint suppression or unrelated server cleanup is included. Full workspace/all-features tests, Python/JS interoperability, and an actual MSRV-1.88 compiler run were not performed locally; CI remains a separate check.

Prepared with AI assistance; reviewed and tested locally before submission.

### DIFF
--- a/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
@@ -9,6 +9,7 @@ use crate::{
     model::{ClientJsonRpcMessage, JsonRpcMessage, ServerJsonRpcMessage},
     transport::{
         common::{
+            client_side_body::read_bounded_body,
             client_side_sse::{DEFAULT_MAX_SSE_EVENT_SIZE, bounded_sse_stream},
             http_header::{
                 EVENT_STREAM_MIME_TYPE, HEADER_LAST_EVENT_ID, HEADER_SESSION_ID, JSON_MIME_TYPE,
@@ -176,13 +177,13 @@ impl StreamableHttpClient for reqwest::Client {
         auth_token: Option<String>,
         custom_headers: HashMap<HeaderName, HeaderValue>,
     ) -> Result<StreamableHttpPostResponse, StreamableHttpError<Self::Error>> {
-        self.post_message_with_max_sse_event_size(
+        self.post_message_with_response_limits(
             uri,
             message,
             session_id,
             auth_token,
             custom_headers,
-            DEFAULT_MAX_SSE_EVENT_SIZE,
+            StreamableHttpResponseLimits::default(),
         )
         .await
     }
@@ -195,6 +196,29 @@ impl StreamableHttpClient for reqwest::Client {
         auth_token: Option<String>,
         custom_headers: HashMap<HeaderName, HeaderValue>,
         max_sse_event_size: usize,
+    ) -> Result<StreamableHttpPostResponse, StreamableHttpError<Self::Error>> {
+        self.post_message_with_response_limits(
+            uri,
+            message,
+            session_id,
+            auth_token,
+            custom_headers,
+            StreamableHttpResponseLimits {
+                max_sse_event_size,
+                ..Default::default()
+            },
+        )
+        .await
+    }
+
+    async fn post_message_with_response_limits(
+        &self,
+        uri: Arc<str>,
+        message: ClientJsonRpcMessage,
+        session_id: Option<Arc<str>>,
+        auth_token: Option<String>,
+        custom_headers: HashMap<HeaderName, HeaderValue>,
+        limits: StreamableHttpResponseLimits,
     ) -> Result<StreamableHttpPostResponse, StreamableHttpError<Self::Error>> {
         let mut request = self
             .post(uri.as_ref())
@@ -276,10 +300,17 @@ impl StreamableHttpClient for reqwest::Client {
         // Non-success responses may carry valid JSON-RPC error payloads that
         // should be surfaced as McpError rather than lost in TransportSend.
         if !status.is_success() {
-            let body = response
-                .text()
-                .await
-                .unwrap_or_else(|_| "<failed to read response body>".to_owned());
+            let body = match read_bounded_body(
+                response.bytes_stream(),
+                content_length,
+                limits.max_error_response_size,
+            )
+            .await
+            {
+                Ok(body) => String::from_utf8_lossy(&body).into_owned(),
+                Err(StreamableHttpError::Client(_)) => "<failed to read response body>".to_owned(),
+                Err(error) => return Err(error),
+            };
             if content_type
                 .as_deref()
                 .is_some_and(|ct| ct.as_bytes().starts_with(JSON_MIME_TYPE.as_bytes()))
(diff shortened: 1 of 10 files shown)


crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
```diff
@@ -276,10 +300,17 @@ impl StreamableHttpClient for reqwest::Client {
         // Non-success responses may carry valid JSON-RPC error payloads that
         // should be surfaced as McpError rather than lost in TransportSend.
         if !status.is_success() {
-            let body = response
-                .text()
-                .await
-                .unwrap_or_else(|_| "<failed to read response body>".to_owned());
+            let body = match read_bounded_body(
```
DaleSeo: Have you considered using `read_truncated_body` instead? Without it, a legacy server with a 4xx page larger than 64 KiB never reaches `legacy_discover_response`. That means a client that used to fall back to the `initialize` handshake now fails to connect entirely. Since we stop reading the stream at the limit either way, rejecting it does not reduce the memory limit.

crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
```diff
@@ -276,10 +300,17 @@ impl StreamableHttpClient for reqwest::Client {
...
+                response.bytes_stream(),
+                content_length,
+                limits.max_error_response_size,
```
DaleSeo: The same payload gets 16 MiB with a `200` response and 64 KiB with a `400`. (see line 316) Is this intentional?

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -369,12 +373,40 @@ pub(super) fn legacy_discover_response(
...
+    pub max_json_response_size: usize,
+    /// Maximum buffered non-success HTTP response body.
+    pub max_error_response_size: usize,
+}
+
+impl Default for StreamableHttpResponseLimits {
+    fn default() -> Self {
+        Self {
+            max_sse_event_size: DEFAULT_MAX_SSE_EVENT_SIZE,
+            max_json_response_size: 16 * 1024 * 1024,
```
DaleSeo: I'd raise the default for base64-encoded tool results.

```suggestion
            max_json_response_size: 64 * 1024 * 1024,
```

crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
GTanger: You're right: rejecting a diagnostic page broke legacy discovery without improving the buffering bound. Fixed in a94edbcb9f1cb50b9a5e4a3deae58cab23d8f494.

Both built-in backends now use `read_truncated_body` for non-JSON error pages. It retains at most `max_error_response_size` bytes and stops immediately at that limit, without polling another chunk or waiting for EOF. The existing discovery fallback eligibility checks are unchanged.

The regression tests now complete `server/discover` -> `initialize` -> `notifications/initialized` within two seconds for a >64 KiB page, both fixed-length and unfinished chunked. Unix socket coverage also verifies fallback from an unfinished stream whose prefix is exactly the limit. Auth/session precedence and ineligible 401/403/500 responses remain covered.

JSON error responses are handled separately: the complete bounded body is parsed before any diagnostic truncation, using the same JSON limit as HTTP 200. An oversized JSON protocol message still fails rather than parsing a prefix or triggering fallback. I've corrected the PR description accordingly.

crates/rmcp/src/transport/common/reqwest/streamable_http_client.rs
GTanger: That status-based split was a mistake for JSON-RPC errors. Fixed in a94edbcb9f1cb50b9a5e4a3deae58cab23d8f494: `application/json` uses `max_json_response_size` on both success and failure statuses, with a 64 MiB default. `max_error_response_size` now only limits diagnostic text.

The full JSON body is bounded and parsed before truncating diagnostics; prefixes are never parsed as complete messages. Tests send the identical JSON-RPC error on HTTP 200 and 400 with fixed/chunked bodies at N-1, N, and N+1, including an independent diagnostic limit of zero.

This applies to both built-in backends. Unix socket HTTP JSON-RPC errors now also surface as `Json` before synthetic discovery fallback, matching reqwest; that backend behavior change is documented and tested. All 686 selected tests and focused strict Clippy pass locally.

crates/rmcp/src/transport/streamable_http_client.rs
GTanger: Applied in a94edbcb9f1cb50b9a5e4a3deae58cab23d8f494: the JSON default is now 64 MiB, including JSON-RPC errors carried by non-2xx responses. The independent SSE-event default remains 16 MiB, and the diagnostic prefix remains 64 KiB.

Added a regression using the existing POST API that accepts a tool result containing 17 MiB of base64 data, plus assertions for all three defaults. Callers can still configure the limits independently; this bounds accumulated body bytes, not total process memory.

PR #1248, opened 2026-09-07
fix(server): keep discover lifecycle bootstrap-neutral

faisal-shah:
## Summary

- keep metadata-bearing `server/discover` requests bootstrap-neutral
- select the classic lifecycle on a subsequent `initialize`
- select the inline lifecycle only on the first valid ordinary metadata-bearing request
- leave malformed and unsupported bootstrap requests recoverable
- enqueue the lifecycle-selecting inline request into the normal service loop so bidirectional handlers remain safe

## Verification

- `cargo test --package rmcp --features "client transport-io macros schemars"`
- `cargo clippy --package rmcp --features "client transport-io macros schemars" --lib --test test_server_initialization --test test_stateless_server_requests --test test_subscriptions -- -D warnings`

### DIFF
--- a/crates/rmcp/src/service/server.rs
+++ b/crates/rmcp/src/service/server.rs
@@ -559,9 +559,14 @@ where
     let mut transport = transport.into_transport();
     let id_provider = <Arc<AtomicU32RequestIdProvider>>::default();
 
-    // Get initialize request; the MCP spec permits ping before initialize.
+    let (peer, peer_rx) = Peer::new(id_provider, None);
+
+    // Select the lifecycle only after an initialize request or the first valid
+    // non-discover request with complete inline metadata. A discover request is
+    // a bootstrap probe: respond to it, but remain open to either lifecycle.
+    // The MCP spec also permits ping before initialize.
     // See: https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization
-    let (request, id) = loop {
+    let (initialize_request, id) = loop {
         let msg = expect_next_message(&mut transport, "initialize request").await?;
         match msg {
             ClientJsonRpcMessage::Request(req)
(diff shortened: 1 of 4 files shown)


crates/rmcp/src/service/server.rs
```diff
@@ -580,60 +585,96 @@ where
...
+                        let response = match service.handle_request(request, context).await {
+                            Ok(result) => ServerJsonRpcMessage::response(result, id),
+                            Err(error) => ServerJsonRpcMessage::error(error, Some(id)),
+                        };
+                        transport.send(response).await.map_err(|error| {
+                            ServerInitializeError::transport::<T>(
+                                error,
+                                "sending bootstrap request response",
+                            )
+                        })?;
```
DaleSeo, line 664: The PR description calls out enqueueing the inline request into the service loop "so bidirectional handlers remain safe", but the discover branch still awaits `service.handle_request` directly while `peer_rx` has no one draining it.

crates/rmcp/src/service/server.rs
```diff
@@ -580,60 +585,96 @@ where
...
+                                })?;
+                            continue;
+                        }
+
+                        let is_discover = matches!(&request, ClientRequest::DiscoverRequest(_));
+                        if !is_discover {
+                            let requested_version = request
+                                .get_meta()
+                                .protocol_version()
+                                .expect("complete inline metadata has a protocol version");
```
DaleSeo, line 617: Since a panic here can take down the server's accept path based on client-controlled input, would you consider getting the version and treating `None` as a missing key? That way, the two checks can't drift apart.

crates/rmcp/src/service/server.rs
```diff
@@ -580,60 +585,96 @@ where
...
+                                    missing_request_metadata_error(&missing_metadata),
+                                    Some(id),
+                                ))
+                                .await
+                                .map_err(|error| {
+                                    ServerInitializeError::transport::<T>(
+                                        error,
+                                        "sending pre-init metadata error response",
+                                    )
+                                })?;
```
DaleSeo, line 608: Is a bound on pre-lifecycle attempts something you'd want here, or is that better left to the transport layer that owns the connection?

PR #1253, opened 2026-09-09
fix(sse): saturate exponential backoff to prevent u32 overflow

### LINKED ISSUE #1198: fix(sse): saturate exponential reconnect backoff

### LINKED ISSUE #1242: Empty cacheScope on tools/list drops every tool

yunaremaia:
## Summary

Fixes #1198 and #1242.

### Fix 1: Saturate exponential backoff (Fixes #1198)

ExponentialBackoff::retry computed 2u32.pow(current_times) which overflows when current_times >= 32 — panics in debug builds, wraps in release.

Saturates the multiplier at u32::MAX once current_times >= 32.

### Fix 2: Tolerate empty cacheScope (Fixes #1242)

Some hosted MCP servers send cacheScope: "" on tools/list and resources/read results. SEP-2549 only allows "public", "private", or omitting the field, so serde rejects the empty string. Because ServerResult is an untagged enum, that failure does not stay on ListToolsResult — the payload falls through to CustomResult, and the client ends up with no tools.

Add deserialize_cache_scope that treats empty string as None (absent) instead of erroring. Also handles the non-optional case (DiscoverResult.cache_scope) by defaulting to Public.

## Changes

- crates/rmcp/src/transport/common/client_side_sse.rs: replace 2u32.pow(current_times as u32) with saturated multiplier
- crates/rmcp/src/model.rs: add deserialize_cache_scope and deserialize_cache_scope_non_optional, apply to cache_scope fields
- Add test exponential_backoff_saturates_at_high_retry_count

## Test plan

- [x] cargo test -p rmcp --lib --features client-side-sse exponential_backoff passes
- [x] Test verifies normal case (2^5 = 32s), boundary (32 → u32::MAX), and high (100 → u32::MAX)
- [x] cargo check -p rmcp passes

### DIFF
--- a/crates/rmcp/src/model.rs
+++ b/crates/rmcp/src/model.rs
@@ -1230,6 +1230,7 @@ pub struct DiscoverResult {
     /// How long clients may consider this response fresh, in milliseconds.
     pub ttl_ms: u64,
     /// Whether the cached result may be shared across authorization contexts.
+    #[serde(default, deserialize_with = "deserialize_cache_scope_non_optional")]
     pub cache_scope: CacheScope,
     /// Protocol-level response metadata.
     #[serde(rename = "_meta", skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


=
DaleSeo, line 1: What is this for?

yunaremaia:
Closing: upstream already merged #1231 (ump45nose) with a better fix using  +  on Duration, which also caps the delay. This PR is now redundant. Thanks!

PR #1255, opened 2026-09-10
fix(auth): retain client secrets across credential restoration

### LINKED ISSUE #784: Support Oauth client secret
I'm trying to connect to an MCP server that uses DCR and returns an Oauth client ID and client secret. The auth works fine, but oauth refresh fails due to missing client_secret. Error: invalid_client 

Support storing and then using client secret in refresh. 

Setting up Oauth myself. 

**Additional context**

LSGoCards7:
Fixes #784.

A client registered with a secret can authorize successfully, but restoring its saved credentials reconstructs a client without that secret. Providers requiring client authentication then reject refresh after access-token expiry.

Persist an optional `oauth2::ClientSecret` in `StoredCredentials`, retain it on the authorization manager, and restore it after the existing issuer check. Preserve it during authorization and refresh saves. Add typed accessors for consumers that manage their own credential persistence, including Codex. Keep the existing constructor and `get_credentials()` signature; `StoredCredentials` is already non-exhaustive. Old records deserialize without a secret and public clients remain supported. Debug output redacts the field; empty secrets stay public.

Regression coverage serializes credentials between new manager instances and requires both `client_secret_basic` and `client_secret_post` on two successive refreshes. It also covers public/legacy clients, rotated and omitted replacement refresh tokens, DCR/CIMD issuer changes, serialization, empty secrets, and redaction. Existing refresh locks, client binding, resource handling, and save-before-return behavior remain intact.

Validation on base `3e636cab26c013eca5131103c03d20237f12c4df` (RMCP 3.3.0):

- `cargo +nightly fmt --all`: passed.
- `cargo +1.96.0 test -p rmcp --all-features transport::auth`: 216 passed; 357 other library tests filtered out.
- `cargo +1.96.0 clippy -p rmcp --all-targets --all-features -- -D warnings`: passed.
- `git diff --check`: passed. No dependency or lockfile changes.

Related consumer report: openai/codex#43216. An earlier isolated Codex 0.153.4/RMCP 3.1.3 candidate carrying this repair passed Supabase's natural access-token expiry, authenticated refresh with token rotation, an allowlisted read, and another process restart on 2026-09-10. That live result applies to the earlier candidate; this rebased SDK patch has the automated coverage above. No daily-client or released-fix claim is made.

### DIFF
--- a/crates/rmcp/src/transport/auth.rs
+++ b/crates/rmcp/src/transport/auth.rs
@@ -224,6 +224,9 @@ const DEFAULT_APPLICATION_TYPE: &str = "native";
 #[non_exhaustive]
 pub struct StoredCredentials {
     pub client_id: String,
+    /// Client authentication material required for token refresh after restart.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub client_secret: Option<ClientSecret>,
     pub token_response: Option<OAuthTokenResponse>,
     #[serde(default)]
     pub granted_scopes: Vec<String>,
@@ -237,6 +240,10 @@ impl std::fmt::Debug for StoredCredentials {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("StoredCredentials")
             .field("client_id", &self.client_id)
+            .field(
+                "client_secret",
+                &self.client_secret.as_ref().map(|_| "[REDACTED]"),
+            )
             .field(
                 "token_response",
                 &self.token_response.as_ref().map(|_| "[REDACTED]"),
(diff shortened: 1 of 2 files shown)


crates/rmcp/src/transport/auth.rs
```diff
@@ -1464,7 +1480,9 @@ impl AuthorizationManager {
                 }
             }
 
-            self.configure_client_id(&stored.client_id)?;
+            let mut config = OAuthClientConfig::new(&stored.client_id, self.base_url.to_string());
+            config.client_secret = stored.client_secret.map(|secret| secret.secret().clone());
+            self.configure_client(config)?;
```
DaleSeo, line 1485: Was dropping it here intentional, or would threading the secret through that path keep both restore branches in step?

crates/rmcp/src/transport/auth.rs
```diff
@@ -3697,6 +3738,7 @@ impl OAuthState {
 
             let stored = StoredCredentials {
                 client_id: client_id.to_string(),
+                client_secret: None,
```
DaleSeo, line 3741: How is a consumer that uses `get_credentials` and `set_credentials` instead of a `CredentialStore` expected to restore the secret on the next run?

PR #1256, opened 2026-09-10
fix(transport): stop keeping a session and GET stream at 2026-07-28

### LINKED ISSUE #1108: SEP-2260 follow-up: skip session tracking and the standalone GET stream at protocol >= 2026-07-28
Follow-up to #1033 / PR #1055.

PR #1055 makes the client *reject* SEP-2260-restricted requests that arrive on the standalone GET stream: the streamable HTTP transport marks inbound requests with an `InboundStreamOrigin`, the service layer turns that into a `PeerRequestAssociation`, and `RoleClient::enforce_peer_request_association` answers `-32602` for `Unassociated`.

Rejecting is the right behavior for a server that sends such a request. But at protocol `2026-07-28` the client shouldn't have the stream in the first place. SEP-2567 removes sessions and the standalone GET endpoint at that version, so an `Mcp-Session-Id` and a GET stream are both artifacts of a pre-`2026-07-28` server shape. Keeping them open at `2026-07-28` means maintaining an inbound path whose only purpose is to be refused.

The modern path already gets this right. A `server/discover` startup sets `session_id` to `None` unconditionally (`streamable_http_client.rs`, ~L892), and there's a `debug_assert!` that discover bootstrap creates no session state (~L1054). What's left is the legacy-shaped paths, where a session id survives regardless of the version that was ultimately negotiated. `spawn_common_stream` has three call sites and all three are guarded only by `if let Some(session_id)`, never by `negotiated_version`:

- **Legacy startup** (~L969). A server that answers legacy `initialize` with both a session id and `protocolVersion: 2026-07-28` gets a GET stream. This is the shape the e2e test in #1055 scripts deliberately.
- **Fallback `initialize`** after `server/discover` fails (~L1382, on the `initialized` notification). Same shape, reached through `ClientLifecycleMode::Auto`.
- **Session re-establishment** after an expired-session 404 (~L1218). Re-spawns on the new session id without reconsulting the version.

Work: gate session tracking and `spawn_common_stream` on `negotiated_version < ProtocolVersion::STANDARD_HEADERS`, so that at `2026-07-28` and later no session id is retained and no GET stream is opened. `PeerRequestAssociation::Unassociated` then becomes unreachable over streamable HTTP rather than merely rejected, and the enforcement from #1055 stays as the defense for anything that still slips through.

Two open questions worth settling in the same change:

- What to do when a server returns a session id *and* negotiates `2026-07-28`. Dropping the id silently is the least disruptive, but the server plainly expects it echoed back, so failing startup may be more honest than half-speaking its dialect.
- Whether session cleanup (`SessionCleanupInfo`, the `DELETE` on shutdown) should also be skipped, or kept so a session the server did create still gets torn down.

This also bears on #863 (expose the negotiated `Mcp-Session-Id`): if there is deliberately no session at `2026-07-28`, the accessor is a legacy-only affordance, and that's probably worth saying there either way.

Not a conformance gap — the receive-side check is a SHOULD and #1055 covers the scenarios — so this is hardening rather than a fix.

LizunovSergey:
Closes #1108.

## What was wrong

SEP-2567 removes sessions and the standalone GET endpoint at `2026-07-28`, so an `Mcp-Session-Id` and a GET stream are both artifacts of a pre-`2026-07-28` server shape. The modern path already gets this right — a `server/discover` bootstrap sets `session_id` to `None` unconditionally. The legacy-shaped paths did not: all three `spawn_common_stream` call sites were guarded only on `if let Some(session_id)`, never on `negotiated_version`.

- **Legacy startup.** A server that answers legacy `initialize` with both a session id and `protocolVersion: 2026-07-28` got a GET stream.
- **Fallback `initialize`** after `server/discover` fails, on the `initialized` notification.
- **Session re-establishment** after an expired-session 404, which respawned on the new id without reconsulting the version.

## How

The gate goes where a session id is *adopted*, not at each spawn. `session_id_for_version` drops the id once the negotiated version has no sessions, and because all three spawn sites are already guarded on the id being present, they are left with nothing to open — no change needed at any of them. The id also stops being echoed on requests.

That makes `PeerRequestAssociation::Unassociated` unreachable over streamable HTTP rather than merely rejected, and the receive-side enforcement from #1055 stays as the defence for anything that still slips through.

## The two open questions from the issue

Both needed a decision to write the change at all, so here is what I picked and why. Happy to move either way.

**1. Server returns a session id *and* negotiates `2026-07-28`.** The issue weighed dropping it silently against failing startup. I went with dropping it **plus a `tracing::warn!`**, rather than either extreme. Failing would break clients against servers that work today over a SHOULD-level hardening, which seemed too blunt; dropping in complete silence hides a real server bug from whoever has to debug it. A warning naming the negotiated version puts the evidence in the log without changing whether the client connects.

**2. Should session cleanup be skipped too?** I kept it. `SessionCleanupInfo` is still built from the id the server sent, so the shutdown `DELETE` still fires and a session the server really did create gets torn down. Skipping cleanup would leak server-side state as the price of a client-side conformance fix, which felt like the wrong trade. The ordering in the diff is deliberate: the cleanup record is built first, then the id is dropped for request and stream purposes.

I did not touch #863; if there is deliberately no session at `2026-07-28`, that accessor is a legacy-only affordance, but saying so belongs in that issue.

## Tests

New `tests/test_streamable_http_sessionless_version.rs`, a scripted server that answers legacy `initialize` with an `Mcp-Session-Id` header and a configurable `protocolVersion`, recording every request as `(HTTP method, JSON-RPC method, session header)`:

- `modern_version_drops_the_session_and_opens_no_stream` — at `2026-07-28`, no GET request arrives at all and no post-handshake POST carries the header.
- `legacy_version_keeps_the_session_and_opens_the_stream` — at `2025-11-25`, exactly one GET arrives, it carries the session id, and the id is echoed afterwards.

Both are witnesses, checked by mutation. Removing the three gating lines fails `modern_version_drops_the_session_and_opens_no_stream` while `legacy_version_keeps_the_session_and_opens_the_stream` still passes — so the legacy assertion is genuinely pinning the unchanged path rather than passing by accident.

Also green, unchanged: `test_streamable_http_stale_session` (4, and it covers the recovery path this touches), `test_client_lifecycle_modes` (15), `test_server_discover_http` (12), `test_streamable_http_priming` (5), `test_discover_http_client_startup` (3). `cargo +nightly fmt --all --check` clean; `cargo clippy -- -D warnings` clean on the lib and on the new test.

### DIFF
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -90,6 +90,37 @@ fn request_version_headers(
     (version, headers)
 }
 
+/// Decides whether a session id survives version negotiation.
+///
+/// SEP-2567 removes sessions and the standalone GET endpoint at
+/// [`ProtocolVersion::STANDARD_HEADERS`], so at that version an `Mcp-Session-Id` and a GET
+/// stream are both artifacts of a pre-`2026-07-28` server shape. A legacy-shaped handshake
+/// can still answer with a session id while negotiating that version; the id is dropped
+/// rather than echoed, which also leaves every `spawn_common_stream` call site — all three
+/// of which are guarded on a session id being present — with no stream to open.
+///
+/// Dropping is deliberate rather than fatal: refusing to start would break clients against
+/// servers that work today, and the receive-side enforcement added for SEP-2260 still
+/// rejects anything that reaches the client over a stream it should not have. The caller
+/// keeps the original id for the shutdown `DELETE`, so a session the server really did
+/// create is still torn down.
+fn session_id_for_version(
+    session_id: Option<Arc<str>>,
+    negotiated_version: &ProtocolVersion,
+) -> Option<Arc<str>> {
+    if negotiated_version < &ProtocolVersion::STANDARD_HEADERS {
+        return session_id;
+    }
+    if session_id.is_some() {
+        tracing::warn!(
+            version = negotiated_version.as_str(),
+            "server returned an Mcp-Session-Id while negotiating a version that has no sessions; \
+             the id will not be sent on requests and no standalone GET stream will be opened"
+        );
+    }
+    None
+}
+
 fn cache_tools_from_response(
     cache: &mut HashMap<String, Arc<JsonObject>>,
     message: &mut ServerJsonRpcMessage,
@@ -1133,6 +1164,7 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
             auth_header: config.auth_header.clone(),
             protocol_headers: protocol_headers.clone(),
         });
+        session_id = session_id_for_version(session_id, &negotiated_version);
 
         context.send_to_handler(message).await?;
         if is_legacy_startup {
@@ -1261,6 +1293,7 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
                             auth_header: config.auth_header.clone(),
                             protocol_headers: protocol_headers.clone(),
                         });
+                        session_id = session_id_for_version(session_id, &negotiated_version);
                         // Do not send controls queued during recovery to the new session.
                         context.advance_control_generation();
                         session_cancellation = CancellationToken::new();
@@ -1517,6 +1550,7 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
                                 auth_header: config.auth_header.clone(),
                                 protocol_headers: protocol_headers.clone(),
                             });
+                        session_id = session_id_for_version(session_id, &negotiated_version);
                         context.send_to_handler(initialize_response).await?;
                         awaiting_fallback_initialized = true;
                         continue;
--- a/crates/rmcp/src/transport/streamable_http_client.rs
+++ b/crates/rmcp/src/transport/streamable_http_client.rs
@@ -121,6 +121,20 @@ fn session_id_for_version(
     None
 }
 
+/// The session established by [`StreamableHttpClientWorker::perform_reinitialization`].
+///
+/// The two ids are the same id at different stages of [`session_id_for_version`], and they
+/// are deliberately not interchangeable: `cleanup_session_id` is what the server sent, kept
+/// so the shutdown `DELETE` still tears down a session the server really created, while
+/// `session_id` is what may be echoed on requests and used to open a standalone GET stream.
+/// At [`ProtocolVersion::STANDARD_HEADERS`] the latter is always `None`.
+struct Reinitialized {
+    session_id: Option<Arc<str>>,
+    cleanup_session_id: Option<Arc<str>>,
+    negotiated_version: ProtocolVersion,
+    protocol_headers: HashMap<HeaderName, HeaderValue>,
+}
+
 fn cache_tools_from_response(
     cache: &mut HashMap<String, Arc<JsonObject>>,
     message: &mut ServerJsonRpcMessage,
@@ -981,9 +995,13 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
     /// future remains `Send` without requiring `C: Sync`).  POSTs the saved
     /// initialize request without a session ID, extracts the new session ID and
     /// protocol version, sends `notifications/initialized`, and returns the new
-    /// `(session_id, protocol_headers)` pair.  The init result message is **not**
+    /// session in a [`Reinitialized`].  The init result message is **not**
     /// forwarded to the handler because the handler already processed the original
     /// initialization.
+    ///
+    /// The handshake completes here, so [`session_id_for_version`] is applied here too:
+    /// the `initialized` notification is part of the new session and must not echo an id
+    /// the negotiated version has no sessions for.
     async fn perform_reinitialization(
         client: C,
         saved_init_request: ClientJsonRpcMessage,
@@ -991,14 +1009,7 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
         auth_header: Option<String>,
         custom_headers: HashMap<HeaderName, HeaderValue>,
         max_sse_event_size: usize,
-    ) -> Result<
-        (
-            Option<Arc<str>>,
-            ProtocolVersion,
-            HashMap<HeaderName, HeaderValue>,
-        ),
-        StreamableHttpError<C::Error>,
-    > {
+    ) -> Result<Reinitialized, StreamableHttpError<C::Error>> {
         let (init_msg, new_session_id_str) = client
             .post_message_with_max_sse_event_size(
                 uri.clone(),
@@ -1012,10 +1023,13 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
             .expect_initialized::<C::Error>()
             .await?;
 
-        let new_session_id: Option<Arc<str>> = new_session_id_str.map(|s| Arc::from(s.as_str()));
+        let cleanup_session_id: Option<Arc<str>> =
+            new_session_id_str.map(|s| Arc::from(s.as_str()));
 
         let (negotiated_version, new_protocol_headers) =
             negotiate_version_headers(&init_msg, custom_headers);
+        let new_session_id =
+            session_id_for_version(cleanup_session_id.clone(), &negotiated_version);
 
         let initialized_notification = ClientJsonRpcMessage::notification(
             ClientNotification::InitializedNotification(InitializedNotification {
@@ -1042,7 +1056,12 @@ impl<C: StreamableHttpClient> StreamableHttpClientWorker<C> {
             .await?
             .expect_accepted_or_json::<C::Error>()?;
 
-        Ok((new_session_id, negotiated_version, new_protocol_headers))
+        Ok(Reinitialized {
+            session_id: new_session_id,
+            cleanup_session_id,
+            negotiated_version,
+            protocol_headers: new_protocol_headers,
+        })
     }
 }
 
@@ -1267,7 +1286,12 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
                     ) => result.unwrap_or(Err(StreamableHttpError::SessionRecoveryTimeout)),
                 };
                 match recovery {
-                    Ok((new_session_id, new_version, new_headers)) => {
+                    Ok(Reinitialized {
+                        session_id: new_session_id,
+                        cleanup_session_id,
+                        negotiated_version: new_version,
+                        protocol_headers: new_headers,
+                    }) => {
                         streams.abort_all();
                         while streams.join_next().await.is_some() {}
                         request_stream_cancellations.clear();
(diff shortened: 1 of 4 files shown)


LizunovSergey:
CI went red on `test_sep_2260_stream_enforcement::restricted_request_on_get_stream_rejected_while_unrelated_request_in_flight`, and it was a real consequence of this change rather than flake. Pushed a fix; here is the reasoning, since it touches a test from #1033 rather than my own.

**Why it broke.** That harness reached an unassociated stream through the standalone GET stream, which required the client to hold a session while negotiating `2026-07-28`. Its module doc says so explicitly — "rmcp's client tolerates the session id and opens the standalone GET stream — so this is the reachable path where the client has BOTH a GET stream and strict SEP-2260 enforcement." This PR is precisely what removes that tolerance, so the scenario stopped being reachable and the test sat waiting for a rejection that no longer had a stream to arrive on. It is also why only `Run Tests (no local feature)` caught it: the file is `#![cfg(not(feature = "local"))]`, so it is the one job that compiles it.

**Why not just negotiate a legacy version.** That was my first instinct and it is wrong — `enforce_peer_request_association` only tightens at `>= V_2026_07_28`, so dropping to `2025-11-25` would have switched off the very check under test while leaving it green.

**What I did instead.** Kept the protocol at `2026-07-28` and moved the unassociated stream to the SSE body of a post-startup notification POST. A POST carrying no request id gets `InboundStreamOrigin::Unassociated`, so the stream is unassociated by construction, at a strict version, with an unrelated `tools/list` still in flight — the #1033 scenario intact, minus the session. The scripted server also stops minting a session id it is not entitled to at that version, so it now models a conforming server. One wrinkle worth recording: `notifications/initialized` cannot be the carrier, because startup treats a stream there as fatal (`expect accepted or json, got Sse(None)`); it has to be a notification sent after the handshake.

**Evidence.** Both tests in the file pass. The renamed test is still a witness by mutation — disabling the `strict` gate in `enforce_peer_request_association` fails it while the positive twin (`restricted_request_on_originating_post_stream_is_dispatched`) keeps passing, so it is pinning the enforcement and not the plumbing. The whole `no local feature` feature set is green locally (the same set CI computes, 0 failures), and `cargo +nightly fmt --all --check` and `cargo clippy --all-targets -- -D warnings` are clean.

If you would rather keep that test pinned to the legacy shape and assert the GET stream is gone separately, say the word and I will split it that way instead.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -1261,6 +1293,7 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
                             auth_header: config.auth_header.clone(),
                             protocol_headers: protocol_headers.clone(),
                         });
+                        session_id = session_id_for_version(session_id, &negotiated_version);
```
DaleSeo: This gate runs after `perform_reinitialization` has already POSTed `notifications/initialized` with `new_session_id`. So during recovery, if it negotiates `2026-07-28`, that request still echoes the ID. Should the gate be applied inside `perform_reinitialization`?

LizunovSergey: Good catch — you're right, and it was a real hole. Fixed in d36ce25.

`perform_reinitialization` POSTs `notifications/initialized` itself, so the handshake is already complete by the time it returns and the caller's gate was too late for that one request. The startup path doesn't have this problem because its gate sits above its own `initialized` POST; the recovery path just wasn't matched to it.

So I moved the gate inside `perform_reinitialization`, directly after version negotiation and before the notification goes out. The fallback-`initialize` path was already fine — its gate precedes the notification, which goes through the normal message loop.

The one wrinkle is session cleanup. The caller builds `SessionCleanupInfo` from the id *before* gating on purpose, so the shutdown `DELETE` still tears down a session the server really created — if the gate simply moved inward and returned `None`, that teardown would be lost. So the function now returns a small `Reinitialized` struct carrying both the id as sent (`cleanup_session_id`, for the `DELETE`) and the gated one (`session_id`, for requests and streams). A struct rather than a 4-tuple because two `Option<Arc<str>>` positions next to each other are easy to swap by accident.

Two new tests in `test_streamable_http_sessionless_version.rs`, both driven through a real expired-session 404 so the recovery path actually runs:

- `replacement_handshake_at_a_modern_version_drops_the_session_too` — starts on a `2025-11-25` session, server 404s a `tools/list`, replacement handshake answers with a *new* session id and `2026-07-28`. Asserts the `initialized` notification carries no session header, that nothing afterwards does, that no GET stream opens on the replacement session, and that the shutdown `DELETE` still carries the id.
- `replacement_handshake_at_a_legacy_version_keeps_its_session` — same route ending at `2025-11-25`, pinning that the legacy recovery path is unchanged: the id is echoed and each session gets its GET stream.

Both checked by mutation. Restoring the old ordering fails the first on exactly your scenario (`left: Some("session-issued-by-the-replacement-handshake"), right: None`) while the other three still pass; separately, building cleanup from the gated id instead fails the `DELETE` assertion. The handler also records the teardown `DELETE` now instead of trying to parse its empty body as JSON, which was panicking in the server task.

Green and unchanged: `test_streamable_http_stale_session` (4), `test_client_lifecycle_modes` (15), `test_server_discover_http` (12), `test_streamable_http_priming` (5), `test_discover_http_client_startup` (3). `cargo +nightly fmt --all --check` clean; clippy clean on the lib and the test target.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -1517,6 +1575,7 @@ impl<C: StreamableHttpClient> Worker for StreamableHttpClientWorker<C> {
                                 auth_header: config.auth_header.clone(),
                                 protocol_headers: protocol_headers.clone(),
                             });
+                        session_id = session_id_for_version(session_id, &negotiated_version);
```
DaleSeo, line 1578: Starting the client in `ClientLifecycleMode::Auto` with the existing recorder takes the fallback path and fails once this line is removed. Would you parameterize the startup tests by lifecycle mode, or add a dedicated pair of fallback tests?

crates/rmcp/src/transport/streamable_http_client.rs
DaleSeo, line 1599: With a per-request `_meta.protocolVersion`, `negotiated_version` gets updated here without rechecking `session_id`.

crates/rmcp/tests/test_streamable_http_sessionless_version.rs
```diff
@@ -0,0 +1,420 @@
...
+        .and_then(|value| value.to_str().ok())
+        .map(str::to_owned);
+
+    if method == axum::http::Method::GET {
+        state.seen.lock().expect("recorder poisoned").push((
+            "GET".to_owned(),
+            "-".to_owned(),
+            session,
+        ));
+        // Hang so the client keeps the stream if it opens one; the test cancels it.
```
DaleSeo, line 151: The comment says the handler hangs so the client keeps the stream open, but this branch returns `405` immediately.

crates/rmcp/src/transport/streamable_http_client.rs
```diff
@@ -90,6 +90,51 @@ fn request_version_headers(
...
 }
 
+/// Decides whether a session id survives version negotiation.
+///
+/// SEP-2567 removes sessions and the standalone GET endpoint at
+/// [`ProtocolVersion::STANDARD_HEADERS`], so at that version an `Mcp-Session-Id` and a GET
+/// stream are both artifacts of a pre-`2026-07-28` server shape. A legacy-shaped handshake
+/// can still answer with a session id while negotiating that version; the id is dropped
+/// rather than echoed, which also leaves every `spawn_common_stream` call site — all three
+/// of which are guarded on a session id being present — with no stream to open.
```
DaleSeo, line 100:
```suggestion
/// rather than echoed, which also leaves every `spawn_common_stream` call site — each
/// guarded on a session id being present — with no stream to open.
```

LizunovSergey: Applied in ce40fb8 — and better without the count, since it stops being a thing to keep in sync with the call sites.

crates/rmcp/tests/test_streamable_http_sessionless_version.rs
LizunovSergey: Fixed in 04fd607. The comment was left over from an earlier version of the harness that did hold the stream open; the branch records the attempt and then refuses it the way a server without the endpoint does, which is all these tests read. Comment now says that.

crates/rmcp/src/transport/streamable_http_client.rs
LizunovSergey: Added a dedicated pair, and parameterized the helper to get them: `start_client` now takes the lifecycle mode, the two assertion bodies moved into `assert_session_dropped` / `assert_session_kept`, and the four startup tests are named per shape so a failure says which path broke.

`fallback_handshake_at_a_modern_version_drops_the_session_too` and `fallback_handshake_at_a_legacy_version_keeps_its_session` go through `Auto`; the scripted server already 404s `server/discover`, so startup falls through to the in-loop `initialize` — a different gate from the startup one, adopted at a different place in the worker.

They pin that gate rather than doubling up on the one already covered: removing only the fallback gate fails exactly `fallback_handshake_at_a_modern_version_drops_the_session_too` while `modern_version_drops_the_session_and_opens_no_stream` still passes (it fails on the GET stream — `left: [("GET", "-", Some("session-the-server-should-not-have-issued"))]`).

One small harness change came with it: `session_headers_after_handshake` now also skips `server/discover`, since that probe precedes the handshake and carries no id either way, so it is not evidence in either direction.

In 04fd607.

crates/rmcp/src/transport/streamable_http_client.rs
LizunovSergey: Right, and that one was reachable without any server misbehaviour at all — the client moves itself there. Fixed in ce40fb8.

The gate now sits directly under the `negotiated_version` assignment, above the POST that request builds, so the request that moves the version is the first one to go without the id rather than the last one to carry it. `session_cleanup_info` is built earlier from the id the server sent, so the shutdown `DELETE` is unaffected — same split as the startup and recovery paths.

Two things I decided rather than discovered, both worth stating:

**The gate is one-way.** A per-request version replaces `negotiated_version` wholesale in the existing code, so once it has moved to 2026-07-28 the transport is at a version with no sessions and the id is gone. A later request that declares a legacy version again does not get it back — it goes out sessionless, and a server that still has the session answers 404, which is the expired-session recovery path. Keeping the server-sent id in shadow state to survive a flip-flop seemed like state carried for a shape with no legitimate use.

**An already-open GET stream is left running.** It was opened while the session was still legitimate; aborting it mid-flight would drop whatever response is in transit on it, and the SEP-2260 receive-side enforcement still rejects anything it should not carry. Happy to tear it down instead if you would rather the invariant hold for streams too and not just for requests.

Tests in 04fd607: `a_request_that_moves_to_a_modern_version_drops_the_session` sends a control `tools/list` on the negotiated version first, so the recording shows the id being echoed and then not; it also asserts the teardown `DELETE` still carries it. `a_request_that_moves_to_another_legacy_version_keeps_the_session` moves to 2025-06-18 to pin that the gate does not over-fire on a move that stays legacy. Removing the gate fails the first on exactly your scenario (`left: Some("session-the-server-should-not-have-issued"), right: None`) and leaves the other seven passing.

PR #1262, opened 2026-09-12
fix(rmcp): route peer cancellation by lifecycle

### LINKED ISSUE #1251: Client ignores legacy reverse-request cancellation and may cancel an unrelated outbound request
## Summary

In a legacy MCP session, a server's `notifications/cancelled` does not cancel the matching server-originated request's `RequestContext.ct`. The `RoleClient` service loop removes an outbound response waiter instead of cancelling the inbound request token. If an unrelated outbound request has the same ID, its waiter can be cancelled instead.

Reproduced on current main at `302319861a4b5ab538f6aebf25befdc3c7dfe039`. The same routing is present in the released rmcp 3.2.0 source.

## Reproduction

1. Complete an `initialize` handshake selecting protocol version `2025-11-25`.
2. Have the client's `create_elicitation` handler signal that it has started, then await `context.ct.cancelled()`.
3. Send this request from the server:

```json
{"jsonrpc":"2.0","id":"elicitation-1","method":"elicitation/create","params":{"message":"Continue?","requestedSchema":{"type":"object","properties":{}}}}
```

4. After the handler starts, send:

```json
{"jsonrpc":"2.0","method":"notifications/cancelled","params":{"requestId":"elicitation-1"}}
```

5. Keep the connection open. The handler's token remains active. Sending a subsequent ping confirms the service can still process traffic.

### Runnable regression against unpatched main

The [regression file](https://github.com/aurokin/rust-sdk/blob/a389b20049bad9de1d165951366075b124793840/crates/rmcp/tests/test_cancelled_response.rs) contains a raw-peer reproduction. The following commands copy only that test file into a fresh checkout of unpatched upstream; no production source changes are applied:

```sh
git clone https://github.com/modelcontextprotocol/rust-sdk.git rmcp-cancel-repro
cd rmcp-cancel-repro
git switch --detach 302319861a4b5ab538f6aebf25befdc3c7dfe039
git fetch https://github.com/aurokin/rust-sdk.git a389b20049bad9de1d165951366075b124793840
git show a389b20049bad9de1d165951366075b124793840:crates/rmcp/tests/test_cancelled_response.rs \
  > crates/rmcp/tests/test_cancelled_response.rs
cargo test -p rmcp --test test_cancelled_response \
  --features client,elicitation,transport-io
```

On unpatched main, three tests pass and two fail:

- `peer_cancels_reverse_request_without_cancelling_outbound_request` detects that an unrelated outbound waiter is no longer pending after cancellation of an ID that does not identify an inbound request.
- `cancelled_reverse_request_stays_suppressed_during_eof_drain` times out waiting for the handler's cancellation event. This happens **before** the test sends EOF or closes the connection, so the failure is the missing token cancellation, not a disconnect-handling issue.

The reverse-request cases cover a normal legacy handshake, directly configured legacy peers, and direct peers without handshake information. They also cover equal IDs in opposite directions and distinct string/numeric IDs.

## Expected behavior

For legacy sessions, cancellation should remove and cancel the exact matching inbound request token. Unknown IDs must not affect an outbound request merely because it has the same ID.

The handler should observe `RequestContext.ct` without a connection shutdown. Unsent responses should be suppressed as recommended by the cancellation specification, while already-sent response races remain valid. The connection should remain usable.

## Root cause

In the [incoming notification branch](https://github.com/modelcontextprotocol/rust-sdk/blob/302319861a4b5ab538f6aebf25befdc3c7dfe039/crates/rmcp/src/service.rs#L1604-L1631), `R::IS_CLIENT` selects `local_responder_pool.remove(request_id)`. That pool tracks locally originated requests. Tokens for peer-originated requests are in `local_ct_pool`.

This is separate from #1193, which concerns cancellation of an outgoing Streamable HTTP request before its response stream starts, and #857, which concerned TCP disconnects.

## Proposed approach for discussion

Route incoming cancellation by the established lifecycle and request direction:

- Legacy peer cancellation cancels the exact matching inbound token.
- Modern server cancellation retires only an actual outbound `subscriptions/listen` request, including its response waiter. Removing outbound cleanup entirely would break subscription termination.
- Successful discovery retains modern lifecycle semantics even when it selects an older application version. Existing startup metadata can preserve that distinction.

Keep string and numeric IDs distinct, preserve uncancelled response draining during graceful shutdown, and apply normal peer-cancellation suppression during EOF draining. No dependency changes or outgoing server-teardown redesign are needed.

I have a candidate fix and regressions, but have not opened a PR. Does this lifecycle-aware split match the intended behavior? I'm opening the issue first for feedback, following `docs/CONTRIBUTE.MD`.

## Protocol references

- [Legacy cancellation](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation)
- [Modern cancellation](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/cancellation)
- [Lifecycle and application-version compatibility](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#backward-compatibility-with-initialization-based-versions)

aurokin:
Fixes #1251.

Implements the approach agreed to in https://github.com/modelcontextprotocol/rust-sdk/issues/1251#issuecomment-5641235524.

## Problem

In legacy sessions, incoming server `notifications/cancelled` messages remove an outbound client waiter instead of cancelling the server-originated request's `RequestContext.ct`. This leaves the handler running and can affect an unrelated request with the same ID.

Removing outbound cleanup entirely would break modern subscription termination.

## Fix

Route cancellation by the established lifecycle:
- Cancel the exact inbound token for legacy requests.
- Retire only an actual outbound `subscriptions/listen` waiter for modern server cancellation, independently of notification-channel registration.
- Reuse existing startup metadata so discovery retains modern semantics even when selecting an older application version.
- Keep string and numeric request IDs distinct.
- Suppress cancelled handler responses and errors during EOF draining while preserving uncancelled responses during graceful shutdown.

No dependency, wire-schema, or macro changes. Outgoing server teardown and same-direction ID reuse are outside this fix.

## Regression coverage

The tests cover string-ID reverse cancellation on a live connection, handler-token observation, response suppression, subsequent traffic, legacy initialization, direct peers, unknown IDs, and equal IDs in opposite directions.

The subscription regression inspects the actual waiter after an ordered ping. It covers modern discovery selecting either application version, ordinary outbound request immunity, and listen requests without a registered notification channel.

Drain regressions cover successful and failed cancelled handlers and graceful completion of uncancelled handlers. Existing RoleServer stdio coverage remains in place.

## Validation

The clean candidate passed:
- Targeted tests: 209 unit, 5 cancellation, 12 subscription, and 1 drain.
- Nightly formatting and focused Clippy.
- SemVer and public-API checks with default and non-local features.
- Full-range commitlint.

Full all-feature and non-local suites passed on the same production source before the final test-only error-response variant. That variant passed targeted validation, including against the published release with its original lockfile.

The same cancellation files also passed targeted validation after merging upstream `3023198` in a separate development branch. That merge is not included in this single-commit PR head. The regressions fail against unpatched upstream; an independent reproduction on `46db531` is recorded in #1251.

Full Clippy previously failed at an unchanged `clippy::question_mark` warning in `auth.rs`. The unchanged base reproduced the same warning with the same compiler and lockfile. No unrelated lint fix or suppression is included.

### DIFF
--- a/crates/rmcp/src/service.rs
+++ b/crates/rmcp/src/service.rs
@@ -136,6 +136,14 @@ pub trait ServiceRole: std::fmt::Debug + Send + Sync + 'static + Copy + Clone {
     fn peer_cancelled_params(_notification: &Self::PeerNot) -> Option<&CancelledNotificationParam> {
         None
     }
+    #[doc(hidden)]
+    fn peer_cancels_subscriptions(_peer: &Peer<Self>) -> bool {
+        false
+    }
+    #[doc(hidden)]
+    fn is_subscription_request(_request: &Self::Req) -> bool {
+        false
+    }
     /// Invalidate any response cache affected by an inbound peer notification.
     ///
     /// The serve loop calls this for every notification *before* subscription
@@ -522,6 +530,10 @@ impl ProgressNotificationToken for ServerNotification {
 }
 
 type Responder<T> = tokio::sync::oneshot::Sender<T>;
+struct PendingResponse<T> {
+    responder: Responder<T>,
+    is_subscription: bool,
+}
 type ProgressTimeoutWatchers = Arc<tokio::sync::RwLock<HashMap<ProgressToken, mpsc::Sender<()>>>>;
 type SubscriptionChannel<N> = (mpsc::Sender<N>, usize);
 type SubscriptionChannelMap<N> = HashMap<RequestId, SubscriptionChannel<N>>;
@@ -1348,7 +1360,7 @@ where
     }
 
     let mut local_responder_pool =
-        HashMap::<RequestId, Responder<Result<R::PeerResp, ServiceError>>>::new();
+        HashMap::<RequestId, PendingResponse<Result<R::PeerResp, ServiceError>>>::new();
     let mut local_ct_pool = HashMap::<RequestId, CancellationToken>::new();
     let shared_service = Arc::new(service);
     // for return
@@ -1445,7 +1457,7 @@ where
                 Event::SendTaskResult(SendTaskResult::Request { id, result }) => {
                     if let Err(e) = result
                         && let Some(responder) = local_responder_pool.remove(&id) {
-                            let _ = responder.send(Err(ServiceError::TransportSend(e)));
+                            let _ = responder.responder.send(Err(ServiceError::TransportSend(e)));
                         }
                 }
                 Event::SendTaskResult(SendTaskResult::Notification {
@@ -1463,7 +1475,7 @@ where
                         && let Some(request_id) = &param.request_id
                             && let Some(responder) = local_responder_pool.remove(request_id) {
                                 tracing::info!(id = %request_id, reason = param.reason, "cancelled");
-                                let _response_result = responder.send(Err(ServiceError::Cancelled {
+                                let _response_result = responder.responder.send(Err(ServiceError::Cancelled {
                                     reason: param.reason.clone(),
                                 }));
                             }
@@ -1500,7 +1512,10 @@ where
                     id,
                     responder,
                 }) => {
-                    local_responder_pool.insert(id.clone(), responder);
+                    local_responder_pool.insert(id.clone(), PendingResponse {
+                        responder,
+                        is_subscription: R::is_subscription_request(&request),
+                    });
                     let send = transport.send(JsonRpcMessage::request(request, id.clone()));
                     {
                         let id = id.clone();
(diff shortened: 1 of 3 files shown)


crates/rmcp/src/service.rs
```diff
@@ -1601,31 +1616,38 @@ where
...
+                                        .is_some_and(|pending| pending.is_subscription)
+                                        && let Some(pending) = local_responder_pool.remove(request_id)
                                     {
-                                        let _ = responder.send(Err(ServiceError::Cancelled {
+                                        let _ = pending.responder.send(Err(ServiceError::Cancelled {
                                             reason: cancelled.reason.clone(),
                                         }));
+                                    } else {
+                                        tracing::debug!(%request_id, "ignoring cancellation of unknown subscription");
+                                        continue;
```
DaleSeo, line 1635: This `continue` will drop a modern peer's cancellation for any non-subscription ID before it gets dispatched.

crates/rmcp/src/service.rs
```diff
@@ -1601,31 +1616,38 @@ where
...
-                                if R::IS_CLIENT {
-                                    if let Some(responder) =
-                                        local_responder_pool.remove(request_id)
+                            // Modern servers cancel listen requests; legacy peers cancel
+                            // requests they originated, even when both directions share an ID.
+                            if R::peer_cancels_subscriptions(&peer) {
+                                if let Some(request_id) = request_id.as_ref() {
+                                    if local_responder_pool.get(request_id)
+                                        .is_some_and(|pending| pending.is_subscription)
+                                        && let Some(pending) = local_responder_pool.remove(request_id)
```
DaleSeo, line 1628: The modern branch never checks `local_ct_pool`. So, on a session on 2026-07-28, if a server cancels its own `elicitation/create` or `sampling/createMessage`, the client handler's `RequestContext.ct` stays uncancelled.

crates/rmcp/tests/test_cancelled_response.rs
```diff
@@ -97,14 +150,374 @@ impl ServerHandler for WaitForCancelServer {
...
+            return Err(McpError::internal_error("late handler error", None));
+        }
+        Ok(ElicitResult::new(ElicitationAction::Decline))
+    }
+}
+
+#[cfg(all(feature = "client", not(feature = "local")))]
+#[tokio::test]
+async fn peer_cancels_reverse_request_without_cancelling_outbound_request() -> anyhow::Result<()> {
+    for startup in ["initialize", "direct-legacy", "direct-unknown"] {
```
DaleSeo: Could we use `rstest` instead to get better failure reports?

crates/rmcp/src/service.rs
aurokin: That distinction is intentional. The [2026-07-28 cancellation spec](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/cancellation) says server cancellations “**MUST** reference a `subscriptions/listen` request” and “Servers **MUST NOT** send `notifications/cancelled` for any other purpose.” Cancelling a server-originated elicitation or sampling request this way is therefore invalid in the modern lifecycle. Looking in `local_ct_pool` would allow that prohibited purpose and could cancel an unrelated inbound request whose ID matches an outbound subscription. For legacy peers, where either side can cancel its own requests, the other branch does cancel the inbound `RequestContext.ct`. Keeping the lifecycle split unchanged.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

crates/rmcp/tests/test_cancelled_response.rs
aurokin: Changed in a2cded7818454ee14db4dade52e6243dbd1c6d09: `rstest` now reports all six startup/ID combinations separately, with named startup cases and the existing per-case timeout preserved.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

crates/rmcp/src/service.rs
aurokin: This drop is intentional for modern peers. The [2026-07-28 cancellation rules](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/cancellation) restrict server-sent cancellation to `subscriptions/listen`: “Servers **MUST NOT** send `notifications/cancelled` for any other purpose.” The Error Handling section also says “Invalid cancellation notifications **SHOULD** be ignored,” including unknown and already-completed IDs. We therefore ignore these notifications before application dispatch, rather than treating them as valid cancellation callbacks. Legacy cancellations still reach the normal handler after cancelling the matching inbound request. Keeping this branch unchanged.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

DaleSeo (approved):

PR #1269, opened 2026-09-14
fix: preserve legacy fallback after rejected discovery

JulienCote:
Fix legacy fallback after a rejected `server/discover` request.

## Motivation and Context

A client supporting both lifecycle models should probe with `server/discover`, then fall back to `initialize` when the server reports that it is legacy.

rmcp previously marked the connection as requiring modern per-request metadata before discovery completed. If discovery returned an error, the client correctly fell back to `initialize`, but the server retained the modern metadata requirement and rejected subsequent legacy requests.

This change:

- Treats servers advertising only initialization-based protocol versions as legacy and returns `MethodNotFound` without dispatching discovery.
- Keeps the server in its pre-initialization state when discovery is rejected.
- Requires per-request metadata only after a request successfully establishes the modern lifecycle.

## How Has This Been Tested?

Added separate regression tests covering:

- A legacy-only server rejects discovery before invoking its discovery handler.
- A rejected discovery can be followed by legacy `initialize` and a metadata-light `tools/list`.

`cargo test --all-features` otherwise passed 568 tests, but the unrelated Windows-specific `transport::auth::tests::default_http_client_preserves_connection_failure_cause` assertion failed due to the platform's connection-refused error shape.

## Breaking Changes

None. This restores the expected backward-compatible lifecycle behavior.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The stdio backward-compatibility rules require clients to fall back on discovery errors other than recognized modern errors. The server therefore must not commit the connection to the modern lifecycle until discovery succeeds


crates/rmcp/src/service/server.rs
```diff
@@ -564,70 +564,107 @@ where
...
+                let (peer, peer_rx) = Peer::new(id_provider.clone(), None);
+                if matches!(&request, ClientRequest::DiscoverRequest(_)) {
+                    let context = RequestContext {
+                        ct: ct.child_token(),
+                        id: id.clone(),
+                        meta: std::mem::take(request.get_meta_mut()),
+                        extensions: std::mem::take(request.extensions_mut()),
+                        peer: peer.clone(),
+                    };
+                    let response = match service.handle_request(request, context).await {
```
DaleSeo: In `serve_server_with_ct_inner`, which manages the connection lifecycle, the discovery handler now waits forever on `context.peer.notify_progress(...)`. Sending the notification waits for an acknowledgment from the service loop, but `serve_inner` does not start until the handler returns.

PR #1270, opened 2026-09-14
fix(transport): match explicit default ports in Origin allowlist

### LINKED ISSUE #1268: Explicit default-port Origin entries reject browser-serialized origins
Configuring `StreamableHttpServerConfig::allowed_origins` with `https://client.example:443` rejects `Origin: https://client.example` with HTTP 403. Supplying `Origin: https://client.example:443` succeeds. This prevents an explicit default-port entry from matching the origin’s normal browser serialization.

Reproduced with rmcp 3.3.0 through a Streamable HTTP server, using MCP Inspector CLI to call a tool with each Origin header explicitly supplied:

| Allowed entry | Request Origin | Observed | Expected |
| --- | --- | --- | --- |
| `https://client.example:443` | `https://client.example` | HTTP 403 | Accepted |
| `https://client.example:443` | `https://client.example:443` | Accepted | Accepted |
| `https://client.example:443` | `https://client.example:8443` | HTTP 403 | HTTP 403 |
| `https://client.example` | `https://client.example:8443` | Accepted | Accepted |

[RFC 6454 §4](https://www.rfc-editor.org/rfc/rfc6454.html#section-4) assigns an omitted port the scheme’s default; [§6.2](https://www.rfc-editor.org/rfc/rfc6454.html#section-6.2) omits that default port when serializing an origin. The [current matcher](https://github.com/modelcontextprotocol/rust-sdk/blob/3075dc9152d4678775f20634fcb467a7b995dbab/crates/rmcp/src/transport/streamable_http_server/tower.rs#L838-L873) instead compares the incoming absent port directly against the configured explicit port.

Please compare an explicitly configured port against the incoming origin’s effective port: 443 for HTTPS and 80 for HTTP when omitted. Preserve the existing rule that an omitted **configured** port permits any port. This makes default-port restrictions work through the existing configuration while retaining the wildcard behavior that [Spice intentionally uses](https://github.com/spiceai/spiceai/blob/42eefac87d23ecc6c658c81231267b0177c8cecf/crates/spicepod/src/component/runtime.rs#L931-L942).

The [rustdoc](https://github.com/modelcontextprotocol/rust-sdk/blob/3075dc9152d4678775f20634fcb467a7b995dbab/crates/rmcp/src/transport/streamable_http_server/tower.rs#L111-L122) should describe that distinction: an omitted configured port permits any port; an explicit configured port restricts matching to that effective port. This preserves the omitted-port policy left unchanged by #1191 and #1192.

DaleSeo:
Fixes #1268.

## Motivation and Context

An `allowed_origins` entry with an explicit default port, like `https://client.example:443`, rejected the origin that browsers actually send, which is `Origin: https://client.example`. So the entry did not work in practice.

RFC 6454 treats an omitted port as the scheme's default port, and then removes that default port when serializing the origin. The matcher compared the missing port from the incoming request directly with the configured port. That meant a configured `:443` could only match a client that explicitly included the port.

The fix resolves the effective port on the incoming origin before comparing it. That means 443 for `https` and 80 for `http`. The configured side stays unchanged.

## How Has This Been Tested?

Add tests

## Breaking Changes

None

## Types of changes
- [x] Documentation update

## Checklist


crates/rmcp/src/transport/streamable_http_server/tower.rs
```diff
@@ -113,7 +113,10 @@ pub struct StreamableHttpServerConfig {
     /// Defaults to an empty list, which disables Origin validation for backward
     /// compatibility. A non-empty list enables validation. Requests carrying
     /// an `Origin` header must match per RFC 6454 `(scheme, host, port)`;
-    /// missing-`Origin` requests still pass. Entries must include a scheme;
+    /// missing-`Origin` requests still pass. An entry that omits the port
+    /// permits any port; an entry with an explicit port matches only that
```
alexhancock, line 117: Should it really permit any port, or if it's not listed should it permit only the default port for the scheme?

People can always configure it with a specific known port if they need one...

If we need the unconditional match, perhaps it would be better to introduce a syntax for specifying that like:

`http://localhost:*`

DaleSeo: @alexhancock A bare host shouldn't mean that. The allowlist is there to prevent DNS rebinding, so it should fail closed. I went with your `:*` idea. I didn't change the default here though since it's always behaved this way, and Spice relies on it.

PR #1281, opened 2026-09-16
fix(rmcp): tolerate empty cacheScope instead of silently dropping the whole result

### LINKED ISSUE #1242: Empty cacheScope on tools/list drops every tool

WAHIB-EL-KHADIRI:
`ListToolsResult`/`ReadResourceResult.cache_scope: Option<CacheScope>` has no
custom deserializer, unlike the sibling `ttl_ms` field, which already
tolerates out-of-range input via `deserialize_ttl_ms`. A server sending
`cacheScope: ""` (SEP-2549 only permits `"public"`/`"private"`/absent) fails
deserialization of the whole result — and because `ServerResult` is
`#[serde(untagged)]`, that failure doesn't surface as an error. It falls
through variant-by-variant to `CustomResult` (the catch-all), so callers
silently lose typed access to `.tools` / `.contents` instead of getting a
clear error or a usable result.

Fix mirrors the existing `deserialize_ttl_ms` normalize-don't-error pattern:
a new `deserialize_cache_scope` treats `""`/`null` as absent, delegates
everything else to `CacheScope`'s normal deserialization (so a genuinely
invalid value like `"PUBLIC"` still errors, it just no longer takes the
whole result down with it).

**Test:** added `repro_empty_cache_scope_drops_every_tool_via_untagged_fallthrough`
to `test_cache_hints.rs`, asserting both direct `ListToolsResult`
deserialization and the full `ServerResult` untagged path preserve the tool
list instead of degrading to `CustomResult`.

**Test results:** `cargo test -p rmcp --test test_cache_hints` → 6/6 pass (5
pre-existing + 1 new), no regressions.

Closes #1242.


Rchie121 (approved):

DaleSeo (approved):
Thanks for the fix, @WAHIB-EL-KHADIRI!

PR #1285, opened 2026-09-18
[rmcp] Notify credential stores when a refresh token is rejected

mzeng-openai:
## Summary

A provider can reject a refresh token with `invalid_grant`, but the credential store currently has no callback to retire that token. Later calls can load it again and repeat the failed exchange.

This adds an optional `CredentialStore::on_refresh_token_rejected` callback at the shared refresh boundary. It receives the credentials used for the exchange while the refresh guard is still held. The default is a no-op, so existing stores keep their behavior. Callback failures propagate as storage errors; other refresh failures do not invoke the callback.

Key review points:
- `crates/rmcp/src/transport/auth.rs`: the callback must not reacquire the refresh guard or overwrite credentials that have been replaced.

The branch starts from RMCP 3.2.0 so downstream clients can test this hook without taking an unrelated SDK upgrade.

## Test Plan

- Command: `cargo test -p rmcp --lib --features auth,client,transport-streamable-http-client,reqwest transport::auth::tests::` passed all 185 tests.
- Tests cover the attempted credential snapshot, guard lifetime, callback failure, transient errors, and unchanged default behavior.
- Command: `git diff --check` passed.


crates/rmcp/src/transport/auth.rs
```diff
@@ -2291,26 +2306,32 @@ impl AuthorizationManager {
...
+            Ok(token_result) => token_result,
+            Err(error) => match &error {
                 RequestTokenError::ServerResponse(response)
                     if response.error() == &BasicErrorResponseType::InvalidGrant =>
                 {
-                    AuthError::TokenRefreshRejected(error.to_string())
+                    self.credential_store
+                        .on_refresh_token_rejected(&stored_credentials)
+                        .await?;
+                    return Err(AuthError::TokenRefreshRejected(error.to_string()));
```
DaleSeo, line 2330: When the provider returns `invalid_grant` but the store callback fails, the `?` surfaces a `CredentialStoreError`. As a result, the client is not told to re-authorize, so it replays the same rejected exchange on its next call.

